PR #2, opened 2019-03-21
Adding the MinMaxScaler function

joaopcnogueira:
I have added one implementation of the min-max scaler along with the unit test. The min-max scaler implemented scales all the selected features between 0 and 1.


solegalli:
Hi Joao,

Thanks for reaching out and making the PR.

Scikit-Learn has an implementation of the MinMaxScaler. Why would you like to have a separate one in feature_engine instead of using sklearn?

The idea of feature engine is to provide solutions that do not exist elsewhere. 

Cheers.

joaopcnogueira:
Hi Solegalli,

I have been worked with feature_engine and I realized that it's amazing for custom pipelines. The MinMaxScaler does not have a parameter to pass the variables we want the scaler to act, so I thought I could implementing one with this parameter.

solegalli:
Yes, that is true. 

I am thinking of adding a complete new class, called Scikit-learn wrapper or similar, that takes any (or as many as possible) sklearn feature engineering transformers, and apply it to selected group of features. 

The code for that class is here:
https://github.com/trainindata/testing-and-monitoring-ml-deployments/blob/master/packages/gradient_boosting_model/gradient_boosting_model/processing/preprocessors.py#L5

Would you be happy to contribute with the addition of that piece of code to Feature-engine? If so, let me know and we get started by creating a separate issue. Thank you!

PR #35, opened 2020-04-27
circle ci automatic publish to pypi

solegalli:
@ChristopherGS would you mind having a look at my attempt to automate build and publish to Pypi from circle ci? I have specific questions, that I will try and add at each line. Thanks a lot.


.circleci/config.yml
```diff
@@ -68,3 +112,13 @@ workflows:
...
+      - package_and_upload_to_pypi:
+          requires:
+            - test_feature_engine_py36
+            - test_feature_engine_py37
+            - test_feature_engine_py38
+            - test_style
+          filters:
+            branches:
+              only:
+                - master
```
solegalli, line 124: does this mean that it will only publish the master branch? @ChristopherGS

.circleci/config.yml
```diff
@@ -68,3 +112,13 @@ workflows:
       - test_feature_engine_py37
       - test_feature_engine_py38
       - test_style
+      - package_and_upload_to_pypi:
```
solegalli, line 115: @ChristopherGS will all tests run on every branch? but only what is in master published and uploaded? see below.

.circleci/config.yml
```diff
@@ -60,6 +73,37 @@ jobs:
...
+    <<: *defaults
+    steps:
+      - checkout
+      - *prepare_venv
+      - run:
+          name: Install requirements
+          command: |
+            . venv/bin/activate
+            pip install -r requirements.txt
+            pip install -e .
```
solegalli, line 86: not sure I need to install reqs if I am installing from the setup below? this was confusing in the file from which I took inspiration. The reqs are part of the setup.py, so shouldb't be needed as additional line. right? @ChristopherGS

solegalli:
Messed up trying to delete the data sets from the repo :(

PR #37, opened 2020-04-27
update CC config for auto publish

solegalli:
@ChristopherGS here is a clean PR. Could you revise this one? thanks a lot!


.circleci/config.yml
```diff
@@ -91,7 +91,7 @@ jobs:
           command: |
             echo -e "[pypi]" >> ~/.pypirc
             echo -e "username = solegalli" >> ~/.pypirc
-            echo -e "password = $PYPI_PASSWORD" >> ~/.pypirc
+            echo -e "password = $pypi_password" >> ~/.pypirc
```
ChristopherGS, line 94: should work, but more standard is with curly braces: `${pypi_password}` caps is also preferred for evars (https://circleci.com/docs/2.0/env-vars/)

.circleci/config.yml
```diff
@@ -121,4 +121,4 @@ workflows:
           filters:
             branches:
               only:
-                - master
\ No newline at end of file
+                - master
```
ChristopherGS, line 124: yep 👍

.circleci/config.yml
```diff
@@ -91,7 +91,7 @@ jobs:
           command: |
```
ChristopherGS, line 91: above, pretty sure you don't need `pip install -e` since you're just uploading

PR #48, opened 2020-05-23
Update README.md usage section to fix import and show better example

gsheni:
- I noticed that the import statement in the usage section of the README is incorrect, this PR fixes that.
- Also, noticed the example doesn't actually have data, so I added that as well. Simple pd.DataFrame with 4 categorical columns.
- I also added some output cells with the new example to show what the `RareLabelCategoricalEncoder` does
- All of this should help to make the usage section more readable, and allow users to better understand what is going on. It will also allow users to copy and paste the code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,12 +80,36 @@ git clone https://github.com/solegalli/feature_engine.git
 
 ### Usage
 
-```
-from feature_engine.categorical_encoders import RareLabelEncoder
+```python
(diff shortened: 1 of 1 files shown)


solegalli:
Hey @gsheni thanks a lot for the fixes! Really appreciated. 
Would you mind re-making the PR to the "develop" branch?
Cheers
Sole

gsheni:
See #49

solegalli:
@gsheni thank you! Merged :)

Issue #75, opened 2020-07-10
RareLabelEncoder allow user to decide maximum number of returned categories per variable


solegalli:
Add a parameter with which the user can specify the maximum number of categories that they want for the variables

SuryaThiru:
Hi! Will a `max_n_categories` parameter and indexing with `[:max_n_categories]` suffice?
How can we handle the case where multiple categories under the tolerance percentage are present in a variable?

PR #79, opened 2020-07-24
add maximum categories param and test

SuryaThiru:
Allow user to decide the maximum number of returned categories per variable in the `RareLabelEncoder` class

Adresses issue #75 

## Example

```python
>>> from feature_engine.categorical_encoders import RareLabelCategoricalEncoder
>>> import pandas as pd
>>> data = {'var_A': ['A'] * 10 + ['B'] * 10 + ['C'] * 2 + ['D'] * 1}
>>> data = pd.DataFrame(data)
>>> data['var_A'].value_counts()
```
```
Out[1]
A    10
B    10
C     2
D     1
Name: var_A, dtype: int64
```

```python
>>> rare_encoder = RareLabelCategoricalEncoder(tol=0.05, n_categories=3)
>>> rare_encoder.fit_transform(data)['var_A'].value_counts()
```
```
Out[2]
A       10
B       10
C        2
Rare     1
Name: var_A, dtype: int64
```
```python
>>> rare_encoder = RareLabelCategoricalEncoder(tol=0.05, n_categories=3, max_n_categories=2)
>>> rare_encoder.fit_transform(data)['var_A'].value_counts()
```
```
Out[2]
A       10
B       10
Rare     3
Name: var_A, dtype: int64
```


solegalli:
Hi @SuryaThiru, can I get greedy and ask you to include the example you write here in the commit in the docs:
https://github.com/solegalli/feature_engine/blob/master/docs/encoders/RareLabelCategoricalEncoder.rst
:)

solegalli:
final request, would you mind updating the version as well so I can merge to master without errors? please use 05.01.12

The file that needs updating lives here:

SuryaThiru:
> I see an issue with this implementation. If the user sets a high tol, so that freq_idx is smaller (contain less categories) than max_n_categories, this will through an error.

If `freq_idx` is smaller than `max_n_categories`, then `freq_idx[:max_n_categories]` will simply return `freq_idx` according to python indexing. So the current implementation will keep all the categories above the tolerance but not more than max_n_categories. Is it not the intended behavior?

**Example:**
| Val | Count |
| --- | --- |
| A | 10 |
| B | 5 |
| C | 2 |
| D | 1 |

For tol = 0.1, so freq_idx = ['A', 'B', 'C']
a) max_n_categories = 4  (>3)
| Val | Count |
| --- | --- |
| A | 10 |
| B | 5 |
| C | 2 |
| Rare | 1 |

b) max_n_categories = 2  (<3)
| Val | Count |
| --- | --- |
| A | 10 |
| B | 5 |
| Rare | 3 |

solegalli:
@SuryaThiru you are right. Thanks for pointing it out. Would you be able to update the version to 0.5.12 and expand the docs with the examples in this thread?

SuryaThiru:
> @SuryaThiru you are right. Thanks for pointing it out. Would you be able to update the version to 0.5.12 and expand the docs with the examples in this thread?

Yes, I will take care of the changes.

>    1. Could we change this slightly so that the transformer either works with n_categories, or max_n_categories, but not both together?

Do we need this change as well? I thought the n_categories was just to select columns on which the transformation will be applied.

feature_engine/categorical_encoders.py
```diff
@@ -689,6 +689,15 @@ class RareLabelCategoricalEncoder(BaseEstimator, TransformerMixin):
         to find frequent labels. If the variable contains less categories, all
         of them will be considered frequent.
 
+    max_n_categories: int, default=None
+        the maximum no of categories that should be considered frequent.
+        If None, all categories with frequency above the tolerance will be
+        considered.
+
+        .. note::
```
solegalli, line 697: Hi @SuryaThiru I think this 3 lines in the docs are not needed, since you well pointed out that the desired behaviour is to check for the max categories that fullfill the tol criteria.

solegalli:
> Do we need this change as well? I thought the n_categories was just to select columns on which the transformation will be applied.

Nope, you pointed it out well in your previous comment. Thanks

solegalli:
Merged thank you!!!

SuryaThiru:
Welcome solegalli ! Happy to contribute to your great work 😁

PR #80, opened 2020-07-27
fix API reference

michalgromiec:
It was referencing to EqualWidthDiscretiser, should referencing to UserInputDiscretizer


solegalli:
Hi @michalgromiec thank you so much for the fix. Could I ask you to update the version to 0.5.11? Otherwise I will have a build error when merging to master.

You can find the file I am talking about here:

thank you!

solegalli:
Thank you! @michalgromiec

PR #81, opened 2020-07-27
Update missing_data_imputers.py

GLevV:
Changed CategoricalVariableImputer to accept custom entries in imputation method


solegalli:
Hi @GLevV this is looking really good. Thank you so much.!

Would you mind adding a test for this new functionality? 
You would only need to duplicate the bit of code of test case 1 in this script:
https://github.com/solegalli/feature_engine/blob/master/tests/test_missing_data_imputer.py#L190
(lines 190 to 208)
where you change the string of the **fill_value** to something else, example "Other" and make sure that the transformed dataset contains the string "Other". You would have to change all strings "Missing" for "Other" in that bit of code you duplicate. 

And finally, so that I can merge to master, would you mind updating the version number in this file:
to 14? so it should say 0.5.14

Thanks a lot!

solegalli:
Awesome work @GLevV thank you so much!

PR #82, opened 2020-07-28
FIX filling NaNs in example of ArbitraryOutlierCapper

michalgromiec:
without filling NaNs exception is being raised


solegalli:
Hi @michalgromiec Sorry to be a pain, could you increase version to end in 13? so 0.5.13. Thank you!

solegalli:
Never mind, I have a following PR I will update version there! Thanks a lot!

Issue #73, opened 2020-07-10
CategoricalVariableImputer


solegalli:
Allow user to select the string with which to replace missing data. At the moment, the only value allowed is the default one == 'missing'

GLevV:
I rewrite code to allow custom string for imputation, and also we can modify mode imputer for case of multiple modes: we can uniformly randomly choose mode for every nan. It will save the frequencies of modes but transform method should be redefined

solegalli:
Hi @GLevV thanks for getting in touch, and sorry for the late response, I took a few days off :) 

Would you like to modify the transformer to allow the user to input the string they wish for the imputation? If yes, could you make a feature branch with the name of this issue and make a PR to master with your changes? It would be great if you can also add a test to check that your modification works, and then update the version before we merge the PR. But let's take it one step at the time and start with the class modification and the test?

For the bimodal imputation, if you would like this addition, could you create a separate issue, and add some evidence as to when and how often this imputation is / could be used?

Cheers
Sole

PR #83, opened 2020-07-29
Arbitrary number imputer dict attribute

michalgromiec:
addresses issue #73 

Added `imputer_dict` parameter for `ArbitraryNumberImputer`. It allows passing different Arbitrary Numbers for different variables.
If `imputer_dict` is None, previously developed process with `variables` and `arbitrary_number` is valid.


feature_engine/missing_data_imputers.py
```diff
@@ -285,10 +293,16 @@ def fit(self, X, y=None):
...
-        self.variables = _find_numerical_variables(X, self.variables)
+        if self.imputer_dict:
+            self.variables = _find_numerical_variables(X, self.imputer_dict.keys())
+        else:
+            self.variables = _find_numerical_variables(X, self.variables)
 
         # create the imputer dictionary
-        self.imputer_dict_ = {var: self.arbitrary_number for var in self.variables}
+        if self.imputer_dict:
+            self.imputer_dict_ = {var: self.imputer_dict[var] for var in self.variables}
```
solegalli, line 303: would it not work here if we just do: `self.imputer_dict_ = self.imputer_dict`, so that we avoid the loop?

Because in any case, lines 296-299 will check that the variables are numerical. So if that passes, then I think we are safe to use the dictionary?

michalgromiec: indeed, good point, I've applied this in next commit.

solegalli:
HI @michalgromiec I changed the wording of the `_define_imputer_dict()` so we can use it in other classes, like for example here:
https://github.com/solegalli/feature_engine/blob/master/feature_engine/outlier_removers.py#L252

Are you ok with it? Awaiting confirmation to merge. Cheers!

michalgromiec:
Hi @solegalli good idea! just only we have to be aware that we are requiring that all dictionary values have to be float or integer, so maybe we could change method name to `_define_numerical_dict`, the same way there are `_find_numerical_variables` and `_find_categorical_variables`? Or even add parameter in this method e.g. `values_type` which will allow us to steering the code.

Something like:
```python
def _define_dict(the_dict, **dict_values_type**):
    # Check that the entered dictionary is indeed a dictionary of integers and floats
    # Can take None as value
    if not the_dict:
        the_dict = the_dict
    elif isinstance(the_dict, dict):
        if **dict_values_type == 'numerical'** and not all([isinstance(x, float) or isinstance(x, int) for x in the_dict.values()]):
            raise ValueError('All values in the dictionary must be integer or float')
    else:
        raise TypeError('The parameter can only take a dictionary or None')
    return the_dict
```

solegalli:
@michalgromiec good point! Perhaps `_define_numerical_dict `is better so we keep consistency with names. Shall you do the modification? thank you!

michalgromiec:
@solegalli done, method name changed to `_define_numerical_dict` :)

solegalli:
> @solegalli done, method name changed to `_define_numerical_dict` :)

thank you!!!

Issue #69, opened 2020-07-10
add feature combination by math operations


solegalli:
https://github.com/solegalli/packt_featureengineering_cookbook/blob/master/ch09-mathematical-transformations/Recipe1-Add-Multiply-Features.ipynb

michalgromiec:
@solegalli If it's possible, I can take this one. I am planning to allow user to pass math operation as a list in parameter or calculate all methods from above Jupyter if parameter will be None.

PR #91, opened 2020-08-06
Math addition transformer

michalgromiec:
issue #69 - implementation of AdditionTransformer from mathematical transformation


feature_engine/mathematical_transformers.py
```diff
@@ -0,0 +1,86 @@
...
+
+        Each operation will result in operation column in result dataset.
+
+        If None, the transformer will calculate each of permitted
+        operations: ['sum', 'prod', 'mean', 'std', 'max', 'min']
+    """
+
+    def __init__(self, variables=None, math_operations=None):
+        self.variables = variables
+        self.math_operations = math_operations
```
solegalli, line 29: here I would put something like:
```
        if isinstance(math_operations, list):
            if any(elem_par not in self.math_operations_permitted for elem_par in self.math_operations):
                raise KeyError("At least one of math_operations is not found in permitted operations set. "
                               "Choose one of ['sum', 'prod', 'mean', 'std', 'max', 'min']")
            self.math_operations = math_operations
```
which is what you wrote below but here

feature_engine/mathematical_transformers.py
```diff
@@ -0,0 +1,86 @@
...
+        self.math_operations_permitted = ['sum', 'prod', 'mean', 'std', 'max', 'min']
+
+        if self.math_operations is None:
+            self.operations_ = self.math_operations_permitted
```
solegalli, line 33: we can't include attributes with trailing score outside the fit() method, because the is_fitted() method from scikit-learn, which we use in the transform method  checks exactly for trailing underscores.

feature_engine/mathematical_transformers.py
```diff
@@ -0,0 +1,86 @@
...
+            The training input samples.
+            Can be the entire dataframe, not just the variables to transform.
+
+        y : None
+            y is not needed in this transformer. You can pass y or None.
+        """
+        X = super().fit(X, y)
+        self.input_shape_ = X.shape
+
+        if len(self.variables) <= 1:
```
solegalli, line 57: this check should probably go in the init. Here feels a bit late for me. We can alert the user earlier.

feature_engine/mathematical_transformers.py
```diff
@@ -0,0 +1,86 @@
...
+        Returns
+        -------
+
+        X_transformed : pandas dataframe of shape = [n_samples, n_features + n_operations]
+            The dataframe with operations results added.
+        """
+        X = super().transform(X)
+
+        for operation in self.operations_:
+            variables_set_name = f"{operation}({','.join(self.variables)})"
```
solegalli, line 83: is this returning the operation plus all the columns separated by a comma? I wonder if this is a good idea. I think for cases where the user wants to sum variables indiscriminately and without really giving a lot of thought on what they are doing it make work. But in my experience, when we add variables, we add variables that make sense to add. For example, if a car had an accident, and we have binary variables indicating which part of the car is broken, we add those up into a variable called "total_damage"

I think I would prefer giving the user the posibility to name their variables. Could we add this in the init? and explain it in the docstrings?

feature_engine/mathematical_transformers.py
```diff
@@ -0,0 +1,86 @@
+from feature_engine.base_transformers import BaseNumericalTransformer
+
+
+class AdditionTransformer(BaseNumericalTransformer):
```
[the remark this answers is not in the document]
michalgromiec: will change in next commit

feature_engine/mathematical_transformers.py
```diff
@@ -0,0 +1,86 @@
...
+        Each operation will result in operation column in result dataset.
+
+        If None, the transformer will calculate each of permitted
+        operations: ['sum', 'prod', 'mean', 'std', 'max', 'min']
+    """
+
+    def __init__(self, variables=None, math_operations=None):
+        self.variables = variables
+        self.math_operations = math_operations
+        self.math_operations_permitted = ['sum', 'prod', 'mean', 'std', 'max', 'min']
```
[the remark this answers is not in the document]
michalgromiec: This variable is used also in line 36, I would remove self from here and use it as simple variable within `__init__` method, what do you think?

feature_engine/mathematical_transformers.py
```diff
@@ -0,0 +1,86 @@
...
+    The AdditionTransfomer() applies basic mathematical operations using aggregation of features.
+
+    Parameters
+    ----------
+
+    variables: list, default=None
+        The list of numerical variables to be transformed. If None, the transformer
+        will find and select all numerical variables.
+
+    math_operations: list, default=None
```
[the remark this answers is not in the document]
michalgromiec: good point, I'll change this in next commit

feature_engine/mathematical_transformers.py
```diff
@@ -0,0 +1,86 @@
...
+        Each operation should be a string and must be one of elements
+        from list: ['sum', 'prod', 'mean', 'std', 'max', 'min']
+
+        Each operation will result in operation column in result dataset.
+
+        If None, the transformer will calculate each of permitted
+        operations: ['sum', 'prod', 'mean', 'std', 'max', 'min']
+    """
+
+    def __init__(self, variables=None, math_operations=None):
```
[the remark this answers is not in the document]
michalgromiec: yes, exactly what I'll implement :)

feature_engine/mathematical_transformers.py
michalgromiec: thx, very good point!

feature_engine/mathematical_transformers.py
michalgromiec: good insight, I'll change this

feature_engine/mathematical_transformers.py
michalgromiec: Yes, exactly as you described - operation plus all columns separated by comma. 
It's funny, I had this in my intermediate version of code, but decided not to allow user to do this. If user had two MathematicalVariableCombinator with other columns set but same "alias", the second result set would overwrite the first...

But it seems it's better to allow user to use such alias, I'll implement this in next commit.

michalgromiec: just one thing to verify - if there are more than one operation, we need to reflect this in results, so I'll implement this that if there are more than one operations to calculate, I'll leave operation in result column.

feature_engine/mathematical_transformers.py
```diff
@@ -0,0 +1,86 @@
+from feature_engine.base_transformers import BaseNumericalTransformer
+
+
+class AdditionTransformer(BaseNumericalTransformer):
+    """
+    The AdditionTransfomer() applies basic mathematical operations using aggregation of features.
```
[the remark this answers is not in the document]
michalgromiec: thx!

feature_engine/mathematical_transformers.py
michalgromiec: right, I'll change this.

feature_engine/mathematical_combination.py
```diff
@@ -0,0 +1,94 @@
...
+        X = super().transform(X)
+
+        for operation in self.operations:
+            if self.variables_set_alias:
+                if len(self.operations) == 1:
+                    variables_set_name = self.variables_set_alias
+                else:
+                    variables_set_name = f"{operation}({self.variables_set_alias})"
+            else:
+                variables_set_name = f"{operation}({','.join(self.variables)})"
```
solegalli, line 91: could we change , for  - ? I kind of feel that is nicer and easier to handle for column names.

feature_engine/mathematical_combination.py
```diff
@@ -0,0 +1,94 @@
+from feature_engine.base_transformers import BaseNumericalTransformer
+
+
+class MathematicalVariableCombinator(BaseNumericalTransformer):
+    """
+    The MathematicalVariableCombinator() applies basic mathematical operations across features,
+    returning one additional feature as a result.
```
solegalli, line 7: returning 1 or more additional features as a result. For example, if we have the variables number_payments_first_quarter, number_payments_second_quarter, number_payments_third_quarter and number_payments_fourth_quarter, we can use the MathematicalVariableCombinator to calculate the total number of payments and mean number of payments as follows:

transformer = MathematicalCombinator(variables=['number_payments_first_quarter', 'number_payments_second_quarte'r, 'number_payments_third_quarter', 'number_payments_fourth_quarter'], math_operations=['sum','mean'], new_variables_name=['total_number_payments', 'mean_number_payments'])

trasnformer.fit_transform(X)

The transformed X will contain the additional features total_number_payments and mean_number_payments, plus the original set of variables.

Could you add all of the above to the docstrings?

feature_engine/mathematical_combination.py
```diff
@@ -0,0 +1,94 @@
...
+
+    variables: list, default=None
+        The list of numerical variables to be transformed. If None, the transformer
+        will find and select all numerical variables.
+
+    math_operations: list, default=['sum', 'prod', 'mean', 'std', 'max', 'min']
+        The list of basic math operations to be used in transformation.
+
+        Each operation should be a string and must be one of elements
+        from list: ['sum', 'prod', 'mean', 'std', 'max', 'min']
```
solegalli, line 20: from **the** list: ['sum', 'prod', 'mean', 'std', 'max', 'min']

feature_engine/mathematical_combination.py
```diff
@@ -0,0 +1,94 @@
...
+
+        Each operation will result in operation column in result dataset.
```
solegalli, line 22: Each operation will result in a new variable that will be added to the transformed dataset.

Could you replace by the above sentence?

feature_engine/mathematical_combination.py
```diff
@@ -0,0 +1,94 @@
...
+
+    variables_set_alias: string, default=None
```
solegalli, line 24: could we replace by "new_variables_name":

feature_engine/mathematical_combination.py
```diff
@@ -0,0 +1,94 @@
...
+        Alias for variables set, allow for replacing all columns name from result column with this string.
```
solegalli, line 25: Name of the newly created variables. The user can enter a name or a list of names for the newly created features (recommended). User must enter one name for each mathematical transformation indicated in the math_operations attribute. That is, if you want to perform mean and sum of features, you should enter 2 new variable names. If you perform only mean of features, enter 1 variable name. Alternatively,  if you chose to perform all mathematical transformations, please enter 6 new variable names. 

The name of the variables indicated by the user should coincide with the order in which the mathematical operations are initialised in the transformer. That is, if you set math_operations = ['mean', 'prod'], the first new variable name will be assigned to the mean of the variables and the second variable name to the product of the variables.

If new_variable_names=None, the transformer will assign an arbitrary name to the newly created features starting by the name of the mathematical operation, followed by the variables combined separated by -. 

Could you add the above to the docstrings?

feature_engine/mathematical_combination.py
```diff
@@ -0,0 +1,94 @@
...
+
+        If math_operation list contain more than one element, operation will be added to variables_set_alias.
```
solegalli, line 27: I think we can remove this line.

feature_engine/mathematical_combination.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+
+    def __init__(self, variables=None, math_operations=['sum', 'prod', 'mean', 'std', 'max', 'min'],
+                 variables_set_alias=None):
+        self.variables = variables
+        self.variables_set_alias = variables_set_alias
+        _math_operations_permitted = ['sum', 'prod', 'mean', 'std', 'max', 'min']
```
solegalli, line 34: nice!

feature_engine/mathematical_combination.py
```diff
@@ -0,0 +1,94 @@
...
+        else:
+            raise KeyError("math_operations parameter must be a list or None")
+
+        if self.variables and len(self.variables) <= 1:
+            raise KeyError(
+                "MathematicalVariableCombinator requires two or more features to make proper transformations.")
+
+    def fit(self, X, y=None):
+        """
+        Fits source dataset, verifies if variables parameter contains more than one variable.
```
solegalli, line 50: Performs dataframe checks. Selects variables to transform if None were indicated by the user. Creates dictionary of column to transformation mappings

Could we replace by the above docstring?

feature_engine/mathematical_combination.py
```diff
@@ -0,0 +1,94 @@
...
+
+        X : pandas dataframe of shape = [n_samples, n_features]
+            The training input samples.
+            Can be the entire dataframe, not just the variables to transform.
+
+        y : None
+            y is not needed in this transformer. You can pass y or None.
+        """
+        X = super().fit(X, y)
+        self.input_shape_ = X.shape
```
solegalli, line 60: Should we add a dictionary attribute: combination_dict_, that contains the mappings of the new column names to the transformation applied to the features? for example: combination_dict_ = {'total_payments':'sum', 'mean_payments':'mean'}

What do you think?

I am thinking, what if 3 months later you want to know what this transformer did, maybe this helps? thoughts? with this dictionary plus the variables attribute, I think the user should be in a good shape to reconstruct this.? maybe it is an overkill?

feature_engine/mathematical_combination.py
```diff
@@ -0,0 +1,94 @@
...
+
+        Returns
+        -------
+
+        X_transformed : pandas dataframe of shape = [n_samples, n_features + n_operations]
+            The dataframe with operations results added.
+        """
+        X = super().transform(X)
+
+        for operation in self.operations:
```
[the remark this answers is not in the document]
solegalli: I added this comment first, and then modified the docs above. Maybe with that modification the comment here is clearer. Sorry, I am thinking as I review :)

solegalli:
Hi @michalgromiec thanks a lot for the amount of work. Could I also ask the following:

Because this is a new feature that adds functionality at the moment not present in Feature-engine we should also add a bit of information to the documentation. Could I also ask you to include the following files:

- [ ]  nside the folder **docs** make a subfolder called **mathematical_combination**
- [ ] inside the **mathematical_combination** folder we need an index.rst and another rst file with the name of the transformer you just created where we add details to populate the dosctring of the transformer with sphinx, plus an example of how it can be used (code).
- [ ] in the doc/index.rst we need to update the index to include the mathematical combination functionality under the wrappers
- [ ] in the doc/index.rst, please add the**mathematical_combination** folder to the toctree at the end.

Test that the docs run and build successfully. 

Thanks a lot!

feature_engine/mathematical_combination.py
michalgromiec: thx, I'll change this in next commit

feature_engine/mathematical_combination.py
michalgromiec: sure, I'll change this in next commit

feature_engine/mathematical_combination.py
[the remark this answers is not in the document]
michalgromiec: I thought alias is for variables set and only replaces variables join in new features names (like "sum(user_alias)"). Now I get that alias is simply new feature name, that way it need to be a list like you said above. I'll change this in next commit.

feature_engine/mathematical_combination.py
michalgromiec: yes, thanks

feature_engine/mathematical_combination.py
michalgromiec: will add in next commit

feature_engine/mathematical_combination.py
michalgromiec: I think it's a very nice idea. What's more, we will use this dictionary in `transform` method to replace this strange "if-then" code.

feature_engine/mathematical_combination.py
michalgromiec: sounds more precise, will change this in next commit

feature_engine/mathematical_combination.py
```diff
@@ -0,0 +1,94 @@
...
+                raise KeyError("At least one of math_operations is not found in permitted operations set. "
+                               "Choose one of ['sum', 'prod', 'mean', 'std', 'max', 'min']")
+            self.operations = math_operations
+        else:
+            raise KeyError("math_operations parameter must be a list or None")
+
+        if self.variables and len(self.variables) <= 1:
+            raise KeyError(
+                "MathematicalVariableCombinator requires two or more features to make proper transformations.")
+
```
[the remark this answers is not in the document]
michalgromiec: added in next commit

feature_engine/mathematical_combination.py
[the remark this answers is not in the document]
michalgromiec: also propose that new_variables_names would be a list, it would be more clear not to allow for strings

solegalli:
@michalgromiec all done! Thank you!

PR #95, opened 2020-08-08
Fixed typos and grammatical errors in README

richardcsuwandi:
I fixed some typos and grammatical errors in README  to aid readability.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,16 +8,16 @@
 ![Documentation Status](https://readthedocs.org/projects/feature-engine/badge/?version=latest)
 
 
-Feature-Engine is a Python library with multiple transformers to engineer features for use in machine learning models. Feature-engine's transformers follow scikit-learn's functionality with fit() and transform() methods to first learn the transforming parameters from data and then transform the data.
+Feature-engine is a Python library with multiple transformers to engineer features for use in machine learning models. Feature-engine's transformers follow scikit-learn's functionality with fit() and transform() methods to first learn the transforming parameters from data and then transform the data.
 
 
-## Feature-Engine features in the following resources:
+## Feature-engine features in the following resources:
 
 * [Feature Engineering for Machine Learning, Online Course](https://www.udemy.com/feature-engineering-for-machine-learning/?couponCode=FEATENGREPO).
 
 * [Python Feature Engineering Cookbook](https://www.packtpub.com/data/python-feature-engineering-cookbook)
 
-## Blogs about Feature-Engine:
+## Blogs about Feature-engine:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,32 +8,32 @@
 ![Documentation Status](https://readthedocs.org/projects/feature-engine/badge/?version=latest)
 
 
-Feature-engine is a Python library with multiple transformers to engineer features for use in machine learning models. Feature-engine's transformers follow Scikit-learn functionality with fit() and transform() methods to first learn the transforming paramenters from data and then transform the data.
+Feature-Engine is a Python library with multiple transformers to engineer features for use in machine learning models. Feature-engine's transformers follow scikit-learn's functionality with fit() and transform() methods to first learn the transforming parameters from data and then transform the data.
 
 
-## Feature-engine features in the following resources:
```
solegalli, line 14: Feature-engine is an arbitrary name so for some reason I chose it to be F with capital and e with small one, and I wrote it that way in all docs. So I would prefer if we kept it this way. Could you reverse the change?

README.md
```diff
@@ -147,9 +147,9 @@ BSD 3-Clause
 
 ### References
 
-Many of the engineering and encoding functionality is inspired by this [series of articles from the 2009 KDD competition](http://www.mtome.com/Publications/CiML/CiML-v3-book.pdf).
+Many of the engineering and encoding functionalities are inspired from this [series of articles from the 2009 KDD Competition](http://www.mtome.com/Publications/CiML/CiML-v3-book.pdf).
```
solegalli, line 150: I think inspired should be by instead of from. Could we keep the original?

README.md
```diff
@@ -147,9 +147,9 @@ BSD 3-Clause
...
 
-To learn more about the rationale, functionality, pros and cos of each imputer, encoder and transformer, refer to the [Feature Engineering for Machine Learning, Online Course](https://www.udemy.com/feature-engineering-for-machine-learning/?couponCode=FEATENGREPO)
+To learn more about the rationale, functionality, pros, and cons of each imputer, encoder, and transformer, refer to the [Feature Engineering for Machine Learning, Online Course](https://www.udemy.com/feature-engineering-for-machine-learning/?couponCode=FEATENGREPO)
```
solegalli, line 152: could we remove the comma after pros?

README.md
richardcsuwandi: Sure, I will change it back!

README.md
richardcsuwandi: You're right! Actually both are grammatically okay, but 'by' would be more suitable in this context.

README.md
richardcsuwandi: Oh yes, my bad! I will change that too!

richardcsuwandi:
Hi, I've made some changes to the updated README as you suggested. Please kindly review it and there's anything I could help!

PR #96, opened 2020-08-08
initial commit for FeatureEliminator

KarthikKothareddy:
added a suite of tests
added a section under docs for selector
updated the index to include new class

### DIFF
--- a/feature_engine/feature_selection.py
+++ b/feature_engine/feature_selection.py
@@ -9,15 +9,14 @@ from feature_engine.variable_manipulation import _define_variables
 
 class DropFeatures(BaseEstimator, TransformerMixin):
     """
-    The FeatureEliminator() drops the list of variable(s) as provided by the user
-    from the dataframe and returns the subset of original dataframe with remaining
-    variables.
+    DropFeatures() drops the list of variable(s) indicated by the user
+    from the original dataframe and returns the remaining variables.
 
     Parameters
     ----------
 
     features_to_drop : str or list, default=None
-        Desired variable/s to be dropped from the dataframe
+        Variable/s to be dropped from the dataframe
 
     """
 
@@ -26,13 +25,13 @@ class DropFeatures(BaseEstimator, TransformerMixin):
 
     def fit(self, X, y=None):
         """
-        Verifies that the passed input X if of the type pandas dataframe
+        Verifies that the input X is a pandas dataframe
 
         Parameters
         ----------
 
         X: pandas dataframe of shape = [n_samples, n_features]
-            The input dataframe on which the feature elimination has to be performed
+            The input dataframe
 
         y: None
             y is not needed for this transformer
@@ -40,6 +39,14 @@ class DropFeatures(BaseEstimator, TransformerMixin):
         """
         # check input dataframe
         X = _is_dataframe(X)
+
+        # check for non existent columns
+        non_existent = [x for x in self.features if x not in X.columns]
+        if non_existent:
+            raise KeyError(
+                f"Columns '{', '.join(non_existent)}' does not exist in the input dataframe, "
+                f"please check the columns and enter a new list of features to drop"
+            )
 
         # add input shape
         self.input_shape_ = X.shape
@@ -71,7 +78,6 @@ class DropFeatures(BaseEstimator, TransformerMixin):
         # check for input consistency
         _check_input_matches_training_df(X, self.input_shape_[1])
 
-        X = X.copy()
         X = X.drop(columns=self.features)
 
         # check for a case where all columns are dropped
--- a/feature_engine/feature_selection.py
+++ b/feature_engine/feature_selection.py
@@ -1,16 +1,13 @@
-# Authors: Soledad Galli <solegalli@protonmail.com>
-# License: BSD 3 clause
 
-import numpy as np
-import pandas as pd
 import warnings
 
 from sklearn.base import BaseEstimator, TransformerMixin
-from feature_engine.dataframe_checks import _is_dataframe
+from sklearn.utils.validation import check_is_fitted
+from feature_engine.dataframe_checks import _is_dataframe, _check_input_matches_training_df
 from feature_engine.variable_manipulation import _define_variables
 
 
-class FeatureEliminator(BaseEstimator, TransformerMixin):
+class DropFeatures(BaseEstimator, TransformerMixin):
     """
     The FeatureEliminator() drops the list of variable(s) as provided by the user
     from the dataframe and returns the subset of original dataframe with remaining
@@ -25,10 +22,6 @@ class FeatureEliminator(BaseEstimator, TransformerMixin):
     """
 
     def __init__(self, features_to_drop=None):
-
-        if not any(isinstance(features_to_drop, t) for t in [str, list]):
-            raise ValueError("features_to_drop must be a string or list object")
-
         self.features = _define_variables(features_to_drop)
 
     def fit(self, X, y=None):
@@ -40,12 +33,17 @@ class FeatureEliminator(BaseEstimator, TransformerMixin):
 
         X: pandas dataframe of shape = [n_samples, n_features]
             The input dataframe on which the feature elimination has to be performed
+
         y: None
             y is not needed for this transformer
 
         """
         # check input dataframe
         X = _is_dataframe(X)
+
+        # add input shape
+        self.input_shape_ = X.shape
+
         return self
 
     def transform(self, X):
@@ -64,6 +62,24 @@ class FeatureEliminator(BaseEstimator, TransformerMixin):
             The transformed dataframe with subset of variables.
 
         """
-        return X.drop(columns=self.features).copy()
+        # check if fit is called prior
+        check_is_fitted(self)
+
+        # check input dataframe
+        X = _is_dataframe(X)
+
+        # check for input consistency
+        _check_input_matches_training_df(X, self.input_shape_[1])
+
+        X = X.copy()
+        X = X.drop(columns=self.features)
+
+        # check for a case where all columns are dropped
+        if X.shape[1] == 0:
+            warnings.warn(
+                "The resulting dataframe has no columns after dropping all existing variables"
+            )
+
+        return X
 
 
--- a/feature_engine/feature_selection.py
+++ b/feature_engine/feature_selection.py
@@ -1,0 +1,69 @@
+# Authors: Soledad Galli <solegalli@protonmail.com>
+# License: BSD 3 clause
+
+import numpy as np
+import pandas as pd
+import warnings
+
+from sklearn.base import BaseEstimator, TransformerMixin
+from feature_engine.dataframe_checks import _is_dataframe
+from feature_engine.variable_manipulation import _define_variables
+
+
+class FeatureEliminator(BaseEstimator, TransformerMixin):
+    """
+    The FeatureEliminator() drops the list of variable(s) as provided by the user
+    from the dataframe and returns the subset of original dataframe with remaining
+    variables.
+
+    Parameters
+    ----------
+
+    features_to_drop : str or list, default=None
+        Desired variable/s to be dropped from the dataframe
+
+    """
+
+    def __init__(self, features_to_drop=None):
+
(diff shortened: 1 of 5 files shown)


KarthikKothareddy:
Yes, this is doable. I can cancel the PR 97 (given that I have permissions if not, can you please do that for me?) and PR 96 will have only code related to feature eliminator.

KarthikKothareddy:
Here are the summary of changes made in the recent commit.
- Removed the first two lines 
- Changed the name of transformer from FeatureEliminator to DropFeatures
- Removed the redundant check in the constructor for str and list
- Added "input_shape_" creation in the fit method 
- Added empty line in the doc string between X and y
- Added check_is_fitted method in transform function
- Added _is_dataframe check in the transform function
- Added check for matching input shape in transform
- Added a warning when all columns are dropped

feature_engine/feature_selection.py
```diff
@@ -0,0 +1,85 @@
...
+from sklearn.utils.validation import check_is_fitted
+from feature_engine.dataframe_checks import _is_dataframe, _check_input_matches_training_df
+from feature_engine.variable_manipulation import _define_variables
+
+
+class DropFeatures(BaseEstimator, TransformerMixin):
+    """
+    The FeatureEliminator() drops the list of variable(s) as provided by the user
+    from the dataframe and returns the subset of original dataframe with remaining
+    variables.
```
solegalli, line 14: DropFeatures() drops the list of variable(s) indicated by the user
    from the original dataframe and returns the remaining variables.

Could we replace doscstrings by the above?

feature_engine/feature_selection.py
```diff
@@ -0,0 +1,85 @@
...
+
+    Parameters
+    ----------
+
+    features_to_drop : str or list, default=None
+        Desired variable/s to be dropped from the dataframe
```
solegalli, line 20: Variable/s to be dropped from the dataframe

(remove Desired)

feature_engine/feature_selection.py
```diff
@@ -0,0 +1,85 @@
...
+
+    """
+
+    def __init__(self, features_to_drop=None):
+        self.features = _define_variables(features_to_drop)
+
+    def fit(self, X, y=None):
+        """
+        Verifies that the passed input X if of the type pandas dataframe
```
solegalli, line 29: Verifies that the input X **is** a pandas dataframe

could we change for the above?

feature_engine/feature_selection.py
```diff
@@ -0,0 +1,85 @@
...
+
+        Parameters
+        ----------
+
+        X: pandas dataframe of shape = [n_samples, n_features]
+            The input dataframe on which the feature elimination has to be performed
```
solegalli, line 35: **The input dataframe.**

We can remove the rest of the string.

feature_engine/feature_selection.py
```diff
@@ -0,0 +1,85 @@
...
+
+        y: None
+            y is not needed for this transformer
```
solegalli, line 38: y is not needed for this transformer. You can pass y or None. 

Could you add the above?

feature_engine/feature_selection.py
```diff
@@ -0,0 +1,85 @@
...
+        # add input shape
+        self.input_shape_ = X.shape
+
+        return self
+
+    def transform(self, X):
+        """
+        Drops the variable or list of variables provided from the original dataframe
+        and returns the dataframe with subset of variables.
+
```
solegalli, line 53: Drops the variable or list of variables indicated by the user from the original dataframe
and returns a new dataframe with the remaining subset of variables.

Could we rephrase as above please?

feature_engine/feature_selection.py
```diff
@@ -0,0 +1,85 @@
...
+        Parameters
+        ----------
+        X: pandas dataframe
+            The input dataframe on which the feature elimination has to be performed
```
solegalli, line 57: The input dataframe from which the features will be dropped.

Could we rephrase by the above please?

feature_engine/feature_selection.py
```diff
@@ -0,0 +1,85 @@
...
+
+        Returns
+        -------
+        X_transformed: pandas dataframe of shape = [n_samples, n_features - len(features_to_drop)]
+            The transformed dataframe with subset of variables.
```
solegalli, line 62: The transformed dataframe with the remaining subset of variables.

Could we rephrase by the above please?

tests/test_feature_selection.py
```diff
@@ -0,0 +1,101 @@
+# Authors: Soledad Galli <solegalli1@gmail.com>
+# License: BSD 3 clause
+import pytest
```
solegalli, line 3: please remove lines with my name and email. You are creating the code. So no need to have my name on it :)

tests/test_feature_selection.py
```diff
@@ -0,0 +1,101 @@
+# Authors: Soledad Galli <solegalli1@gmail.com>
```
solegalli, line 1: could we change the name of the testing script to test_DropFeatures.py please?

tests/test_feature_selection.py
```diff
@@ -0,0 +1,101 @@
...
+# License: BSD 3 clause
+import pytest
+import pandas as pd
+
```
solegalli, line 5: according to pep8, imports are done alphabetically, so first pandas then pytest. Could you re-order?

tests/test_feature_selection.py
```diff
@@ -0,0 +1,101 @@
...
+from feature_engine.feature_selection import DropFeatures
+
+
+def test_FeatureEliminator_initial(dataframe_vartypes, dataframe_na):
+    # test case 1: with dataframe_vartypes
```
solegalli, line 10: def test_DropFeatures_drop_2_variables(dataframe_vartypes)

could we rename test to something that indicates what the test is testing? like this example above?

Like this, we also do not need the comment on line 10, which does not explain really clearly what is testing

also, I don't think you need dataframe_na here

feature_engine/feature_selection.py
```diff
@@ -0,0 +1,85 @@
...
+    Parameters
+    ----------
+
+    features_to_drop : str or list, default=None
+        Desired variable/s to be dropped from the dataframe
+
+    """
+
+    def __init__(self, features_to_drop=None):
+        self.features = _define_variables(features_to_drop)
```
solegalli, line 25: could we rename this attribute to features_to_drop please? also then later in the tests?

feature_engine/feature_selection.py
```diff
@@ -0,0 +1,85 @@
...
+        X: pandas dataframe of shape = [n_samples, n_features]
+            The input dataframe on which the feature elimination has to be performed
+
+        y: None
+            y is not needed for this transformer
+
+        """
+        # check input dataframe
+        X = _is_dataframe(X)
+
```
solegalli, line 43: could we add a check in the fit method to corroborate that the features to drop are indeed present in the training dataframe, and if not, raise an error?

tests/test_feature_selection.py
```diff
@@ -0,0 +1,101 @@
...
+    # init params
+    assert transformer.features == ['City', 'dob']
+    # transform params
+    assert X.shape == (4, 3)
+    assert type(X) == pd.DataFrame
+    pd.testing.assert_frame_equal(X, df)
+
+
+def test_FeatureEliminator_non_existing_variables(dataframe_vartypes):
+    # test case 2: passing variables that doesn't exist
```
solegalli, line 30: nice name. Could you please replace by DropFeautres? And could you do this for all the tests below?

feature_engine/feature_selection.py
solegalli: For this, i think we could use the _define_all_variables in the variable_manipulation script, if I remember correctly

tests/test_feature_selection.py
```diff
@@ -0,0 +1,101 @@
...
+    with pytest.raises(KeyError):
+        transformer = DropFeatures(features_to_drop="last_name")
+        X = transformer.fit_transform(dataframe_vartypes)
+
+
+def test_FeatureEliminator_non_existing_index(dataframe_vartypes):
```
solegalli, line 36: how is this error different from the previous one? and from the following one? could we not just have the one on line 29 and remove the following 2 on lines 36 and 43?
In my head they are the same, I don't understand the difference. If they are different, could you please add more info to help the reviewer understand why this tests are important? what function of the code you wrote in feature_selection.py are you testing?

tests/test_feature_selection.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def test_FeatureEliminator_integer_as_variable(dataframe_vartypes):
+    # test case 4: with integer as feature_to_drop
+    with pytest.raises(KeyError):
+        transformer = DropFeatures(features_to_drop=99)
+        X = transformer.fit_transform(dataframe_vartypes)
+
+
+def test_FeatureEliminator_different_input():
```
solegalli, line 50: here you are testing an error when input is not a dataframe?

can we rename the test to test_drop_features_errors_if_fit_input_not_dataframe? or if transform_input_not_dataframe, and maybe separate fit and transform? it should error in both

tests/test_feature_selection.py
```diff
@@ -0,0 +1,101 @@
...
+def test_FeatureEliminator_drop_all_columns(dataframe_vartypes):
+    # test case 6: dropping all columns
+    transformer = DropFeatures(features_to_drop=list(dataframe_vartypes.columns))
+    X = transformer.fit_transform(dataframe_vartypes)
+
+    assert len(X) == len(dataframe_vartypes)
+    assert X.shape == (4, 0)
+
+
+def test_FeatureEliminator_drop_all_columns_warn(dataframe_vartypes):
```
solegalli, line 66: test_drop_featres_raises_warning_when_returning_empty_dataframe

tests/test_feature_selection.py
```diff
@@ -0,0 +1,101 @@
...
+    # test case 7: dropping all columns produces warning check
+    with pytest.warns(UserWarning):
+        transformer = DropFeatures(features_to_drop=list(dataframe_vartypes.columns))
+        X = transformer.fit_transform(dataframe_vartypes)
+
+
+def test_FeatureEliminator_empty_list(dataframe_vartypes):
```
solegalli, line 73: I think if the user passes and empty list, the transformer should return a warning, or an error, at the init level. Could we add that check and then test for the error here instead?

tests/test_feature_selection.py
```diff
@@ -0,0 +1,101 @@
...
+    # test case 8: passing an empty list
+    transformer = DropFeatures(features_to_drop=[])
+    X = transformer.fit_transform(dataframe_vartypes)
+
+    assert len(X) == len(dataframe_vartypes)
+    assert X.shape == dataframe_vartypes.shape
+
+
+def test_FeatureEliminator_valid_string(dataframe_vartypes):
```
solegalli, line 82: test_drop_feaures_with_variable_to_drop_as_string()

this test is testing the same functionality that the tests on lines 29, 36 and 43, so maybe let's put them together? one after the other? also, mind my comment above, maybe we remove some of those tests?

solegalli:
@I saw your answer, I copy here:
As far having a check for variables that doesn't exist, pandas drop() does raise a "KeyError" if it finds a variable not present in the dataframe. I also have a test case to check this (test case 2). Will that be sufficient or do you need an explicit check?

Good point. I normally add the checks so that the user understands automatically why this is failing. I think pandas errors are also clear, but then in the error traceback, it will take the user all the way to pandas docs, instead of to feature-engine's docs and for those who are not super savy, may be difficult to trouble shoot. Whereas if we add an error saying column so and so are not in your dataframe, please check the columns and enter a new list of features to drop, they know what is going on.

Thoughts? I am open

KarthikKothareddy:
@solegalli Below are the summary of most recent changes,
- Removed redundant test cases and renamed all of them to camel case
- Modified doc strings as per your suggestion
- added explicit check for non existent columns 

Let me know if that sounds good or need any further changes.

KarthikKothareddy:
So, @solegalli I merged the other PR you opened. Just to cross check again, can I cancel this PR ?

PR #97, opened 2020-08-09
test_return_object_rarelabelencoder

KarthikKothareddy:


feature_engine/feature_selection.py
```diff
@@ -0,0 +1,69 @@
+# Authors: Soledad Galli <solegalli@protonmail.com>
+# License: BSD 3 clause
```
solegalli, line 2: please erase rows 1 and 2

feature_engine/feature_selection.py
```diff
@@ -0,0 +1,69 @@
...
+import numpy as np
+import pandas as pd
+import warnings
+
+from sklearn.base import BaseEstimator, TransformerMixin
+from feature_engine.dataframe_checks import _is_dataframe
+from feature_engine.variable_manipulation import _define_variables
+
+
+class FeatureEliminator(BaseEstimator, TransformerMixin):
```
solegalli, line 13: is the name FeatureEliminator clear enough? 
I think I would prefer to call the transformer **DropFeatures**? 

Eliminator gives me the sensation that there is some sort of fancy algorithm running, and we are only removing the features that the user inputs. What do you think?

feature_engine/feature_selection.py
```diff
@@ -0,0 +1,69 @@
...
+        ----------
+
+        X: pandas dataframe of shape = [n_samples, n_features]
+            The input dataframe on which the feature elimination has to be performed
+        y: None
+            y is not needed for this transformer
+
+        """
+        # check input dataframe
+        X = _is_dataframe(X)
```
solegalli, line 48: the fit method needs to return at least 1 attribute with a trailing underscore, for example self.input_shape_, because the trailing underscore is that the is_fitted method checks when executing the transform method, to check that the user used fit to fit the transformer before using transform. I will point to example code as I go on with the review.

solegalli: fit method should include the `self.input_check`, as here:
https://github.com/solegalli/feature_engine/blob/master/feature_engine/variable_transformers.py#L222

I think we should also include a check to corroborate that the variables entered by the user are present in the dataframe. If not, return an error. 

What I am not sure is if we should also check absence of na in the training set. Thoughts?

feature_engine/feature_selection.py
```diff
@@ -0,0 +1,69 @@
...
+    features_to_drop : str or list, default=None
+        Desired variable/s to be dropped from the dataframe
+
+    """
+
+    def __init__(self, features_to_drop=None):
+
+        if not any(isinstance(features_to_drop, t) for t in [str, list]):
+            raise ValueError("features_to_drop must be a string or list object")
+
```
solegalli, line 31: `_define_variables` checks if the input is a string or a list, so I don't think we need to include these 2 lines of code.

feature_engine/feature_selection.py
```diff
@@ -0,0 +1,69 @@
...
+    def fit(self, X, y=None):
+        """
+        Verifies that the passed input X if of the type pandas dataframe
+
+        Parameters
+        ----------
+
+        X: pandas dataframe of shape = [n_samples, n_features]
+            The input dataframe on which the feature elimination has to be performed
+        y: None
```
solegalli, line 43: please add an empty line between parameters, otherwise, it does not look nice when the docs are built.

feature_engine/feature_selection.py
```diff
@@ -0,0 +1,69 @@
...
+        X: pandas dataframe
+            The input dataframe on which the feature elimination has to be performed
+
+        Returns
+        -------
+        X_transformed: pandas dataframe of shape = [n_samples, n_features - len(features_to_drop)]
+            The transformed dataframe with subset of variables.
+
+        """
+        return X.drop(columns=self.features).copy()
```
solegalli, line 67: in the transform method we need to check that the transformer is fitted, that the X provided by the user is a dataframe, and that the dataframe matches the size of that provided to train the transformer. This is to help the user keep its pipelines tidy. See here for an example:
https://github.com/solegalli/feature_engine/blob/master/feature_engine/base_transformers.py#L155

In the return, please return just the final output and not an operation. Doesn't look too nice in my opinion.

tests/test_categorical_encoder.py
```diff
@@ -446,7 +446,7 @@ def test_OneHotCategoricalEncoder(dataframe_enc_big, dataframe_enc_big_na):
         encoder.transform(dataframe_enc_big_na)
 
 
-def test_RareLabelEncoder(dataframe_enc_big, dataframe_enc_big_na):
+def test_RareLabelEncoder(dataframe_enc_big, dataframe_enc_big_na, dataframe_enc_rare):
     # test case 1: defo params, automatically select variables
```
solegalli, line 450: could you make this file a separate PR?

tests/test_feature_selection.py
```diff
@@ -0,0 +1,80 @@
+# Authors: Soledad Galli <solegalli1@gmail.com>
+# License: BSD 3 clause
+import pytest
+import pandas as pd
+
+from feature_engine.feature_selection import FeatureEliminator
+
```
solegalli, line 7: could you change the format of the tests so that each test case is actually a separate test? see for example this file:
https://github.com/solegalli/feature_engine/blob/master/tests/test_wrappers.py

I will be reformatting all the tests to this, because it is much simple to know which test is falling if they are individual. Each test should have a name describing that it is testing.

tests/test_feature_selection.py
```diff
@@ -0,0 +1,80 @@
...
+    with pytest.raises(ValueError):
+        transformer = FeatureEliminator(features_to_drop=99)
+        X = transformer.fit_transform(dataframe_vartypes)
+
+    # test case 5: passing a different input than dataframe
+    with pytest.raises(TypeError):
+        transformer = FeatureEliminator(features_to_drop=["Name"])
+        X = transformer.fit_transform({"Name": ["Karthik"]})
+
+    # test case 6: dropping all columns
```
solegalli, line 48: this is actually a good point. Should we not warn the user that they are dropping all the variables and returning an empty dataframe in the FeatureEliminator?

tests/test_feature_selection.py
```diff
@@ -0,0 +1,80 @@
...
+    transformer = FeatureEliminator(features_to_drop=list(dataframe_vartypes.columns))
+    X = transformer.fit_transform(dataframe_vartypes)
+
+    assert len(X) == len(dataframe_vartypes)
+    assert X.shape == (4, 0)
+
+    # test case 7: passing an empty list
+    transformer = FeatureEliminator(features_to_drop=[])
+    X = transformer.fit_transform(dataframe_vartypes)
```
solegalli, line 57: I think we should catch this at the transformer. A check that the list is not empty, if not, return an error. Could you please add it?

solegalli:
hi @KarthikKothareddy this is great work! Thank you so much for the contribution. I have a few suggestions here and there, and a few checks that need to be added in to the transformer. Could you please have a look? Many thanks!

solegalli:
@renamed to PR to something more descriptive of the issue.

solegalli:
@KarthikKothareddy  I just noticed that there is another PR for the feature eliminator, and this pr should be only for the rarelabel encoder. I suggest you follow the steps I indicate in PR96. Then we close this PR. And you make a new PR just with the change for the rare label encoder test. Does that sound doable? 

Please make sure you check out each feature branch from master. and that master is in sync with the remote master in feature-engine's repo.

Thank you!

tests/test_feature_selection.py
KarthikKothareddy: Absolutely, I kind of thought along the same lines but didn't wanted to mess with the existing structure of tests. I will reformat them.

tests/test_categorical_encoder.py
KarthikKothareddy: I was under the assumption that this will be treated as a separate one. I will drop it and create another.

feature_engine/feature_selection.py
KarthikKothareddy: I am very bad in naming things so, I think DropFeatures makes sense. If you're okay with DropFeatures I will go ahead and rename it.

feature_engine/feature_selection.py
KarthikKothareddy: > the fit method needs to return at least 1 attribute with a trailing underscore, for example self.input_shape_, because the trailing underscore is that the is_fitted method checks when executing the transform method, to check that the user used fit to fit the transformer before using transform. I will point to example code as I go on with the review.

I will make sure that is added.

KarthikKothareddy: > fit method should include the `self.input_check`, as here:
> https://github.com/solegalli/feature_engine/blob/master/feature_engine/variable_transformers.py#L222
> 
> I think we should also include a check to corroborate that the variables entered by the user are present in the dataframe. If not, return an error.
> 
> What I am not sure is if we should also check absence of na in the training set. Thoughts?

As far having a check for variables that doesn't exist, pandas drop() does raise a "KeyError" if it finds a variable not present in the dataframe. I also have a test case to check this (test case 2). Will that be sufficient or do you need an explicit check?

Issue #66, opened 2020-07-10
encode categorical variables with trees


solegalli:
Add encder to replace labels directly by the output of a decision tree.

SuryaThiru:
Hi! I'm curious as to how this works. Can you point me to some relevant resources?

solegalli:
The method was used for a KDD competition and is described in this series of articles:
http://www.mtome.com/Publications/CiML/CiML-v3-book.pdf

In particular, article starting on page 21, and the method is described on page 27.

This method is implemented already in Feauture-engine for numerical variables, in the discretisation transformers:
https://github.com/solegalli/feature_engine/blob/master/feature_engine/discretisers.py#L276

To extend this to categorical variables, I think the solution is straightforward, we would need to first transform the categories into numbers, I think arbitrary encoding should work just fine. And then, at the back of it, apply the decision tree as I do with numerical variables in the bit of code from discretisers.

We could even provide the option to transform categories into numbers arbitratily or following the target mean, so we could use the ordinal categorical encoder first:
https://github.com/solegalli/feature_engine/blob/master/feature_engine/categorical_encoders.py#L128

and then add a bit of code to add the decision tree layer to the encoding.

Would you be up for this? if yes, please let me know before starting with the code, because I was advised to re-organise the code a bit into smaller chunks, so this might be a good opportunity to start.

Thank you for the engagement!

SuryaThiru:
Yes, I can work on this.

SuryaThiru:
> 
> Would you be up for this? if yes, please let me know before starting with the code, because I was advised to re-organise the code a bit into smaller chunks, so this might be a good opportunity to start.
> 

Are you planning to refactor the library a bit before any further additions?

solegalli:
@SuryaThiru please ignore what I said. I will make refactor as part of a big PR later in the year or in 2021. Let's continue as always. Cheers. Sole

solegalli:
#98

PR #98, opened 2020-08-09
DecisionTreeCategoricalEncoder

SuryaThiru:
**Implementation of DecisionTreeCategoricalEncoder** 

Uses categorical encoders to convert categories to numerical values and performs discretisation using `DecisionTreeDiscretiser` on the numerical values.

- [x] Implementation
- [x] Test cases
- [x] Documentation

Relevant issues: #66


SuryaThiru:
@solegalli Can you take a look at the current implementation and let me know if this would work for us?

feature_engine/categorical_encoders.py
```diff
@@ -10,6 +10,8 @@
 from feature_engine.dataframe_checks import _is_dataframe, _check_input_matches_training_df, _check_contains_na
```
solegalli, line 10: could we move this transformer to its own python script? 

I was advised that current code format of Feature-engine is not best of practices as it contains more than 500 lines of code on each script.

Please move the new transformer to a new script called decision_tree_encoders.py

feature_engine/categorical_encoders.py
```diff
@@ -827,3 +829,144 @@ def transform(self, X):
...
+        +-------------+------------------------------------+
+        | 'woe'       |  WoERatioCategoricalEncoder        |
+        +-------------+------------------------------------+
+
+    cat_encoder_args: dict, default=None
+        Arguments for the categorical encoder which will be passed as keyword
+        arguments for initialization. Refer the encoder's doc for the
+        details of the parameters.
+
+    discretiser_args: dict, default=None
```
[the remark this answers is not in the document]
solegalli: It would be just copy and paste from the other script, but keeping only the bits of information relevant for classification.

feature_engine/categorical_encoders.py
```diff
@@ -827,3 +829,144 @@ def transform(self, X):
...
+        X : pandas dataframe of shape = [n_samples, n_features]
+            The training input samples.
+            Can be the entire dataframe, not just the categorical variables.
+
+        y : pandas series.
+            Target, value depends on the cat_to_num_encoder used, see docs of
+            the corresponding encoder.
+        """
+        # check input dataframe
+        X = _is_dataframe(X)
```
solegalli, line 925: I think some of this code can be avoided when importing base numerical imputer

tests/test_categorical_encoder.py
```diff
@@ -10,6 +10,7 @@
 from feature_engine.categorical_encoders import WoERatioCategoricalEncoder
```
solegalli, line 10: could we move the tests to a separate script called test_decision_tree_encoder.py?

tests/test_categorical_encoder.py
```diff
@@ -520,3 +521,28 @@ def test_RareLabelEncoder(dataframe_enc_big, dataframe_enc_big_na):
     df = pd.DataFrame(df)
     pd.testing.assert_frame_equal(X, df)
 
+
+def test_DecisionTreeCategoricalEncoder(dataframe_enc, dataframe_enc_rare, dataframe_enc_na):
+    # test case 1: user provides the categorical encoder
```
solegalli, line 526: could we separate each test case into a separate test as done here:
https://github.com/solegalli/feature_engine/blob/62d1ed19a6c32b6710516f262561cddca19ae7e9/tests/test_wrappers.py

It is better practice and it helps debugging later on. I know I do it as you did here, but I was advised to break i down :) so I am trying to incorporate better practices as we add more bits of code.

feature_engine/categorical_encoders.py
```diff
@@ -827,3 +829,144 @@ def transform(self, X):
             X[self.variables] = X[self.variables].astype('O')
 
         return X
+
+
+class DecisionTreeCategoricalEncoder(BaseEstimator, TransformerMixin):
```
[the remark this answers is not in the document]
SuryaThiru: Shouldn't we be using the base categorical transformer? If that is the case it is missing a fit method that could be added (the other categorical encoders could also use this). The base also requires an encoder dict, while we're currently using the pipeline as the encoder. We can add an additional `encoder_dict`  attribute to the class as well.

feature_engine/categorical_encoders.py
SuryaThiru: In my opinion, we should keep the encoder in the categorical encoder for the time being to keep the API structure intact. During a future refactor, we can move the individual encoders to separate modules under packages.
for eg :
```
categorical_encoders/
  __init__.py
  decision_tree.py
  count_frequency.py
  ...
discretisers/
  __init__.py
  equal_frequency.py
  equal_width.py
  ...
```
Or if you feel DecisionTreeCategoricalEncoder being in a separate module outside of categorical_encoders is more appropriate, that would work too. Let me know your thoughts on this.

solegalli: makes sense. Thank you

feature_engine/categorical_encoders.py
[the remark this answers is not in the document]
solegalli: Good point! I thought of the baseNumerical, because it has all the code we need in the transform method. But it will be confusing for future developers (or even myself) if we inherit for a categorical encoder a numerical base transformer. So I would say keep as it is.
The base categorical has a lot of functionality that we don't need here. And I am not sure we can reconstruct the original variable after first encoding and then using trees. So I would not include the inverse_transform for this transformer. And thus, we do not need a new encoder_dict.

But I think it would be important to help users navigate the result of this transformer from the docstrings.

Thanks a lot for this! Well spotted!

feature_engine/categorical_encoders.py
```diff
@@ -827,3 +829,144 @@ def transform(self, X):
...
+
+        cat_encoder = CatEncoder(variables=variables, **cat_encoder_args)
+
+        # initialize decision tree discretiser
+        tree_discretiser = DecisionTreeDiscretiser(variables=variables, **discretiser_args)
+
+        # pipeline for the encoder
+        self.encoder_ = Pipeline([('categorical_encoder', cat_encoder),
+            ('tree_discretiser', tree_discretiser)])
+
```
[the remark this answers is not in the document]
solegalli: thank you!

feature_engine/categorical_encoders.py
```diff
@@ -827,3 +829,144 @@ def transform(self, X):
...
+        # find categorical variables or check that those entered by the user
+        # are of type object
+        self.variables = _find_categorical_variables(X, self.variables)
+
+        # check if dataset contains na
+        _check_contains_na(X, self.variables)
+
+        self.input_shape_ = X.shape
+
```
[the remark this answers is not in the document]
solegalli: thank you!

solegalli:
@SuryaThiru All good. Agreed what you suggested. Cheers.

Issue #113, opened 2020-09-08
feature_selection: remove constant and quasiconstant features


solegalli:
constant feature: only shows 1 value in all observations
quasi_constant feature: most values are the same.

Main code:

```
constant_features = [
    feat for feat in X_train.columns if X_train[feat].nunique() == 1
]
```

```
quasi_constant_feat = []
for feature in X_train.columns:

    # find the predominant value
    predominant = (X_train[feature].value_counts() / np.float(
        len(X_train))).sort_values(ascending=False).values[0]

    # evaluate predominant feature
    if predominant > 0.998:
        quasi_constant_feat.append(feature)

len(quasi_constant_feat)
```

Tejash-Shah:
Hi @solegalli,

I was curious about the [DropFeatures class](https://github.com/solegalli/feature_engine/blob/045eb14f14b359b5bbbc100bd71573abe6f2a5b3/feature_engine/feature_selection.py#L7) in [feature_selection.py](https://github.com/solegalli/feature_engine/blob/master/feature_engine/feature_selection.py) and I was wondering what is the purpose of this class? Looking at the code it looks like it just drops the features as provided to ```features_to_drop``` argument and feels like a wrapper around ```df.drop()```. 

Is it work in progress? Do you have any plans to add constant_feature and quasi_constant functionality to this package? 

Thanks, 
Tejash

solegalli:
Hi @Tejash-Shah 

Yes, that is the function of drop features. And yes, I want to include functionality to drop constant and quasi-constant features. Would you like to give it a go at trying to make this transformer?

Tejash-Shah:
Sure, would be happy to contribute:) I can create a Transformer class based on TransformerMixin, with two attributes, ```constant_features_``` and ```non_constant_features_```. In that way it is scikit-learn compatible. However, we would be missing ```get_support()``` functionality which comes from ```SelectorMixin```. Any idea what functionality we would perhaps miss if we inherit ```TransformerMixin``` and not ```SelectorMixin``` . I think ```get_support()``` could be replaced by ```constant_features_``` and ```non_constant_features_``` attributes?

solegalli:
Yes, I thought we should include the get_support() method to keep it similar to sklearn selectors. 

[...]

PR #116, opened 2020-09-16
Remove constant and non quasi constant features

Tejash-Shah:
Hi @solegalli as per our discussion at #113 , I have added the following functionalities:
Transformer: SelectFeatures() 
1. By default drop constant features
2. Drop quasi constant features if asked by the user based on threshold
3. Retain the values of ```constant_features_```, ```quasi_constant_features``` and ```selected_features_``` as attributes. 

I have added test suites in ```tests/test_feature_selection.py```. Let me know if you have any suggestions. Thanks!


Tejash-Shah:
On a side note, I just realized ```SelectFeatures()``` might be a vague name according to the functionality. I can probably change it to ```DropConstantAndQuasiConstantFeatures()``` if that is preferable.

feature_engine/feature_selection.py
```diff
@@ -88,3 +89,138 @@ def transform(self, X):
...
+    By default, all constant variables are dropped.
+    Quasi-constant variables are those variables which are almost constant. In other words, those features have the same
+    value for large subset of total observations.
+    This transformer works for both, numerical and categorical variables.
+
+    Parameters
+    ----------
+
+    drop_quasi_constant_features : bool, default=False
+        Whether to drop quasi constant features
```
solegalli, line 107: I don't think we need this parameter. From the threshold the user can indicate if it will be constant or quasi_constant as well. If threshold is 1, then drops only constant. If threshold is smaller than 1, then quasi_constant as well. 
The less parameters the easier to understand and use

feature_engine/feature_selection.py
```diff
@@ -88,3 +89,138 @@ def transform(self, X):
...
+
+    quasi_constant_threshold: float, default=None
```
solegalli, line 109: can we rename the parameter as **threshold**?

feature_engine/feature_selection.py
```diff
@@ -88,3 +89,138 @@ def transform(self, X):
...
+
+    constant_features_: list
+        List of constant features in a dataframe
+
+    quasi_constant_features_: list
+        List of quasi constant features in a dataframe
+
+    selected_features_: list
+        List of selected features which are non-constant and non-quasi constant
+
```
solegalli, line 124: these 3 attributes are created in the fit() method. We can only create attributes with trailing underscores _ in fit, as the sklearn is_fitted() class looks for these underscores to determine if the transformer is fitted. Please move to fit()

feature_engine/feature_selection.py
```diff
@@ -88,3 +89,138 @@ def transform(self, X):
...
+    """
+    def __init__(self, drop_quasi_constant_features=False, quasi_constant_threshold=None):
+
```
solegalli, line 127: init parameters should be threshold, with value 1 by default to remove only constant variables. Threshold can also take values between 0 and 1 to remove quasi-constant.

Another parameters for init is variables, which gives the user the opportunity to select which variables they want examined and removed. By default, the parameter takes None, and would select all variables in the dataframe.

feature_engine/feature_selection.py
```diff
@@ -88,3 +89,138 @@ def transform(self, X):
...
+        self.drop_quasi_constant_features = drop_quasi_constant_features
+        self.quasi_constant_threshold = quasi_constant_threshold
+
+        self.constant_features_ = []
+        self.selected_features_ = []
+        self.quasi_constant_features_ = []
+        self._input_shape_ = ()
+
```
solegalli, line 135: these attributes should go in fit(). I am also thinking that we could have only **selected_features_**, **features_to_remove_**  and **input_shape_** (please note input shape should not have a leading _).

This will simplify the code quite a bit and make the transformer run faster

feature_engine/feature_selection.py
```diff
@@ -88,3 +89,138 @@ def transform(self, X):
...
+        if not self.drop_quasi_constant_features and self.quasi_constant_threshold is not None:
+            warnings.warn(
+                'Setting a quasi_constant_threshold has no effect since drop_quasi_constant_features is False. You '
+                'should leave quasi_constant_threshold to its default (None), or set drop_quasi_constant_features=True.'
+                , FutureWarning
+            )
```
solegalli, line 141: please change this as per using only a threshold parameter. Check that threshold takes values between 0 and 1

feature_engine/feature_selection.py
```diff
@@ -88,3 +89,138 @@ def transform(self, X):
...
+            y is not needed for this transformer. You can pass y or None.
+
+        """
+
+        # data frame check
+        X = _is_dataframe(X)
+
+        # get input shape
+        self._input_shape_ = X.shape
+
```
solegalli, line 168: without leading underscore

feature_engine/feature_selection.py
```diff
@@ -88,3 +89,138 @@ def transform(self, X):
...
+        # if total constant features is equal to total features raise an error
+        if len(self.constant_features_) == len(X.columns):
+            raise ValueError("The resulting dataframe will have no columns after dropping all constant features")
+
+        if self.drop_quasi_constant_features:
+            # quasi-constant features
+            self.quasi_constant_features_ = [feat for feat in X.columns if X[feat].value_counts(
+                normalize=True, sort=True, ascending=False).iloc[0] > self.quasi_constant_threshold
+                                             and feat not in self.constant_features_
+                                             and feat in self.selected_features_]
```
solegalli, line 184: would it be possible to make this bit more readable? without so many ifs? I think the use of only the parameter threshold would help.

feature_engine/feature_selection.py
```diff
@@ -88,3 +89,138 @@ def transform(self, X):
...
+
+            self.selected_features_ = [feat for feat in self.selected_features_
+                                       if feat not in self.quasi_constant_features_]
+
```
solegalli, line 188: this is good. I wondered if we can optimize the code by using sets and check for differences taking advantage of python functionality? instead of the list comprehension?

solegalli (changes requested):
Hi @Tejash-Shah 

First of all, thank you so much for your contribution! Second, I am really sorry it took me a few days to reply. I will try and be more responsive next week,

Third, I think this transformer could be simplified a bit. I think we should offer only 2 parameters in the init, the first one is variables, to allow the user to select if the transformer should check all variables or only a subset. 

The second parameter is threshold. By default, threshold=1, means remove only constant features. If user changes threshold to 0.98, then transformer would remove constant and quasi_constant features.

The init should check that the input of threshold is a float between 0 and 1.

All attributes with trailing underscore need to be generated only by the fit method. We should create only 2 in my opinion, selected_features_ and input_shape.

The code to remove constant or quasi_constant could be something like:

if threshold ==1, then do as in line 173. 

Else do as in code pasted in the issue:
```

quasi_constant_feat = []
for feature in X_train.columns:

    # find the predominant value
    predominant = (X_train[feature].value_counts() / np.float(
        len(X_train))).sort_values(ascending=False).values[0]

    # evaluate predominant feature
    if predominant > 0.998:
        quasi_constant_feat.append(feature)

len(quasi_constant_feat)
```

These are my suggestions. Let me know if you agree. Thanks a lot!

feature_engine/feature_selection.py
```diff
@@ -88,3 +89,138 @@ def transform(self, X):
         return X
 
 
+class SelectFeatures(TransformerMixin, BaseEstimator):
+    """
```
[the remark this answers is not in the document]
solegalli: Wonder if we should just call the transformer DropConstantFeatures() it is too long otherwise?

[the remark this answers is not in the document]
Tejash-Shah: I was thinking ```DropQuasiConstantFeatures()``` since it gives the user a hint of what it actually does. The threshold of 1 will become constant features. If not 1, it becomes quasi constant + constant. Let me know...

feature_engine/feature_selection.py
```diff
@@ -88,3 +89,138 @@ def transform(self, X):
...
+    SelectFeatures()
+
+    Drops constant and quasi constant variables from a dataframe.
+    Constant variables are those variables which show only one value for all the observations.
+    By default, all constant variables are dropped.
+    Quasi-constant variables are those variables which are almost constant. In other words, those features have the same
```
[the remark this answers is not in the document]
Tejash-Shah: Thanks for catching. Will fix this.

feature_engine/feature_selection.py
Tejash-Shah: Yes, makes sense. Fixed.

feature_engine/feature_selection.py
Tejash-Shah: Yes checking it here, will change the name to threshold
``` 
if self.quasi_constant_threshold <= 0 or self.quasi_constant_threshold >= 1:
    raise ValueError("Value of threshold should be between 0 and 1")
```

feature_engine/feature_selection.py
Tejash-Shah: If we use the following code wouldn't that capture constant feature in quasi_constant_feat list since predominant for the constant feature would be 1? We would ultimately need to remove constant features from the quasi_constant_feat list which does the first if statement. I would revisit ```and feat in self.selected_features_``` and see it's needed or not. 
Rest division by ```len``` and ```sort_values()``` is taken care by ```value_counts()``` arguments. What do you think? And I guess list comprehension would be faster than for loop
```
quasi_constant_feat = []
for feature in X_train.columns:

    # find the predominant value
    predominant = (X_train[feature].value_counts() / np.float(
        len(X_train))).sort_values(ascending=False).values[0]

    # evaluate predominant feature
    if predominant > 0.998:
        quasi_constant_feat.append(feature)

len(quasi_constant_feat)
````

File: feature_engine/feature_selection.py (hunk 9 above)
Comment (Tejash-Shah) [reply feature_engine/feature_selection.py:188]: I did not intentionally use set because it produces unordered elements and it would disrupt the column ordering in original df and could cause inconvenience. Like in below df, I remove D and set produces [B,A,C,E] while with LC ordering is preserved which is [A,B,C,E]. I could go with set just wanted to avoid the issue here. Let me know your thoughts.  

set

File: feature_engine/feature_selection.py (hunk 10 above)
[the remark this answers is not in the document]
Comment (solegalli) [reply feature_engine/feature_selection.py:93]: I am not sure what is best. Let's go with the long name then: DropConstantAndQuasiConstantFeatures()

File: feature_engine/feature_selection.py (hunk 6 above)
Comment (solegalli) [reply feature_engine/feature_selection.py:141]: Something like this: https://github.com/solegalli/feature_engine/blob/master/feature_engine/categorical_encoders.py#L714

File: feature_engine/feature_selection.py (hunk 8 above)
Comment (solegalli) [reply feature_engine/feature_selection.py:184]:
```
if tol==1:
    self.selected_features_ = [feat for feat in X.columns if feat not in self.constant_features_]

else: 
    self.selected_features_ = []
    for feature in X_train.columns:

        # find the predominant value
        predominant = (X_train[feature].value_counts() / np.float(
                 len(X_train))).sort_values(ascending=False).values[0]

       # evaluate predominant feature
       if predominant > self.tol:
           self.selected_features_.append(feature)
```
 That is the functionality I am thinking of. Would you agree. Only thing to check is if this would also capture constant or would through an error. But I think it should work

File: feature_engine/feature_selection.py (hunk 9 above)
Comment (solegalli) [reply feature_engine/feature_selection.py:188]: ok, makes sense. Thanks for the explanation. I don't think we would need this bit of code though, when we refurbish the transformer as per the comments in the review.

**Comment by solegalli** [comment]
@Tejash-Shah thanks for the answers to my comments.

File: feature_engine/feature_selection.py (hunk 2 above)
Comment (Tejash-Shah) [reply feature_engine/feature_selection.py:109]: Changed to ```tol``` as per your suggestion below.

File: feature_engine/feature_selection.py (hunk 8 above)
Comment (Tejash-Shah) [reply feature_engine/feature_selection.py:184]: Maybe there is a typo here
```
if predominant < self.tol:
    self.selected_features_.append(feature)
```

Please have a look at the final code here, ignore the indentation(copy+paste stuff) and if everything looks good I'll use this.
```


 class DropConstantAndQuasiConstantFeatures(TransformerMixin, BaseEstimator):

 """

 Drops constant and quasi constant variables from a dataframe.
 Constant variables are those variables which show only one value for all the observations.
 By default, all constant variables are dropped.
 Quasi-constant variables are those variables which are almost constant. In other words, those features which have
 the same value for a large subset of total observations.
 This transformer works for both numerical and categorical variables.

 Parameters
 ---------

 tol: float, default=1
 Features showing tol percentage of values will be dropped

 variables: list, default=None
 The list of variables to inspect. If None, the transformer will select all variables in the dataframe

 Attributes
 ----------

 selected_features_: list
 List of selected features which are non-constant and non-quasi constant

 """

 def __init__(self, tol=1, variables=None):

 if tol < 0 or tol > 1:
 raise ValueError("tol takes values between 0 and 1")

 self.tol = tol
 self.variables = _define_variables(variables)

 def fit(self, X, y=None):

 """
 Find constant and non-quasi constant features from a dataframe

 Parameters
 ----------

 X: pandas dataframe of shape = [n_samples, n_features]
 The input dataframe

 y: None
 y is not needed for this transformer. You can pass y or None.

 """

 # check input dataframe
 X = _is_dataframe(X)

 # find all variables or check that those entered are present in the dataframe
 self.variables = _find_all_variables(X, self.variables)

 if self.tol == 1:

 # constant features
 constant_features = [feat for feat in self.variables if X[feat].nunique() == 1]

 self.selected_features_ = [feat for feat in X.columns if feat not in constant_features]

 # if total constant features is equal to total features raise an error
 if len(constant_features) == len(X.columns):
 raise ValueError("The resulting dataframe will have no columns after dropping all constant features")

 else:
 constant_plus_quasi_constant_features = []
 self.selected_features_ = []

 for feat in self.variables:

 predominant = (X[feat].value_counts() / np.float(len(X))).sort_values(ascending=False).values[0]

 if predominant > self.tol:
 constant_plus_quasi_constant_features.append(feat)

 # if total constant features is equal to total features raise an error
 if len(constant_plus_quasi_constant_features) == len(X.columns):
 raise ValueError("The resulting dataframe will have no columns after dropping "
 "constant and quasi constant features")

 self.selected_features_ = [feat for feat in X.columns if feat not in constant_plus_quasi_constant_features]

 self.input_shape_ = X.shape

 return self

 def transform(self, X):
 """
 Drops the constant and non-quasi constant features from a dataframe and returns the dataframe with
 selected features.

 Parameters
 ----------
 X: pandas dataframe
 The input dataframe from which features will be dropped

 Returns
 -------
 X_transformed: pandas dataframe of shape = [n_samples, n_features - (constant_features+quasi constant features)]
 The transformed dataframe with the remaining subset of variables.

 """

 # check if fit is performed prior to transform
 check_is_fitted(self)

 # check if input is a dataframe
 X = _is_dataframe(X)

 # check if number of columns in test dataset matches to train dataset
 _check_input_matches_training_df(X, self.input_shape_[1])

 # returned selected features
 X = X[self.selected_features_].copy()

 return X
```

feature_engine/feature_selection.py
Tejash-Shah: Not sure what do you mean by ```features_to_remove_```? Is it the same as ```variables``` in the below code? If not could you please elaborate?

solegalli: ignore that attribute. Let's just go with `selected_features_`

feature_engine/feature_selection.py
solegalli: It is easier to review code if you make it on the script and push them to the branch. You will also be able to know if it is working properly through the tests. Would you be able to make the changes in the code and push them to the branch? Then, we take it from there.

Tejash-Shah: Pushed, let me know for any changes. Thanks

Tejash-Shah:
Hi @solegalli,

Here are the changes 
- [x] separate each test example into a different test function 
- [x] the test function name says exactly what the test is testing
- [x] add asserts for the fit() attributes 
- [x] remove shape from transform

- [x] index.rst: add a new transformer to the list
- [x] within docs/selection add a script with the new transformer and an example of how it is used
- [x] in docs/selection/index.rst add a link to the new rst in 2

Let me know if I missed anything. Thank you :)

Tejash-Shah:
Thanks a lot for your help @solegalli. If everything looks good feel free to merge and close.

solegalli:
@Tejash-Shah all done! Thank you so much! The transformer will be available as from next release. I will let you know and add you to the changelog. Have a good week.

PR #132, opened 2020-10-01
Added SimilarColumns 

thibaultbl:
Added SimilarColumns transformers :
   - If a column is in train (fitted dataset) but not in test, then the column will be created in test dataset with missing value only.
   - If a column is in test but not in train, it will be dropped. 

I use it a lot for research (a lot less in production), and it is really useful.


solegalli:
Hi @thibaultbl thank you so much for the contribution. 

I understand what the transformer is doing. Would you help me understand what are the possible scenarios where the train set variables and the test set variables would not match? How are these datasets generated?

I know that one hot encoding can lead to such scenarios, and this is prevented if using Feature-engine's OneHotEncoder. But I have not encountered this issue otherwise.

Thanks a lot.

thibaultbl:
Hi @solegalli ,

It is mostly useful when we generate lot of feature, most of the time using automated package / library to try many approachs. And sometimes, package will select variables or don't generate all features if certain conditions are met (and it will be different in train and test).

Another useful situation is when using an API that sometime do not return any results. It can be patched by hand (if no result then create a column with only empty values) or using this transformers.

I understand that many people do not encounter this type of problems, but I used before so I know it can be useful for some people.

solegalli:
Hi @thibaultbl very interesting thank you. 

Could I ask which are the libraries you refer to that have automated feature selection? are they open source or in house? and which API are you referring to? Would be really helpful if you could share the names or links :)

Also, I would like to pick your brain on another issue, that could be related to this one, if you don't mind.

thibaultbl:
Sorry, I don't want to talk about the librairies or API I am using because it is related to a private project, but I will try to find librairies or API that can be related to the case I am talking about (it may take some time).

solegalli:
@thibaultbl no problem at all. 

I am trying to understand how widespread would be the use of a transformer like the one your are suggesting in this PR,

i.e., does it apply to / would it be use by, a specific organisation? or does it tackle problems generated by other open-source APIs or libraries and thus would be an issue encountered by most data scientists?

then we can decide if it we should make it part of Feature-engine and also, in which module should we place it. It sounds to me that this is more data cleaning than selecting the more predictive features, which is what the module feature_selection is for.

thibaultbl:
Yes, I am not sure that feature_selection is the right module either.

I just though of an example, querying a document-oriented database with many features can lead to dataset being different in train and test.

banbiossa:
After making something similar for myself (I've called it `RememberFeatures` and put it in feature selection too), I came to see that it had been already done :)

In my use case, we do a lot of `predict than optimize` type of problems. So the model will be trained, the data will be post-processed according to `scenarios`. 
- e.g., "what would have happened if this person had received a campaign email"
  -  -> "receive_campaign" column will be turned into 0 -> 1
A lot of meta data e.g., `scenario_number`, `time_scenario_generated` will be added to the data during the process, and then passed over to the model. 
It is possible to remove the features (columns) before passing to the model, but it's easier and the pipeline code is cleaner if we use this type of transformer.

```py
# without the transformer
df_processed = process_df(df)
additional_columns = get_column_diff(df_processed, df)
df_processed_dropped = df_processed.drop(*additional_columns)
pipeline.predict(df_processed_dropped)
```

```py
# with the transformer
df_processed = process_df(df)
pipeline.predict(df_processed)  # deletion is done within the pipeline
```

Hope this helps with the discussion :)

solegalli:
@banbiossa thanks for the input!

I am happy to move forward with this class if you guys think it is useful. 

For the time being, I think feature selection is the right folder.

But the transformer in this PR needs a lot of refactory. We need to use the same name conventions that we use for other transformers. Add some tests to the parameter inputs. Add docstrings. And add some examples of its use in the rst files.

Also this PR is a bit old now, so we need to rebase master as well :p

Anyone up for it?

thibaultbl:
I can try to modify my PR if you think it is useful.

It is a bit old now, so maybe it is better to create an entirely new one

solegalli:
> 
> 
> I can try to modify my PR if you think it is useful.
> 
> It is a bit old now, so maybe it is better to create an entirely new one, what do you think ?

Hey @thibaultbl 

Thank you!

We could either make a new PR, or rebase master and continue here. That should be fairly straightforward. But, I don't mind. Whatever makes your life easier.

Before making the PR, could we define very clearly what the transformer does either here or in an issue (whatever you prefer)?

For example, taken from your first comment:

1.     If a column is in train (fitted dataset) but not in test, then the column will be created in test dataset with missing value only.
2.     If a column is in test but not in train, it will be dropped.

Is creating a column with missing values the best option? would it not be better to create it with zeroes? pros or cons of each approach? should we give the user the option (probable the best)? If the feature is categorical, should we also add zeroes? or is there a better strategy? 

@thibaultbl @banbiossa what other things should this transformer include/return from your experience?

We also need to think about a name that quickly suggests what the transformer does. I am not a great fun of SimilarColumns. MatchFeatures? I am not very creative sorry.

And also, we need to decide if it fits in the Selection module, or if we need a complete new module for this transformer (cleaning?). I would be keen to hear ideas / suggestions with pros and cons.

Thank you!

@KishManani you also create a lot of features, would you need something like this?

feature_engine/feature_selection.py
```diff
@@ -88,6 +89,67 @@ def transform(self, X):
...
+
+        """
+        # check input dataframe
+        X = _is_dataframe(X)
+
+        self.col = X.columns
+        return self
+
+    def transform(self, X: pd.DataFrame, **transform_params) -> pd.DataFrame:
+        """  Drops the variable that are not in the fited dataframe and returns a new dataframe with the remaining subset of variables.
```
KishManani, line 124: typo - fitted*

KishManani:
> > I can try to modify my PR if you think it is useful.
> > It is a bit old now, so maybe it is better to create an entirely new one, what do you think ?
> 
> Hey @thibaultbl
> 
> Thank you!
> 
> We could either make a new PR, or rebase master and continue here. That should be fairly straightforward. But, I don't mind. Whatever makes your life easier.
> 
> Before making the PR, could we define very clearly what the transformer does either here or in an issue (whatever you prefer)?
> 
> For example, taken from your first comment:
> 
> * If a column is in train (fitted dataset) but not in test, then the column will be created in test dataset with missing value only.
> * If a column is in test but not in train, it will be dropped.
> 
> Is creating a column with missing values the best option? would it not be better to create it with zeroes? pros or cons of each approach? should we give the user the option (probable the best)? If the feature is categorical, should we also add zeroes? or is there a better strategy?
> 
> @thibaultbl @banbiossa what other things should this transformer include/return from your experience?
> 
> We also need to think about a name that quickly suggests what the transformer does. I am not a great fun of SimilarColumns. MatchFeatures? I am not very creative sorry.
> 
> And also, we need to decide if it fits in the Selection module, or if we need a complete new module for this transformer (cleaning?). I would be keen to hear ideas / suggestions with pros and cons.
> 
> Thank you!
> 
> @KishManani you also create a lot of features, would you need something like this?

As far as I can tell this is primarily about data pre-processing to get your data in a format relevant for your model to train and predict. For time series forecasting you need to dynamically generate features a lot from raw data during backtesting into your train data, then test is typically a scenario where some feature values require an assumption. However, you have to specify a sensible value for each of those "future" feature values. I don't believe this use case requires the above functionality (which as far as I can tell is about doing a diff on the columns between the train and test, followed by adding or removing columns depending on what you want).

thibaultbl:
@solegalli  Are you OK with us overwriting the transform method of BaseSelector, I saw that all class that inherit from BaseSelector only use super().transform() for their transform method.

One of the secondary goal of this transformer would be to assure that the order of columns is similar in train and test, which is not possible if we use only the super().transform().

solegalli:
All selector transformers in feture 

> 
> 
> @solegalli Are you OK with us overwriting the transform method of BaseSelector, I saw that all class that inherit from BaseSelector only use super().transform() for their transform method.
> 
> One of the secondary goal of this transformer would be to assure that the order of columns is similar in train and test, which is not possible if we use only the super().transform().

That will break the functionality of all the selector transformers already available in feature engine. So no, we can't modify that base class.

thibaultbl:
> All selector transformers in feture
> 
> > @solegalli Are you OK with us overwriting the transform method of BaseSelector, I saw that all class that inherit from BaseSelector only use super().transform() for their transform method.
> > One of the secondary goal of this transformer would be to assure that the order of columns is similar in train and test, which is not possible if we use only the super().transform().
> 
> That will break the functionality of all the selector transformers already available in feature engine. So no, we can't modify that base class.
> 
> I also don't think this new transformer belongs to the selection module. It should go momentarily to a new module called perhaps "preprocessing"

I didn't want to modify the base class, but I wanted to overwrite it in my own class. Anyway, I agree with you, it should not be in selection folder, so the debate is meaningless now. 
But I think "preprocessing" is not a good name, it should be a name saying that this folder contains transformers useful to ensure that test and train are similar, someting like "sanity_check" ?

solegalli:
> 
> 
> > I can try to modify my PR if you think it is useful.
> > It is a bit old now, so maybe it is better to create an entirely new one, what do you think ?
> 
> Hey @thibaultbl
> 
> Thank you!
> 
> We could either make a new PR, or rebase master and continue here. That should be fairly straightforward. But, I don't mind. Whatever makes your life easier.
> 
> Before making the PR, could we define very clearly what the transformer does either here or in an issue (whatever you prefer)?
> 
> For example, taken from your first comment:
> 
>     * If a column is in train (fitted dataset) but not in test, then the column will be created in test dataset with missing value only.
> 
>     * If a column is in test but not in train, it will be dropped.
> 
> 
> Is creating a column with missing values the best option? would it not be better to create it with zeroes? pros or cons of each approach? should we give the user the option (probable the best)? If the feature is categorical, should we also add zeroes? or is there a better strategy?
> 
> @thibaultbl @banbiossa what other things should this transformer include/return from your experience?
> 
> We also need to think about a name that quickly suggests what the transformer does. I am not a great fun of SimilarColumns. MatchFeatures? I am not very creative sorry.
> 
> And also, we need to decide if it fits in the Selection module, or if we need a complete new module for this transformer (cleaning?). I would be keen to hear ideas / suggestions with pros and cons.
> 
> Thank you!
> 
> @KishManani you also create a lot of features, would you need something like this?

@TremaMiguel  could I gather your thoughts on this transformer? is this something you could use? have you come across this problem of different columns in train and test? what functionality you think might be needed in the transformer?

thibaultbl:
@solegalli what is your thought on my implementation ?

feature_engine/sanity_check/similar_columns.py
```diff
@@ -0,0 +1,68 @@
...
+    Parameters
+    ----
+    None
+
+    """
+
+    def __init__(self, impute_with: Any = np.nan):
+        self.impute_with = impute_with
+        self.col: List[str] = []
+
```
solegalli, line 21: We don't need to define this attribute in the init, not per the sklearn / feature_engine conventions

feature_engine/sanity_check/similar_columns.py
```diff
@@ -0,0 +1,68 @@
...
+    def fit(self, df: pd.DataFrame, y: pd.Series = None, **fit_params):
+        """Fit columns schema
```
solegalli, line 23: why **fit_params*?

feature_engine/sanity_check/similar_columns.py
```diff
@@ -0,0 +1,68 @@
...
+        ----------
+
+        X: pandas dataframe of shape = [n_samples, n_features]
+            The input dataframe
+
+        y: None
+            y is not needed for this transformer. You can pass y or None.
+
+        """
+
```
solegalli, line 35: Please not that our other transformers check that the input is a dataframe, that check should be made part of the transformer.

Also, is this transformer going to allow missing data in the train set? could be yes, could be no, could be let to the user to decide. Please check at some other transformers in the library to see how we give users the option on what to do with NA

feature_engine/sanity_check/similar_columns.py
```diff
@@ -0,0 +1,68 @@
...
+        self.col = df.columns
```
solegalli, line 36: all parameters learned in fit, should have a trailing underscore as per sklearn convetion. Also note that we use variables_ and not cols in feature engine

feature_engine/sanity_check/similar_columns.py
```diff
@@ -0,0 +1,68 @@
...
+        -------
+        X_transformed: pandas dataframe of shape =
+             [n_samples, n_features - len(features_to_drop)]
+
+                The transformed dataframe with the same columns
+                (in same order) than the fitted dataframe
+
+        """
+
+        for col in self.col:
```
solegalli, line 63: is this the fastest possible implementation? can we not find the different variables and add self.impute_with all at once?

feature_engine/sanity_check/similar_columns.py
```diff
@@ -0,0 +1,68 @@
...
+        Returns
+        -------
+        X_transformed: pandas dataframe of shape =
+             [n_samples, n_features - len(features_to_drop)]
+
+                The transformed dataframe with the same columns
+                (in same order) than the fitted dataframe
+
+        """
+
```
solegalli, line 62: please have a look at some of the other transformers in the library. We need to check that the input is a df, and if it contains na or not, depending on whether we allow the user to decide. And we also need to check that the transformer has been fitted.

solegalli (commented):
Hi @thibaultbl, thanks for the contribution. This makes a good start. 

We need to make the transformer compatible with sklearn and feature_engine conventions. 

Please have a look at other transformers in the Feature_engine to bring in the checks that are missing.

Also, as per a previous comment on this PR, we need a series of bullets describing what exactly this transformer needs to do. So we can have a clear view of the end product, before starting with the code.

I also think we should log somehow which variables are being ignored / removed from the dataset during transform, otherwise I find it too dangerous.

thibaultbl:
The objectives of this transformer are the following:

- If a DataFrame in transform have more columns that the dataframe used for fit, those columns are dropped
- If a DataFrame in transform have less columns that the dataframe used for fit, those columns are added with missing values (default) or any value specified by the user.

You said:
> I also think we should log somehow which variables are being ignored / removed from the dataset during transform, otherwise I find it too dangerous.

How do you want to do that ? Furthermore, it doesn't seems to be the case in the others transformers that drop columns, like all transformers in the "selection" directory.

solegalli:
@thibaultbl 

Regarding logging the transformations, I am thinking of having a parameter called verbose=True in the init. And if verbose is True, then print a warning indicating which columns were created with NA or other values, and a second warning indicating which columns were removed.

If the model is deployed in production, I guess the users could use this warnings and store them in their logs?

if verbose = False then we don't raise any warning.

solegalli:
Regarding the selectors in the selection module, they all contain an attribute called features_to_drop_, which stores the names of the features that will be dropped, and these same features will be dropped in every single dataframe passed to transform.

But with this new SimilarColumns transformers, the features to drop will vary depending on the df that the user passes to transform. So, the user is left a bit in the darkness as to what was dropped.

thibaultbl:
> @thibaultbl
> 
> Regarding logging the transformations, I am thinking of having a parameter called verbose=True in the init. And if verbose is True, then print a warning indicating which columns were created with NA or other values, and a second warning indicating which columns were removed.
> 
> If the model is deployed in production, I guess the users could use this warnings and store them in their logs?
> 
> if verbose = False then we don't raise any warning.

Understood, I added a logging.warning if verbose is set to True

feature_engine/sanity_check/similar_columns.py
```diff
@@ -0,0 +1,162 @@
...
+        if not self.drop_if_more_columns and len(_columns_to_drop) > 0:
+            raise ValueError(
+                "drop_if_more_columns is set to False"
+                " and transform df contain more columns "
+                "than the df used in fit "
+                f"(found {set(X.columns)}, "
+                f"expected ({set(self.variables_)}))."
+            )
+
+        if not self.add_if_less_columns and len(_columns_to_add) > 0:
```
solegalli, line 97: same as per previous comment.

In my mind, if we allow the user to decide not to add the missing columns, then we should allow this functionality. But I may be understanding the logic wrongly. Please let me know your thoughts

feature_engine/sanity_check/similar_columns.py
```diff
@@ -0,0 +1,162 @@
...
+                The transformed dataframe with the same columns
+                (in same order) than the fitted dataframe
+
+        """
+        check_is_fitted(self)
+
+        X = self._check_input(X)
+
+        _columns_to_drop = list(set(X.columns) - set(self.variables_))
+        _columns_to_add = list(set(self.variables_) - set(X.columns))
```
solegalli, line 134: why the use of `set`?

tests/test_sanity_check/test_similar_columns.py
```diff
@@ -0,0 +1,78 @@
...
+            "Name": ["tom", "nick", "krish", "jack"],
+            "City": ["London", "Manchester", "Liverpool", "Bristol"],
+            "Studies": [np.nan, np.nan, np.nan, np.nan],
+            "Age": [20, 21, 19, 18],
+            "Marks": [0.9, 0.8, 0.7, 0.6],
+            "dob": pd.date_range("2020-02-24", periods=4, freq="T"),
+        }
+    )
+
+    pd.testing.assert_frame_equal(expected_result, transformed_df)
```
solegalli, line 30: in addition to testing the final output, we do tend to check that the attributes are learned correctly in the transfomer, for example that self.variables_ is what we expect it to be. Please have a look at the tests for other transformers in the package.

feature_engine/sanity_check/similar_columns.py
```diff
@@ -0,0 +1,162 @@
...
+        X: pandas dataframe of shape = [n_samples, n_features]
+            The input dataframe
+
+        y: None
+            y is not needed for this transformer. You can pass y or None.
+
+        """
+        X = self._check_input(X)
+
+        self.variables_ = X.columns
```
solegalli, line 78: the fit method needs to learn a parameter called n_features_in_ with the number of columns in the train set. This is for convention with sklearn, and to pass the sklearn check_estimator checks. 

In this particular transformer, this parameter is a bit useless, but probably the checks will not pass if we do not add it.

solegalli (commented):
Hi @thibaultbl 

Great work on adding all the checks and expanding the functionality of this transformer. Thank you very much!!!

Please see my comments throughout the code for further discussion.

Also, in the test file, we also test that the right errors are raised when the user enters the wrong parameters, and that an error is raised if the user uses transform and did not use fit before.

feature_engine/sanity_check/similar_columns.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def _raise_errors_if_needed(
+        self,
+        _columns_to_drop: List[str],
+        _columns_to_add: List[str],
+        X: pd.DataFrame
+    ):
+
+        if not self.drop_if_more_columns and len(_columns_to_drop) > 0:
+            raise ValueError(
```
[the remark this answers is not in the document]
thibaultbl: I think it should stay an error, sklearn transformer have check that raise errors if their is extra columns in transform df than in fit df.
Yes the transformer doesn't work if the User select False and there is extra columns, but it works if the user select False and the number of columns is equal or inferior (if User select true for  add_if_less_columns) to fit df.

feature_engine/sanity_check/similar_columns.py
thibaultbl: Why not, I thought it was the appropriate data structure because we want a finite, unique list of elements (and list are not unique) ?

feature_engine/sanity_check/similar_columns.py
thibaultbl: Ok for n_features_in, but to my understanding, our transformer will never pass sklearn check_estimator because it cannot apply to numpy array (only pandas dataframe, because it need columns name)

solegalli: > 
> 
> Ok for n_features_in, but to my understanding, our transformer will never pass sklearn check_estimator because it cannot apply to numpy array (only pandas dataframe, because it need columns name) .
> 
> All checks passed.

There is a workaround in the function is_dataframe to allow numpy arrays solely to be able to pass the check_estimator tests. See [this](https://github.com/solegalli/feature_engine/blob/master/feature_engine/dataframe_checks.py#L36).

And, we need to add a check_estimator test for this class as well. See [this as an example](https://github.com/solegalli/feature_engine/blob/master/tests/test_transformation/test_check_estimator_transformers.py).

feature_engine/sanity_check/similar_columns.py
solegalli: > 
> 
> Why not, I thought it was the appropriate data structure because we want a finite, unique list of elements (and list are not unique) ?

If we use set on X.columns, we are in essence, expecting duplicated column names  in the dataframe. Am I correct? 

This shouldn't happen, and if it does, I don't think we should have functionality to resolve it. It is up to the user to check that their dataframes make sense.

I would be happy to use set if it makes things faster, but not otherwise.

feature_engine/sanity_check/similar_columns.py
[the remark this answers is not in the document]
solegalli: @NicoGalli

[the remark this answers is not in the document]
NicoGalli: @solegalli @thibaultbl I would suggest to keep it as simple as possible, at least at the bigining. If it is unclear this validation is needed or not, I would suggest not to include it right now and simplify the amount of paraments for the users to implement this Class.  If you guys would like me to dig deeper, just let me know.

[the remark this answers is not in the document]
solegalli: I agree with @NicoGalli 

Thinking it further, the whole point of this new transformer is to make coming dataframes match the variables of the train set, so that is what the functionality should be.

Allowing the user to not add or not remove some variables, defeats the purpose of the transformer imho.

solegalli:
Hi @thibaultbl 

I see that you added a function to check for complex values in the last commit. There are a couple of tests (including check for complex) that will inevitable fail, because of how Feature-engine was designed. Those checks that fail can be found [here](https://github.com/solegalli/feature_engine/blob/master/feature_engine/validation.py) and we skip them from the check_estimator by adding a tag like for example we do [here](https://github.com/solegalli/feature_engine/blob/master/feature_engine/base_transformers.py#L145) in the class.

Maybe that helps. 

If you need some support please let us know. Also let us know when we should review this again.

If you want, not to waste effort, let us know what would would like to add or remove before actually coding. Just a suggestion.

Thank you

thibaultbl:
> Hi @thibaultbl
> 
> I see that you added a function to check for complex values in the last commit. There are a couple of tests (including check for complex) that will inevitable fail, because of how Feature-engine was designed. Those checks that fail can be found [here](https://github.com/solegalli/feature_engine/blob/master/feature_engine/validation.py) and we skip them from the check_estimator by adding a tag like for example we do [here](https://github.com/solegalli/feature_engine/blob/master/feature_engine/base_transformers.py#L145) in the class.
> 
> Maybe that helps.
> 
> If you need some support please let us know. Also let us know when we should review this again.
> 
> If you want, not to waste effort, let us know what would would like to add or remove before actually coding. Just a suggestion.
> 
> Thank you

Tanks, it is useful.

Nevertheless, this transformer will not pass _check_estimator (even with the tests that are allowed to fail), because one of the _check_estimator test is to ensure that an error is raised when number of features in transform is different than in fit, which is the main and only objective of this transformer.  look around line 1371 in [https://github.com/scikit-learn/scikit-learn/blob/2beed5584/sklearn/utils/estimator_checks.py#L486](https://github.com/scikit-learn/scikit-learn/blob/2beed5584/sklearn/utils/estimator_checks.py#L486)

What is your thoughts on it ? Can we skip this test for this transformer ? Otherwise, we can stop working on it.

solegalli:
Yes, we can skip that test as well. If any other test fails, let me know and we can have a look together, or skip it and then I review and we take it from there.

Cheers

solegalli:
Hi @thibaultbl thanks for the additional commits to the PR.

Are you still working on this? would you have time to address the other comments that I left throughout the code?

thibaultbl:
Hi @solegalli ,

I think most if not all comments have been treated now.

thibaultbl:
Hi @solegalli 

I think most, if not all comments have been treated now.

Nevertheless, I still use "set" in transform because it still seems like the better solution for me, but I am open to suggestion.

feature_engine/sanity_check/__init__.py
```diff
@@ -0,0 +1,9 @@
+"""
```
solegalli, line 1: pls rename module to "preprocessing"

feature_engine/sanity_check/similar_columns.py
```diff
@@ -0,0 +1,125 @@
+from typing import Any
```
solegalli, line 1: pls rename script to "match_columns.py"

feature_engine/sanity_check/similar_columns.py
```diff
@@ -0,0 +1,125 @@
...
+
+from feature_engine.dataframe_checks import (
+    _check_contains_inf,
+    _check_contains_na,
+    _is_dataframe,
+)
+from feature_engine.variable_manipulation import _find_all_variables
+
+
+class SimilarColumns(BaseEstimator, TransformerMixin):
```
solegalli, line 18: pls rename class to "MatchColumnsToTrainSet"

feature_engine/sanity_check/similar_columns.py
```diff
@@ -0,0 +1,125 @@
...
+    """Ensure that similar columns are in test and train dataset.
```
solegalli, line 19: Pls expand docstrings. You can have a look at other transformers to see what parts are required.

feature_engine/sanity_check/similar_columns.py
```diff
@@ -0,0 +1,125 @@
...
+
+    Parameters
+    ----
+    None
+
+    """
+
+    def __init__(
+        self,
+        fill_value: Any = np.nan,
```
solegalli, line 29: fill_value should take np.nan, integers or strings. Not Any. Could you please fix?

feature_engine/sanity_check/similar_columns.py
```diff
@@ -0,0 +1,125 @@
...
+        missing_values: str = "raise",
+        verbose: bool = True
+    ):
+
+        if missing_values not in ["raise", "ignore"]:
+            raise ValueError("missing_values takes only values 'raise' or 'ignore'.")
+
+        if not isinstance(verbose, bool):
+            raise ValueError("verbose takes only booleans True and False")
+
```
solegalli, line 39: could you pls add a check that fill_value takes the expected input format?

feature_engine/sanity_check/similar_columns.py
```diff
@@ -0,0 +1,125 @@
...
+            y is not needed for this transformer. You can pass y or None.
+
+        """
+        X = _is_dataframe(X)
+
+        self.n_features_in_ = X.shape[1]
+
+        self.variables_ = _find_all_variables(X, list(X.columns))
+
+        X = self._check_input(X)
```
solegalli, line 73: this should be at the top of fit() instead of line 67.

feature_engine/sanity_check/similar_columns.py
```diff
@@ -0,0 +1,125 @@
...
+        ----------
+
+        X: pandas dataframe of shape = [n_samples, n_features]
+            The input dataframe
+
+        y: None
+            y is not needed for this transformer. You can pass y or None.
+
+        """
+        X = _is_dataframe(X)
```
solegalli, line 67: pls replace by X = self._check_input(X)

feature_engine/sanity_check/similar_columns.py
```diff
@@ -0,0 +1,125 @@
...
+
+        self.n_features_in_ = X.shape[1]
```
solegalli, line 69: this should be the last line of fit()

feature_engine/sanity_check/similar_columns.py
```diff
@@ -0,0 +1,125 @@
...
+
+        self.variables_ = _find_all_variables(X, list(X.columns))
```
solegalli, line 71: self.variables_ = list(X.columns)

We don't need _find_alll_variables here

feature_engine/sanity_check/similar_columns.py
```diff
@@ -0,0 +1,125 @@
...
+
+        """
+        check_is_fitted(self)
+
+        X = self._check_input(X)
+
+        _columns_to_drop = list(set(X.columns) - set(self.variables_))
+        _columns_to_add = list(set(self.variables_) - set(X.columns))
+
+        X = X.reindex(
```
solegalli, line 107: why do we need reindex? can we not use pandas.align?

feature_engine/sanity_check/similar_columns.py
```diff
@@ -0,0 +1,125 @@
...
+                The transformed dataframe with the same columns
+                (in same order) than the fitted dataframe
+
+        """
+        check_is_fitted(self)
+
+        X = self._check_input(X)
+
+        _columns_to_drop = list(set(X.columns) - set(self.variables_))
+        _columns_to_add = list(set(self.variables_) - set(X.columns))
```
solegalli, line 105: I don't think we need this line

feature_engine/sanity_check/similar_columns.py
```diff
@@ -0,0 +1,125 @@
...
+
+        X = X.reindex(
+                columns=list(X.columns) + _columns_to_add,
+                fill_value=self.fill_value
+            )
+
+        X = X.drop(_columns_to_drop, axis=1)
```
solegalli, line 112: I think we should drop first, and then add the remaining columns

feature_engine/sanity_check/similar_columns.py
```diff
@@ -0,0 +1,125 @@
...
+        """
+        check_is_fitted(self)
+
+        X = self._check_input(X)
+
+        _columns_to_drop = list(set(X.columns) - set(self.variables_))
+        _columns_to_add = list(set(self.variables_) - set(X.columns))
+
+        X = X.reindex(
+                columns=list(X.columns) + _columns_to_add,
```
solegalli, line 108: instead of this, should we not pass just `self.variables_` ?

feature_engine/sanity_check/similar_columns.py
```diff
@@ -0,0 +1,125 @@
...
+                fill_value=self.fill_value
+            )
+
+        X = X.drop(_columns_to_drop, axis=1)
+
+        # reorder columns
+        X = X.loc[:, self.variables_]
```
solegalli, line 115: if we drop first and add the remaining columns later, do we need this line of code?

solegalli (commented):
Hey @thibaultbl thank you for the changes. We are getting there.

Pls have a look at the comments throughout the code :)

feature_engine/sanity_check/similar_columns.py
```diff
@@ -0,0 +1,125 @@
...
+    ----
+    None
+
+    """
+
+    def __init__(
+        self,
+        fill_value: Any = np.nan,
+        missing_values: str = "raise",
+        verbose: bool = True
```
solegalli, line 31: the current implementation does not use this parameter. It should be used in transform. If true. print out which columns were dropped from the data, and which ones were added and the fill value added

feature_engine/sanity_check/similar_columns.py
thibaultbl: We cannot do that, because self._check_input(X) need self._variables to be define.

feature_engine/sanity_check/similar_columns.py
thibaultbl: We cannot do that, because self._check_input(X) need self._variables to be define.

feature_engine/sanity_check/similar_columns.py
thibaultbl: I think it can raise errors in "_find_all_variables" if X is not a dataframe

feature_engine/sanity_check/similar_columns.py
thibaultbl: You want to remove this line and put the code directly when needed ? For readability, I am strongly in favor of the solution with intermediates variables, but if you really want it, I can modify the code.

feature_engine/sanity_check/similar_columns.py
thibaultbl: I agree, I corrected the code.

solegalli:
Hi @thibaultbl 

Thanks for the code changes!! 

and apologies for the delayed response. I am only back from holidays today.

I made a few code changes and a PR to your repo. If you merge, that will automatically update the code here as well.

The transformer is almost ready. To finalize it, we need to expand the tests to examples were more than 1 variable are removed or added, and when both things happen at the same time. And also test that the print statements are correct (the right variable names are printed out) if verbose is true.

Would you mind having a go at those?

solegalli:
we also need to complete the docs:

- [ ] add the transformer to the docs/index.rst
- [ ] create a module called preprocessing and add an example of use
- [ ] add the transformer to the readme

thibaultbl:
Hi @solegalli,

I am back from holidays as well.

I will try to have a look this week, next week will be busy for me.

thibaultbl:
Hi @solegalli ,

I pushed all required changes (expanded tests for multiple features, testing verbose output, added notebook for examples), but cannot create the docs.

I am not really familiar with Sphinx, can you look at it ?

solegalli (commented):
Hi @thibaultbl 

Thank you so much for the changes, we are almost there! I few tiny requests.

1) I made some code changes and a PR to your repo

2) I think the name MatchColumnsToTrainSet is a bit of a mouthful (I know I made it, apologies). Can we change it to MatchVariables? I included this in the PR already

3) the "example" folder with the notebooks were moved out of feature-engine's library. They now live here: https://github.com/feature-engine/feature-engine-examples

Could you move this file to that repo in a different PR and reference this PR please?

Note that I made changes to the notebook that you will find in your PR, so please move the notebook included in the PR with the changes :)

solegalli:
@thibaultbl 

FYI:
https://github.com/thibaultbl/feature_engine/pull/2

docs/index.rst
```diff
@@ -237,7 +242,9 @@ The `issues <https://github.com/solegalli/feature_engine/issues/>`_ and
    outliers/index
    creation/index
    selection/index
+   preprocessing/index
    wrappers/index
+   preprocessing/index
```
solegalli, line 248: this entry is duplicated and it is causing the error in the tests:

/home/circleci/project/docs/index.rst:234:duplicated entry found in toctree: preprocessing/index

thibaultbl:
https://github.com/feature-engine/feature-engine-examples/pull/1

thibaultbl:
I think everything is done, I deleted the "examples" folder, and move the examples to https://github.com/feature-engine/feature-engine-examples/pull/1

solegalli:
Hi @thibaultbl 

Thank you so much for the code changes! The codebase is certainly finished. 

There is a tiny issue though, in the PR, it seems like it would like to delete all the files within the examples folder, but the example folder does not exist any more in the master branch.

I wonder if this would be resolved if you pulled the master in this repo to the master in your local, and then rebased master onto the branch from which you are doing this PR?

thibaultbl:
Hi @solegalli 

I don't know what is happening. 

If I create a new PR, everything is okay, but in this one we can see many changes that should not be here. I think the feature_engine:main is an old version, but I don't know why and I don't know how to resolve the problem at the moment.

I will try to resolve it as soon as possible.

solegalli:
Hi @thibaultbl 

Did you rebase master into your branch? or did you merge it?

Nevertheless, don't worry. I will merge #308 and close this PR automatically from there :)

Thanks a lot for the end-to-end contribution, including suggestion of the new transformer, engagement in the discussions and the code. We really appreciate it. And would be very happy to work with you again.

Have a nice week.

PR #140, opened 2020-10-02
Add drop correlated

thibaultbl:
Added automatic columns drop when feature are too correlated (Pearson correlation, easy to add support for others correlations later)

### DIFF
--- a/feature_engine/feature_selection.py
+++ b/feature_engine/feature_selection.py
@@ -218,17 +218,17 @@ class DropCorrelated(BaseEstimator, TransformerMixin):
     """
     def __init__(self, cols=None, threshold=0.9):
         self.cols = cols
-        self.threshold=threshold
-        self.keeped_cols=[]
+        self.threshold = threshold
+        self.keeped_cols = []
 
     def fit(self, df, y=None, **fit_params):
         if self.cols is None:
             self.cols = list(df.select_dtypes(include=['int','float']).columns)
         c = df.loc[:, self.cols].corr().abs().unstack()
-        s= c[c>self.threshold]
-        so=s[s.index.get_level_values(0)!=s.index.get_level_values(1)]
-        self.cols=list(set(so.drop_duplicates().index.get_level_values(0)))
-        self.keeped_cols=list(set(df.columns).difference(self.cols))
+        s = c[c>self.threshold]
+        so = s[s.index.get_level_values(0)!=s.index.get_level_values(1)]
+        self.cols = list(set(so.drop_duplicates().index.get_level_values(0)))
+        self.keeped_cols = list(set(df.columns).difference(self.cols))
         return self
 
     def transform(self, df, **transform_params):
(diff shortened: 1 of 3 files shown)


solegalli (changes requested):
Hi @thibaultbl Thank you so much for the contribution.

For this transformer, I was thinking it could remove features based of 2 approaches:

1) brute_force: where the transformer removes correlated features on first come first served basis, without any further insight

2) intelligent: where the transformer first finds groups of correlated features, and then from each group it retains features based on some characteristic, for example, it could retain the feature with less number of missing values. Or alternatively, build a ML model on the group of correlated features, and retain the one with the highest feature importance. (we need to think how to code this logic)

The code that inspires this transformer lives here:
https://github.com/solegalli/feature-selection-for-machine-learning/blob/master/04-Correlation/04.1-Correlation-Pearson.ipynb

Would you be up to give it a go?

If yes, there are a few things to consider:

- please checkout the **develop branch** and make the PR to that branch (I am refactoring Feature-engine's modules into submodules, so master is quite behind the current view of the package)

- please check how other transformers within Feature-engine library work, and try to incorporate that logic into the DropCorrelated. For example:

- we have a function in the module variable_manipulation.py that checks that the user enters the values correctly, which should be used in the init of each transformer.

- the fit method should take only params X and y, otherwise it may not be compatible with the sklearn pipeline.

- if user can enter variables, the parameter name we use throughout Feature-engine is `variables`.

- the transform method should only take X, same reason as fit.

- in fit and transform there are dataframe checks methods, that should be incorporated. There is also a check to corroborate that the transformer was fit.

You will get a better view of these, if you take a look at other transformers, for example the drop constant features, or check one of the imputers as well.

- in the init method, allow the argument brute_force=true, to let the user decide if he wants to do brute force, or the intelligent one.

Also, when you make the PR, please reference the issue you are addressing, so we can better keep track, and also helps other contributors understand which issue is being worked on, and which one isn't.

And the final suggestion, let's make it so that the user can decide if they want pearson, spearman or kendall :)

If you are up for this changes, I think the best is to work on a new PR?

Thanks a lot!

solegalli:
@thibaultbl on second thoughts, I think it would be better to have 2 correlation selectors, one with the brute force approach, and one with the intelligent approach. It would be easier for both  users and developers I think. Thoughts?

thibaultbl:
@solegalli Understood, I will works on it !

solegalli:
@thibaultbl the code to drop duplicated features that can be found here:
https://github.com/solegalli/feature_engine/blob/develop/feature_engine/selection/drop_duplicate_features.py 

may be useful for the DropCorrelatedFeatures, version brute force. 

Would you be able to give it a go some time this week? 

If yes, please checkout and and make a PR to the develop branch. I'll let you know when we are back on master. 
Thanks a lot!

thibaultbl:
Hi @solegalli ,

Thanks for the insight.

I don't think I will have time to works on it this week.

PR #145, opened 2020-10-04
Adding docstrings and type hints

Okroshiashvili:
I add docstrings, type hints, and make some stylistic modifications.

### DIFF
--- a/feature_engine/dataframe_checks.py
+++ b/feature_engine/dataframe_checks.py
@@ -2,25 +2,64 @@
 # License: BSD 3 clause
 
 import pandas as pd
+from typing import Union, List
 
-def _is_dataframe(X):
-    # checks if the input is a dataframe. Also creates a copy,
-    # important not to transform the original dataset.
+def _is_dataframe(X: pd.DataFrame) -> pd.DataFrame:
+    """
+    Checks if the input is a DataFrame.
+    Creates the copy of the initial DataFrame.
+
+    Args:
+        X: Argument to perform check against
+
+    Raises:
+        TypeError: If the input is not the Pandas DataFrame
+
+    Returns:
+        The copy of initial DataFrame.
+        Important not to transform the original dataset.
+    """
+
     if not isinstance(X, pd.DataFrame):
         raise TypeError("The data set should be a pandas dataframe")
+
     return X.copy()
 
 
-def _check_input_matches_training_df(X, reference):
-    # check that dataframe to transform has the same number of columns
-    # that the dataframe used during fit method
+def _check_input_matches_training_df(X: pd.DataFrame, reference: int) -> None:
+    """
+    Check that DataFrame to transform has the same number
+    of columns that the DataFrame used during fit method.
+
+    Args:
+        X: Pandas DataFrame to perform comparison
+        reference: Number of columns
+
+    Raises:
+        ValueError: If number of columns is not the same
+
+    Returns:
+        None
+    """
+
     if X.shape[1] != reference:
         raise ValueError('The number of columns in this data set is different from that of the train set used during'
                          'the fit method')
+
     return None
 
 
-def _check_contains_na(X, variables):
+def _check_contains_na(X: pd.DataFrame, variables: Union[str, List[str]]):
+    """
+    Checks if DataFrame columns contains null values.
+
+    Args:
+        X: Pandas DataFrame to perform check against
+        variables: List of variables for check for null values
+
+    Raises:
+        ValueError: If variable(s) contain null values
+    """
     if X[variables].isnull().values.any():
         raise ValueError('Some of the variables to transform contain missing values. Check and remove those '
                          'before using this transformer.')
--- a/feature_engine/variable_manipulation.py
+++ b/feature_engine/variable_manipulation.py
@@ -2,20 +2,48 @@
 # License: BSD 3 clause
 # functions shared across transformers
 
+from typing import List, Optional, Union
 
-def _define_variables(variables):
-    # Check that variable names are passed in a list.
-    # Can take None as value
+import pandas as pd
+
+
+def _define_variables(variables: Union[Optional[str], List[str]]) -> Union[Optional[str], List[str]]:
+    """
+    Takes string or list of strings and checks if argument is list of strings.
+    Can take None as argument.
+
+    Args:
+        variables: string or list of strings
+
+    Returns:
+        List of strings
+    """
+
+    #! Can take Tuple of variables and return list of tuple. Is it accaptable?
     if not variables or isinstance(variables, list):
         variables = variables
     else:
         variables = [variables]
+
     return variables
 
 
-def _define_numerical_dict(the_dict):
-    # Check that the entered dictionary is indeed a dictionary of integers and floats
-    # Can take None as value
+def _define_numerical_dict(the_dict: Union[Optional[str], dict]) -> Union[Optional[str], dict]:
+    """
+    Takes dictionary and checks if all values in dictionary are integers and floats.
+    Can take None as argument.
+
+    Args:
+        the_dict: Dict to perform check against
+
+    Raises:
+        ValueError: If all values of dict are not int or float
+        TypeError: When argument type is not dict
+
+    Returns:
+        None or the dict
+    """
+
     if not the_dict:
         the_dict = the_dict
     elif isinstance(the_dict, dict):
@@ -23,12 +51,28 @@ def _define_numerical_dict(the_dict):
             raise ValueError('All values in the dictionary must be integer or float')
     else:
         raise TypeError('The parameter can only take a dictionary or None')
+
     return the_dict
 
 
-def _find_numerical_variables(X, variables=None):
-    # Find numerical variables in a data set or check that
-    # the variables entered by the user are numerical.
+def _find_numerical_variables(X: pd.DataFrame, variables: List[str]=None) -> List[str]:
+    """
+    Takes Pandas DataFrame along with list of colum names and
+    checks if these columns are numerical. If not provided columns,
+    checks if all of them are numerical.
+
+    Args:
+        X: DataFrame to perform the check against
+        variables: List of variables. Defaults to None.
+
+    Raises:
+        ValueError: If all variables are non-numerical type or DataFrame is empty
+        TypeError: If some of the variables are non-numerical
+
+    Returns:
+        List of variables
+    """
+
     if not variables:
         variables = list(X.select_dtypes(include='number').columns)
         if len(variables) == 0:
@@ -37,12 +81,27 @@ def _find_numerical_variables(X, variables=None):
         if len(X[variables].select_dtypes(exclude='number').columns) != 0:
             raise TypeError("Some of the variables are not numerical. Please cast them as numerical "
                             "before calling this transformer")
+
     return variables
 
 
-def _find_categorical_variables(X, variables=None):
-    # Find categorical variables in a data set or check that
-    # the variables entered by user are categorical.
+def _find_categorical_variables(X: pd.DataFrame, variables: List[str]=None) -> List[str]:
+    """
+    Takes Pandas DataFrame and finds all categorical variables if not provided.
+    If variables are provided, checks if they are indeed categorical.
+
+    Args:
+        X: DataFrame to perform the check against
+        variables: List of variables. Defaults to None.
+
+    Raises:
+        ValueError: If all variables are non-categorical type or DataFrame is empty
+        TypeError: If some of the variables are non-categorical
+
+    Returns:
+        List of variables
+    """
+
     if not variables:
         variables = list(X.select_dtypes(include='O').columns)
         if len(variables) == 0:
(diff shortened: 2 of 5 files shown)


feature_engine/dataframe_checks.py
```diff
@@ -2,25 +2,64 @@
...
     if X.shape[1] != reference:
         raise ValueError('The number of columns in this data set is different from that of the train set used during'
                          'the fit method')
+
     return None
 
 
-def _check_contains_na(X, variables):
+def _check_contains_na(X: pd.DataFrame, variables: Union[str, List[str]]):
+    """
```
solegalli, line 53: variables here should always be a list.

feature_engine/dataframe_checks.py
```diff
@@ -2,25 +2,64 @@
...
+    Checks if DataFrame columns contains null values.
+
```
solegalli, line 55: typo: contain

feature_engine/dataframe_checks.py
```diff
@@ -2,25 +2,64 @@
...
+    Args:
+        X: Pandas DataFrame to perform check against
+        variables: List of variables for check for null values
+
```
solegalli, line 59: List of variables **to** check for null values

feature_engine/variable_manipulation.py
```diff
@@ -2,33 +2,77 @@
...
+from typing import List, Optional, Union
 
-def _define_variables(variables):
-    # Check that variable names are passed in a list.
-    # Can take None as value
+import pandas as pd
+
+
+def _define_variables(variables: Union[None, List[str]]) -> Union[List[str], None]:
+    """
```
solegalli, line 11: _define_variables can take None, a string or a list of strings. Is this what the Union[None, List[str]] is saying?

feature_engine/variable_manipulation.py
```diff
@@ -2,33 +2,77 @@
...
+
+    Args:
+        variables: string or list of strings
+
+    Returns:
+        List of strings
+    """
+
+    #! Can take Tuple of variables and return list of tuple. Is it accaptable?
     if not variables or isinstance(variables, list):
```
solegalli, line 23: No, actually that is not acceptable. Thanks for picking it up. Could you introduce a fix?

feature_engine/variable_manipulation.py
```diff
@@ -2,33 +2,77 @@
...
         variables = variables
     else:
         variables = [variables]
+
```
solegalli, line 27: this else here is also assuming that the user passes a string. But if they passed a dictionary, I think it would also assign it to variables here?
Needs fixing.

feature_engine/variable_manipulation.py
```diff
@@ -2,33 +2,77 @@
...
     return variables
 
 
-def _define_numerical_dict(the_dict):
-    # Check that the entered dictionary is indeed a dictionary of integers and floats
-    # Can take None as value
+def _define_numerical_dict(the_dict: Union[Optional[str], dict]) -> Union[Optional[str], dict]:
+    """
```
solegalli, line 32: the_dict can only take None or a dictionary. Is this what the Union argument is saying? It shouldn't take a string.

also, is `the_dict` a good name for this param? I think we should use `dict_`.

I named it on a rush when we were working on this function.

feature_engine/variable_manipulation.py
```diff
@@ -2,33 +2,77 @@
...
+    Takes dictionary and checks if all values in dictionary are integers and floats.
+    Can take None as argument.
```
solegalli, line 34: Takes **a** dictionary..

feature_engine/variable_manipulation.py
```diff
@@ -2,33 +2,77 @@
...
     if not variables or isinstance(variables, list):
         variables = variables
     else:
         variables = [variables]
+
     return variables
 
 
-def _define_numerical_dict(the_dict):
-    # Check that the entered dictionary is indeed a dictionary of integers and floats
```
solegalli, line 17: Note that this function lives in a separate script as per code for the next release in the develop branch

feature_engine/variable_manipulation.py
```diff
@@ -2,33 +2,77 @@
...
+
     return the_dict
 
 
-def _find_numerical_variables(X, variables=None):
-    # Find numerical variables in a data set or check that
-    # the variables entered by the user are numerical.
+def _find_numerical_variables(X: pd.DataFrame, variables: List[str]=None) -> List[str]:
+    """
+    Takes Pandas DataFrame along with list of colum names and
```
solegalli, line 60: typo: colum**n**

feature_engine/variable_manipulation.py
```diff
@@ -2,33 +2,77 @@
...
+    checks if these columns are numerical. If not provided columns,
+    checks if all of them are numerical.
```
solegalli, line 62: This function checks that variables entered by user are of type numerical. If no variables were entered by the user, it will capture in a list all variables of type numerical present in the dataset.
Not sure this is clear from this docstring. Could we re-word a bit?

feature_engine/variable_manipulation.py
```diff
@@ -2,33 +2,77 @@
...
+
+    Args:
+        X: DataFrame to perform the check against
+        variables: List of variables. Defaults to None.
+
+    Raises:
+        ValueError: If all variables are non-numerical type or DataFrame is empty
+        TypeError: If some of the variables are non-numerical
```
solegalli, line 70: on the type error, maybe I would add that refers to "the variables entered by the user"

feature_engine/variable_manipulation.py
```diff
@@ -37,12 +81,27 @@ def _find_numerical_variables(X, variables=None):
         if len(X[variables].select_dtypes(exclude='number').columns) != 0:
             raise TypeError("Some of the variables are not numerical. Please cast them as numerical "
```
solegalli, line 82: wonder if we should replace this list comparison with a genexp? Did I not do this in the v1.0.0 branch? If not, maybe we should consider

feature_engine/variable_manipulation.py
```diff
@@ -52,10 +111,26 @@ def _find_categorical_variables(X, variables=None):
         if len(X[variables].select_dtypes(exclude='O').columns) != 0:
             raise TypeError("Some of the variables are not categorical. Please cast them as object "
```
solegalli, line 112: same here: should we replace this list comparison by a genexp?

feature_engine/variable_manipulation.py
```diff
@@ -52,10 +111,26 @@ def _find_categorical_variables(X, variables=None):
...
                             "before calling this transformer")
+
     return variables
 
 
-def _find_all_variables(X, variables=None):
+def _find_all_variables(X: pd.DataFrame, variables: List[str]=None) -> List[str]:
+    """
+    Takes Pandas DataFrame and extracts all variables if not provided corresponding list.
```
solegalli, line 120: I would re-phrase to "If None, captures all variables in the dataframe in a list. If user enters variable names list, it returns the list" or similar. The first sentence is not totally clear imho.

feature_engine/variable_manipulation.py
```diff
@@ -52,10 +111,26 @@ def _find_categorical_variables(X, variables=None):
...
+    If variables are provided, returns them back.
+
+    Args:
+        X:  DataFrame to perform the check against
+        variables: List of variables. Defaults to None.
+
+    Raises:
+        TypeError: If user provided list of variables include variable(s),
+                    which does not exist in DataFrame
```
solegalli, line 129: Could we re-word this sentence a bit? "If variable list provided by user contains variables not present in the dataframe" or something on those lines?

feature_engine/variable_manipulation.py
```diff
@@ -64,4 +139,5 @@ def _find_all_variables(X, variables=None):
         if len(set(variables).difference(X.columns)) != 0:
             raise TypeError("Some variables are not present in the dataset. Please check your variable"
```
solegalli, line 140: should we use a genexp?

feature_engine/variable_transformers.py
```diff
@@ -1,11 +1,14 @@
 # Authors: Soledad Galli <solegalli@protonmail.com>
 # License: BSD 3 clause
```
solegalli, line 3: this module was divided in submodules in the develop branch for the v1.0.0 release. So changes here need to be updated with current code architecture. But also, see my comments about type hint.

feature_engine/wrappers.py
```diff
@@ -1,10 +1,15 @@
+from typing import List, Optional
+
 import pandas as pd
```
solegalli, line 3: same here as with transformers module.

solegalli (changes requested):

solegalli:
Hi @Okroshiashvili 

Thank you so much for contributing all these changes and your engagement with Feature-engine.

A few things to note:

1) we are currently doing a major code refactor to get rid of some technical debt and improve readability. Latest changes are in the branch **develop**. Could you please make this PR to that branch?

2) type hint

Adding type hint is not in our priority list at the moment. You can find where we are heading in our issues list here in the repo and also propose changes there, perhaps before embarking on them, so we can align priorities and optimize the effort :). 

My thoughts on type hint are the following:

1) it is not mandatory as per [Pep 848](https://www.python.org/dev/peps/pep-0484/)
2) most libraries affiliated or sponsored by NumFocus (at least most of the ones I checked) have not adopted type hint (egs: Scikit-learn, yellobricks, MDAnalysis, dataretriever, dash, gensim, scikit-bio, statsmodels, PyCM3) and mlxtend (not sponsored by NumFocus but very relevant to our work). Bokeh on the other hand is using type hint
3) I think (and I might be wrong) that many Feature-engine users, and therefore potential contributors, are not familiar with type hint, so using type hint in the project may increase the contribution entry barrier.

I am inclined to postpone the adoption of type hint until we see where the community is going. Could I get your views on this @Okroshiashvili ?

Also @ChristopherGS can I get your thoughts on this?

Thank you!

feature_engine/variable_manipulation.py
ChristopherGS: *nit*: `typing.Optional` is more concise than `Union[None, ...]`
https://docs.python.org/3/library/typing.html#typing.Optional

feature_engine/variable_manipulation.py
ChristopherGS: In that case I think we want `Optional[dict]`

ChristopherGS:
I'm in favor of adding type hints, they help catch mistakes more often than you expect, and you get used to them very quickly. They are a pain at first though. I don't think the barrier is too high now. If you have concerns, I'd suggest a gradual rollout in the following ways:

- Add a new circleci job for the typehints, and make it optional
- Configure mypy to run against a subset of files at first (e.g. not the tests, [example from FastAPI](https://github.com/tiangolo/fastapi/blob/master/scripts/lint.sh#L6))
- Configure mypy to have some more relaxed fules in certain areas (example from [FastAPI](https://github.com/tiangolo/fastapi/blob/master/mypy.ini#L1))

Further down the line we might consider adding [recommit](https://pre-commit.com/) which blocks commits when style/type checks fail, but that also raises the barrier, so more long-term.

Okroshiashvili:
Sorry for being late, was a little bit busy. I will consider all your suggestions @solegalli and @ChristopherGS . I will undo type hints and postpone it for the future. Moreover, I will fix some bugs and improve docstrings. 

Should I move these changes to the develop branch or left them on the master branch? As I checked there are lots of changes in develop branch.

solegalli:
@Okroshiashvili 

Thanks for the message and picking this up. 

@ChristopherGS suggests to go forward with type hint, particularly as our code base is small, so let's keep it. And we expand it gradually as per his comment (he is the professional Python developer, I am just a data scientist :p)

Thanks for bringing type hint up, actually, it seems that it was a needed change that was out of my radar. 

Re: development branch is the current view of the latest code. So please PR there. We are working on multiple issues at a time, so it might change quite a bit on a daily basis. If I merge to master, I let you know.

Look forward to the changes!

Okroshiashvili:
@solegalli 

Let this PR be open until I will transfer all these changed to develop branch and then you can close it. I'm working on it and need some time. :smiley:

feature_engine/dataframe_checks.py
Okroshiashvili: @solegalli this function works when the input is only string due to Padas way to select columns. Is it necessary to restrict ```variables``` to be the only list?

Okroshiashvili: Tests don't fail and let the user be able to check only one variable against the null values. If after all the changes I have planned to do the test will fail, then I will notify you and propose the fix for this function.

feature_engine/variable_manipulation.py
Okroshiashvili: You re right @solegalli  and thanks @ChristopherGS for the suggestion.

feature_engine/variable_manipulation.py
Okroshiashvili: I fixed this bug and will send PR in develop branch.

feature_engine/variable_manipulation.py
Okroshiashvili: Marking as resolved as I fixed it.

feature_engine/variable_manipulation.py
Okroshiashvili: The default logic for ```variables``` is that the user should provide the list of strings. If the user provides only one variable as a string the function raises ```AttributeError``` due to a Pandas logic. Do we need to restrict the function to take the only a list of strings and not a string?
@solegalli

feature_engine/variable_manipulation.py
Okroshiashvili: The default logic for ```variables``` is that the user should provide the list of strings. If the user provides only one variable as a string the function raises ```AttributeError``` due to a Pandas logic. Do we need to restrict the function to take the only a list of strings and not a string?
@solegalli

Okroshiashvili: I already replaced list comprehension by genexp

feature_engine/variable_manipulation.py
Okroshiashvili: Changed list comprehension by genexp

feature_engine/variable_manipulation.py
Okroshiashvili: Using genexp, the test does not fail. So, no problem.

solegalli:
closed with #159

PR #163, opened 2020-10-22
feature_shuffling selector closes #136

### LINKED ISSUE #136: Feature_selection: feature shuffling
A transformer with an implementation of this:
https://github.com/solegalli/feature-selection-for-machine-learning/blob/master/11-Hybrid-methods/11.1-Feature-shuffling.ipynb

See what is available in scikit-learn first. I think they have something to derive importance by shuffling.

SunnyxBd:
added feature shuffling and skeleton code for unit tests


solegalli:
@SunnyxBd I add some suggestions on the issue. If they are not totally clear, don't worry, try to add as much as possible, make a new PR and we take it from there. 

The test_style is failing because we have tests on code style. We follow pep8. For them to pass, please use black to automatically reformat your code:
https://pypi.org/project/black/

It is super easy: you pip install black, if you are using pycharm, you can add it as external tool. Otherwise from console you do black name_of_python_script

We can leave this to the end as well.

Thank you!

feature_engine/selection/__init__.py
```diff
@@ -1,9 +1,10 @@
 """
-The module selection includes classes to select features or remove unwanted features
+The module selection includes classes to select features, remove unwanted features or shuffle feature values
 """
```
solegalli, line 3: I think the original sentence describes this module better:

"The module selection includes classes to select features or remove unwanted features"

Could we revert?

feature_engine/selection/shuffle_features.py
```diff
@@ -0,0 +1,213 @@
...
+from sklearn.ensemble import RandomForestClassifier, RandomForestRegressor
+from sklearn.metrics import roc_auc_score, mean_squared_error, r2_score
+
+from feature_engine.dataframe_checks import _is_dataframe, \
+    _check_input_matches_training_df
+
+from feature_engine.variable_manipulation import _define_variables, \
+    _find_all_variables
+
+
```
solegalli, line 14: we follow pep8 for code style.

According to pep8 imports should be first built in, second external libraries, and third local libraries, each block separated by an empty line, and within each block in alphabetical order.

So, sklearn.ensemble before sklearn.metrics, and then sklearn.models

Also, we do not use \, we make imports within parenthesis if more than 1 import from same module. This in particular, would be automatically corrected by black.

Thus:

```
from feature_engine.dataframe_checks import (
   import1,
   import 2,
)
```

feature_engine/selection/shuffle_features.py
```diff
@@ -0,0 +1,213 @@
...
+    Parameters
+    ----------
+
+    features_to_shuffle : str or list, default=None
+        The list of variable(s) to be shuffled from the dataframe.
+        If None, the transformer will shuffle all variables in the dataset.
+        
+    estimator: object, default = DecisionTreeClassifier or DecisionTreeRegressor
+        depending on 'regression' value
```
solegalli, line 37: default RandomForestClassifier()

feature_engine/selection/shuffle_features.py
```diff
@@ -0,0 +1,213 @@
...
+        Estimator can be a classifier or regressor
+        
+    scoring: str
+        metric used to evaluate to determine if feature needs to be kept or removed.
+
```
solegalli, line 42: can we copy from here:
https://github.com/solegalli/feature_engine/blob/e40457ce8b4baa1e146976bf1af1bbdf6eae1305/feature_engine/discretisation/decision_tree.py#L41

feature_engine/selection/shuffle_features.py
```diff
@@ -0,0 +1,213 @@
...
+    threshold: float
+        the value that defines if a feature will be kept or removed.
+
+    regression: boolean
+        Indicates whether the discretiser should train a regression or a classification
+        decision tree.
+
+    cv: boolean
+         Indicates whether cross-validation will be applied.
```
solegalli, line 51: can we copy from here:
https://github.com/solegalli/feature_engine/blob/e40457ce8b4baa1e146976bf1af1bbdf6eae1305/feature_engine/discretisation/decision_tree.py#L37

feature_engine/selection/shuffle_features.py
```diff
@@ -0,0 +1,213 @@
...
+    
+    """
+
+    def __init__(self,
+                 regression,
+                 cv = False,
+                 threshold = 0,
```
solegalli, line 58: threshold of 0 will not select any feature, I would use a default of 0.01

feature_engine/selection/shuffle_features.py
```diff
@@ -0,0 +1,213 @@
...
+        decision tree.
+
+    cv: boolean
+         Indicates whether cross-validation will be applied.
+    
+    """
+
+    def __init__(self,
+                 regression,
+                 cv = False,
```
solegalli, line 57: cv should be an int, and is the fold crossvalidation, so cv=3 means 3-fold crossvalidation

feature_engine/selection/shuffle_features.py
```diff
@@ -0,0 +1,213 @@
...
+                 threshold = 0,
+                 features_to_shuffle = None,
+                 estimator = None,
+                 scoring = None):
```
solegalli, line 61: Oh, I see that estimator is down here. I think estimator and scoring should be the first parameters followed by cv, then threshold, and then the rest.

feature_engine/selection/shuffle_features.py
```diff
@@ -0,0 +1,213 @@
...
+        
+        
+        if not isinstance(cv, bool):
+            raise ValueError("cv can only take True or False")
```
solegalli, line 65: cv should be int

feature_engine/selection/shuffle_features.py
```diff
@@ -0,0 +1,213 @@
...
+
+        if not isinstance(regression, bool):
+            raise ValueError("regression can only take True or False")
+ 
+        self.features_to_shuffle = \
```
solegalli, line 70: we don't use \

If it doesn't fit in a line, black will auto-correct it to a style that will pass the tests

feature_engine/selection/shuffle_features.py
```diff
@@ -0,0 +1,213 @@
...
+            _define_variables(features_to_shuffle)
+    
+        self.estimator = estimator
+        
+        self.scoring = scoring
+        
+        self.threshold = threshold
+        
+        self.regression = regression
+
```
solegalli, line 80: without blank line in between

feature_engine/selection/shuffle_features.py
```diff
@@ -0,0 +1,213 @@
...
+            model_performance = 1
+            
+        else:
+            
+            model = RandomForestClassifier().fit(X,
+                                                 y)
+            y_pred = model.predict_proba(X)[:, 1]
+            model_performance = roc_auc_score(y,
+                                              y_pred)
+
```
solegalli, line 136: I don't think we need this block, because either the default param or the user will define the estimator.

feature_engine/selection/shuffle_features.py
```diff
@@ -0,0 +1,213 @@
...
+
+        
+        # dict to collect features and their performance_drift 
+        
+        self.features_performance_drifts_ = {}
+
```
solegalli, line 142: I would call this attribute performance_drifts_ without features

feature_engine/selection/shuffle_features.py
```diff
@@ -0,0 +1,213 @@
...
+            model = RandomForestClassifier().fit(X,
+                                                 y)
+            y_pred = model.predict_proba(X)[:, 1]
+            model_performance = roc_auc_score(y,
+                                              y_pred)
+
+
+        
+        # dict to collect features and their performance_drift 
+
```
solegalli, line 140: To train the estimator, could we explore the use of cross_validate:
https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.cross_validate.html

I think if we set return_estimator=True, it returns the trained estimator (1 estimator per cv fold). Then, I understand we could use this to calculate the predictions of the shuffled features

Cross_validate will automatically check that the metric (scoring) and the estimator are valid sklearn functions, and it removes the problem of our hands. (you asked this in the issue).

csoss_validate, takes the scoring and cv as parameter and also the estimator, and does all relevant checks. So we shouldn't have to.

After we obtain the score, we should save it internally, so that we can determine the performance shifts

feature_engine/selection/shuffle_features.py
```diff
@@ -0,0 +1,213 @@
...
+                shuffled_y_pred = model.predict(X_shuffled)
+                shuffled_model_performance = 1
+                
+            else:
+                
+                shuffled_y_pred = model.predict_proba(X_shuffled)[:, 1]
+                shuffled_model_performance = roc_auc_score(y,
+                                                           shuffled_y_pred)
+
+
```
solegalli, line 167: because cross_val_score will return a score automatically, I don't think we need to break it down with the regression if here.

feature_engine/selection/shuffle_features.py
```diff
@@ -0,0 +1,213 @@
...
+             
+            # Calculate drift in model performance after the feature has
+            # been shuffled.
+            drift = model_performance - shuffled_model_performance
+
+            # Save feature and its performance drift in the 
+            # features_performance_drifts_ attribute.
+            self.features_performance_drifts_[feature] = drift
+
```
solegalli, line 175: we should also have another attribute returned by fit which is the selected features, that is all those features which performance drift is above the indicated threshold

feature_engine/selection/shuffle_features.py
```diff
@@ -0,0 +1,213 @@
...
+        X = _is_dataframe(X)
+
+        # Create a list of the features to be kept depending on their 
+        # performance drift and the threshold and the ML problem.
+        
+        if (self.regression):
+            columns_to_drop = [feature for (feature, drift) in self.features_performance_drifts_.items() if drift >= self.threshold ]
+        else:
+            columns_to_drop = [feature for (feature, drift) in self.features_performance_drifts_.items() if drift <= self.threshold ]
+
```
solegalli, line 212: I think we should capture the selected features in an attribute in the fit method, and here we just drop  them

feature_engine/selection/shuffle_features.py
```diff
@@ -0,0 +1,213 @@
...
+        Estimator can be a classifier or regressor
+        
+    scoring: str
+        metric used to evaluate to determine if feature needs to be kept or removed.
+
+    threshold: float
+        the value that defines if a feature will be kept or removed.
+
+    regression: boolean
+        Indicates whether the discretiser should train a regression or a classification
```
solegalli, line 47: I am not sure we need this parameter any more. We need to see.

feature_engine/selection/shuffle_features.py
```diff
@@ -0,0 +1,213 @@
...
+            X_shuffled[feature] = X_shuffled[feature].sample(frac=1).reset_index(drop=True)
+            
+            # Use fitted model to calculate new target prediction with
+            # the shuffled feature.
+            # Then, calculate this model performance.
+            if (self.regression):
+                
+                shuffled_y_pred = model.predict(X_shuffled)
+                shuffled_model_performance = 1
+
```
solegalli, line 160: here we would replace this bit by the fit estimators from cross_validate. Is this possible?

solegalli (commented):
Good progress thank you!

Would it be possible to check if we can incorporate the use of sklearn cross_validate to handle metrics, estimator and cv automatically?

thank you for your help!

feature_engine/selection/__init__.py
SunnyxBd: sure

feature_engine/selection/shuffle_features.py
SunnyxBd: Which version of black are you using? 
It didn't automatically add parenthesis after import when there are more than one module,  in my case.

feature_engine/selection/shuffle_features.py
SunnyxBd: I'm going to work on this next.

They way things are coded now is as follows: ( I'm going to push my code in a few)
fit an estimator for original dataset using cross validation
calculate model performance as the average of each folds' error.

for each feature:
fit the estimator again with the dataset having the feature  features
calculate new model performance as the average of each folds' error.
calculate model drift for each feature.

@solegalli

SunnyxBd:
@solegalli 

If we enter a dataframe with non reset indices, the pandas sampling methods returns nan and the transformer fails.

Here a reproduction of the problem:

`def load_diabetes_data():
    
    diabetes_X, diabetes_y = load_diabetes(return_X_y=True)
    data = pd.DataFrame(diabetes_X)
    target = pd.DataFrame(diabetes_y)
    
    return data, target

data, target = load_diabetes_data()

# Separate into train and test sets
X_train, X_test, y_train, y_test = train_test_split(
    data,
    target,
    test_size=0.2,
    random_state=0)

# initialize linear regresion estimator
linear_model = LinearRegression()
# initialize ShuffleFeaturesSelector 
transformer = ShuffleFeaturesSelector(
    estimator=model,
    scoring='r2',
    cv = 3)

# Fit the selector 
X = transformer.fit_transform(X_train, y_train)`

There are two ways of handling this problem:

1.  Whether we force resetting indices inside the method
2.  Whether we expect the user to enter a dataframe that has indices that are reset 
example:
`X = transformer.fit_transform(X_train.reset_index(), y_train.reset_index())`

What do you think @solegalli ?

solegalli:
@SunnyxBd Good catch! I think we should force reset inside the transformer. Could you add that? should we add this as a test as well?

SunnyxBd:
@solegalli 

Sure thing

solegalli:
@SunnyxBd not sure you saw my PR to your repo? I think I may have done some of the changes? have a look so we do not duplicate effort. Cheers

SunnyxBd:
@solegalli 

I just saw actually.

We both started on the rst document.
So, no surprise it had conflicts when I merged your PR. :)

I resolved the conflicts and kept both our work for now. 
Could you please check and see if we need to delete or add something ?

solegalli:
@SunnyxBd the PR in your repo is still open:
https://github.com/SunnyxBd/feature_engine/pull/2

solegalli:
@SunnyxBd thanks for merging the PR. The tests are failing now, and I think it might be because of the reset_index functionality that was just added. Could you please have a look? It was working before, in the PR that I made to your repo. 

Maybe it is a matter of adding proper column names to the diabetes dataset, but I am not sure.

SunnyxBd:
@solegalli 

I'm going to look into it.

SunnyxBd:
The issue was indeed caused by reset_index().

In fact, it kept a column with the index number in the dataframe.

this was solved by using the parameter drop=True in reset_index.

@solegalli

solegalli:
Amazing thank you @SunnyxBd !!! 

I'll merge this PR later today.

SunnyxBd:
Very happy to hear that!

PR #164, opened 2020-10-23
replace listcomp with genexp and formatted files with black

Okroshiashvili:
Replace all list comprehensions with genexp and format imports

### DIFF
--- a/tests/test_creation/test_mathematical_combination.py
+++ b/tests/test_creation/test_mathematical_combination.py
@@ -44,7 +44,7 @@ def test_default_parameters(df_vartypes):
         "min(Age-Marks)": "min",
     }
     # transform params
-    assert pd.testing.assert_frame_equal(X, ref) is None
+    pd.testing.assert_frame_equal(X, ref)
 
 
 def test_input_variables(df_vartypes):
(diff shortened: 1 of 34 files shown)


Okroshiashvili:
In commit 148bcfb I make assertion statements explicit. Pandas ```assert_frame_equal``` returns ```None``` when two DataFrames are equal. It's better for these assertions to be explicit, especially for newcomers.

solegalli:
Hey @Okroshiashvili I checked multiple examples of the use of pd.assert_frame_equal, and I haven't seen anywhere that they use it explicitly. So I am disinclined to change it.

Most relevant repo is the one of category_encoders, same concept as feature-engine but only for categorical variables:
https://github.com/scikit-learn-contrib/category_encoders/search?q=assert_frame_equal

pandas documentations uses none explicit:
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.testing.assert_frame_equal.html

stackoverflow:
https://stackoverflow.com/questions/38839402/how-to-use-assert-frame-equal-in-unittest
https://stackoverflow.com/questions/27950891/how-to-use-a-pandas-data-frame-in-a-unit-test

and others:
https://github.com/scikit-hep/root_pandas/blob/master/tests/test.py
https://programtalk.com/python-examples/pandas.util.testing.assert_frame_equal/
https://intothedepthsofdataengineering.wordpress.com/category/debugging/

solegalli:
Also, as per PEP8, sklearn imports should go before feature engine imports, as sklearn is a third party library. Is there a reason you move them down in all the tests?

Okroshiashvili:
@solegalli changing import order is due to my VSCode setting and I will make corrections. I had no particular intention to make them that way. My fault.

Regarding explicitly stating assertions, my main reason was to make them explicit for newcomers or new contributors. However, I can undo those changes and restore the previous state of the code.

Issue #152, opened 2020-10-07
roll out type hint


solegalli:
it might be beneficial to take advantage of Feature-Engine's relatively small code base to do it whilst it's still not so much work.

I'd suggest a gradual rollout in the following ways:

- Add a new circleci job for the typehints, and make it optional for merging prs

- Configure mypy to run against a subset of files at first (e.g. not the tests, example from [FastAPI](https://github.com/tiangolo/fastapi/blob/master/scripts/lint.sh#L6))

- Configure mypy to have some more relaxed rules in certain areas (example from [FastAPI](https://github.com/tiangolo/fastapi/blob/master/mypy.ini#L1))

Further down the line we might consider adding [precommit](https://pre-commit.com/) which blocks commits when style/type checks fail, but that also raises the barrier, so more long-term.

solegalli:
@Okroshiashvili I know this is a big ask, but I'll ask anyway :)

Would you be able to help extend type_hint to the rest of the scripts? 

Should we do each of the folders discretisation, encoding, imputation, outliers and selection in a different PR? Or should we do them altogether in one?

I don't think that these folders will change much in the next months so we could do in 1 PR as well. Maybe that's best.

Thank you!

solegalli:
Great! that's fantastic news. Thank you so much. 

Any time you can volunteer is extremely welcome. Thanks a lot.

Develop branch was merged to master, so next PR should be to master now :)

Okroshiashvili:
@solegalli next PRs should be on the master branch?

solegalli:
@Okroshiashvili yes please :)

Okroshiashvili:
@solegalli I'm on my way. There will be some changes tomorrow. 

Btw, I had lots of conflicts while trying to make my fork even with the current master. I decided to remove my fork of the repo and reforked it again. I do hope my past work is not lost. :)

solegalli:
you passed work is merged in master. So you should be able to recover it ;). I can also see you there as one of the contributors (and I already added your name in the changelog.rst in docs, for the next release), so no stresses :)

Sorry you had problems with the fork. Look forward to the changes. Have a good day!

solegalli:
@Okroshiashvili 1 suggestion, if you don't mind, please leave reformatting the folder "selection" to the end, cause we are adding transformers there. In case it generates conflicts with other contributors. Cheers

Okroshiashvili:
[...]

PR #166, opened 2020-10-27
add type annotations and docstrings

Okroshiashvili:
Add new type annotations and docstrings, as well as fix some old docstrings to be consistent.

### DIFF
--- a/feature_engine/base_transformers.py
+++ b/feature_engine/base_transformers.py
@@ -16,8 +16,15 @@ from feature_engine.variable_manipulation import _find_numerical_variables
 
 
 class BaseNumericalTransformer(BaseEstimator, TransformerMixin):
-    # shared set-up procedures across numerical transformers, i.e.,
-    # variable transformers, discretisers, math combination
+    """
+    Shared set-up procedures across numerical transformers, i.e.,
+    variable transformers, discretisers, math combination.
+
+    Methods:
+        fit(): Fits the transformation to the data
+        transform(): Apply the transformation to the data
+
+    """
 
     def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None) -> pd.DataFrame:
         """
@@ -55,6 +62,7 @@ class BaseNumericalTransformer(BaseEstimator, TransformerMixin):
 
         Returns:
             Transformed DataFrame
+
         """
 
         # Check method fit has been called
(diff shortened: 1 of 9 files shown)


solegalli:
Hi @Okroshiashvili 

before making more changes, it would be good to discuss the format of the docstrings. I like how it is done in the package imbalanced learn:
https://imbalanced-learn.readthedocs.io/en/stable/generated/imblearn.under_sampling.CondensedNearestNeighbour.html#

and here the source code:
https://github.com/scikit-learn-contrib/imbalanced-learn/blob/12b2e0d/imblearn/under_sampling/_prototype_selection/_condensed_nearest_neighbour.py#L27

But for some reason, if in feature engine we pass the attributes to the init docstring, and the methods as well. I did not manage to make them appear under the right headings (methods, attributes).

Does this work for you?

I see that here:
https://feature-engine.readthedocs.io/en/latest/vartransformers/BoxCoxTransformer.html

lambda_dict does not appear under attributes, not sure if the source code says it should or not.

Could you let me know ?

cheers
Sole

solegalli:
see #167

Okroshiashvili:
@solegalli  I've checked the links you provide and they seem to use Numpy style docstrings. I'm using Google style docstrings. To build docs properly, we need to enable Sphinx extension Napoleon and build docs with it. After that, I hope everything will be as expected. For more info check the following: https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html

We can discuss and agree with unified style for docstrings in the code base and then I will rewrite all docs in that style.

solegalli:
@Okroshiashvili thanks for the link. That was new to me :)

Could we stick to Numpy style then? most of the docs are already with the underline and without indentation.

At the top of the dosctrings, could we include **explanation**, **params**, **attributes** and **methods**, plus **examples** if any? could we add the >>> to the code examples if they are not there already?

and then the docstrings for the methods themselves, with sections **args**, **raises**,  and **returns**?

Also, could we have the type_hint changes in one PR and the changes in the docstrings in a different PR? it is easier to review. 
And also, if we need changes, I think it makes the back and forth easier as well.

Thanks a lot!

Okroshiashvili:
@solegalli Sure we can stick to the Numpy style docstrings. Even, we can use ```imbalanced-learn``` as a template to build better Docs for ```feature-engine```.

I will reference this thread in #167 and in #152 

I will start working on writing type hints and send one PR for that and one separate PR for docstrings. I will do my best to include everything you suggested.

Let this PR be open until I will do all the necessary preparation work and then I will close this PR.

It's better first to complete type annotations and then go for better Docs, as the latter requires more work. Hence, I will send new PR which will only include type annotations. After we finish writing type annotations then I will work on rewriting Docs.

If this sounds good, please confirm. :)

solegalli:
@Okroshiashvili sounds great! thank you!

Okroshiashvili:
closing this PR in favor of separate PR for type annotations

PR #170, opened 2020-10-29
Target mean encoding

ttungl:
I would like Soledad to review my transformer code. Thanks.

### DIFF
--- a/feature_engine/selection/target_mean_encoding.py
+++ b/feature_engine/selection/target_mean_encoding.py
@@ -270,5 +270,5 @@ class TargetMeanEncoderFeatureSelector(BaseEstimator, TransformerMixin):
             None
 
         """
-        X_test_enc, y_test = self.fit(X, y, quantiles=quantiles, test_size=test_size, random_state=random_state)
-        return self.transform(X_test_enc, y_test)
+        self.fit(X, y, quantiles=quantiles, test_size=test_size, random_state=random_state)
+        return self.transform(self.X_test_enc, self.y_test)
--- a/feature_engine/selection/target_mean_encoding.py
+++ b/feature_engine/selection/target_mean_encoding.py
@@ -83,8 +83,7 @@ class TargetMeanEncoderFeatureSelector(BaseEstimator, TransformerMixin):
         
         Returns
         -------
-            X_test_enc
-            y_test
+            None
 
         Parameters
         ----------
--- a/feature_engine/selection/target_mean_encoding.py
+++ b/feature_engine/selection/target_mean_encoding.py
@@ -37,9 +37,9 @@ class TargetMeanEncoderFeatureSelector(BaseEstimator, TransformerMixin):
 
             Public methods
             --------------
-                `fit(self, X, y=None)`
+                `fit(self, X, y)`
                 `transform(self)`
-                `fit_transform(self, X, y=None)`
+                `fit_transform(self, X, y)`
         
     Parameters
     ----------
@@ -200,7 +200,7 @@ class TargetMeanEncoderFeatureSelector(BaseEstimator, TransformerMixin):
 
         return self
 
-    def transform(self, X=None, y=None):
+    def transform(self):
         """
         Performs
         --------
--- a/feature_engine/selection/target_mean_encoding.py
+++ b/feature_engine/selection/target_mean_encoding.py
@@ -48,16 +48,6 @@ class TargetMeanEncoderFeatureSelector(BaseEstimator, TransformerMixin):
         The list of variables to evaluate. If None, the transformer will evaluate all 
         variables in the dataset associated with the variables_type.
 
-    target: string, default=None
-        The target variable to evaluate. If None, the transformer will return the 
-        relevant message for the missing value.
-
-    variables_type: boolean, default=True
-        If variables_type = True: the transformer will evaluate the variables as the 
-            categorical variables. 
-        If variables_type = False: the transformer will evaluate the variables as the
-            numerical variables.
-
     scoring: string, default='roc_auc_score'
         This indicates the metrics score to perform the feature selection.
         The current support includes 'roc_auc_score' and 'r2_score'.
--- a/feature_engine/selection/target_mean_encoding.py
+++ b/feature_engine/selection/target_mean_encoding.py
@@ -1,17 +1,24 @@
+import pandas as pd 
+
 from sklearn.base import TransformerMixin, BaseEstimator
+from sklearn.pipeline import Pipeline
 from sklearn.utils.validation import check_is_fitted
 from sklearn.model_selection import train_test_split
+
 from sklearn.metrics import roc_auc_score
 from sklearn.metrics import r2_score
-from feature_engine.variable_manipulation import _find_all_variables
-from feature_engine.variable_manipulation import _find_categorical_variables
-from feature_engine.variable_manipulation import _find_numerical_variables
-from feature_engine.variable_manipulation import _define_variables
-from feature_engine.dataframe_checks import _is_dataframe
-from feature_engine import categorical_encoders as ce
-from feature_engine import discretisers as dsc
 
-import pandas as pd 
+from feature_engine.variable_manipulation import (
+        _find_categorical_variables,
+        _find_numerical_variables,
+        _define_variables
+)
+
+from feature_engine.dataframe_checks import _is_dataframe
+from feature_engine.categorical_encoders import MeanCategoricalEncoder
+from feature_engine.discretisers import EqualFrequencyDiscretiser 
+
+
 
 class TargetMeanEncoderFeatureSelector(BaseEstimator, TransformerMixin):
     """
@@ -55,13 +62,16 @@ class TargetMeanEncoderFeatureSelector(BaseEstimator, TransformerMixin):
 
     """
 
-    def __init__(self, variables=None, scoring='roc_auc_score'):
+    def __init__(self, variables=None, scoring='roc_auc_score', quantiles=5, test_size=0.3, random_state=0):
         self.variables = _define_variables(variables)
         self.scoring = scoring
+        self.quantiles = quantiles
+        self.test_size = test_size
+        self.random_state = random_state
         self.X_test_enc = None
         self.y_test = None
 
-    def fit(self, X, y, quantiles=5, test_size=0.3, random_state=0):
+    def fit(self, X, y):
         """
         Performs
         --------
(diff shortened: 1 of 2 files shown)


feature_engine/selection/target_mean_encoding.py
```diff
@@ -0,0 +1,400 @@
+from sklearn.base import TransformerMixin, BaseEstimator
+from sklearn.utils.validation import check_is_fitted
+from sklearn.model_selection import train_test_split
+from sklearn.metrics import roc_auc_score, r2_score
+
```
solegalli, line 5: we follow pep8 for imports:
https://pep8.org/#imports

would you mind having a look and reordering?

feature_engine/selection/target_mean_encoding.py
```diff
@@ -0,0 +1,400 @@
...
+from feature_engine.variable_manipulation import _find_all_variables, _define_variables
+from feature_engine.dataframe_checks import _is_dataframe
+
+import pandas as pd 
+
+class TargetMeanEncoding(BaseEstimator, TransformerMixin):
+    """
```
solegalli, line 12: we need to think of a different name for this transformer. This transformer aims to select features, so it need to me more or less clear from the name, what the functionallity does.

Something like TargetMeanEncoderFeatureSelector(), or maybe you are more creative?

feature_engine/selection/target_mean_encoding.py
```diff
@@ -0,0 +1,400 @@
...
+        
+    Parameters
+    ----------
+
+    variables: list, default=None
+        The list of variables to evaluate. If None, the transformer will evaluate all 
+        variables in the dataset associated with the variables_type.
+
+    target: string, default=None
+        The target variable to evaluate. If None, the transformer will return the
```
solegalli, line 52: we do not need to specify target, because this is not a parameter from the init method. Please remove

feature_engine/selection/target_mean_encoding.py
```diff
@@ -0,0 +1,400 @@
...
+    variables: list, default=None
+        The list of variables to evaluate. If None, the transformer will evaluate all 
+        variables in the dataset associated with the variables_type.
+
+    target: string, default=None
+        The target variable to evaluate. If None, the transformer will return the 
+        relevant message for the missing value.
+
+    variables_type: boolean, default=True
+        If variables_type = True: the transformer will evaluate the variables as the
```
solegalli, line 56: I don't think we need this parameter.  _define_all_variables in fit will select all variables. If the user wants a specific set, they should input it manually. This way this transformer follows the same logic and use of all the remaning transformers from sklearn. Could you please remove?

feature_engine/selection/target_mean_encoding.py
```diff
@@ -0,0 +1,400 @@
...
+        relevant message for the missing value.
+
+    variables_type: boolean, default=True
+        If variables_type = True: the transformer will evaluate the variables as the 
+            categorical variables. 
+        If variables_type = False: the transformer will evaluate the variables as the
+            numerical variables.
+
+    metrics_score: string, default='roc_auc_score'
+        This indicates the metrics score to perform the feature selection.
```
solegalli, line 62: this parameter should be called "scoring" so that we use the same name in all feature_engine transformers and it is also the same name given in sklearn transformers. It will also take the values permitted in sklearn metrics:

https://scikit-learn.org/stable/modules/model_evaluation.html
https://scikit-learn.org/stable/modules/model_evaluation.html
(https://scikit-learn.org/stable/modules/model_evaluation.html#the-scoring-parameter-defining-model-evaluation-rules)

feature_engine/selection/target_mean_encoding.py
```diff
@@ -0,0 +1,400 @@
...
+        self.variables = _define_variables(variables)
+        self.vars_type = variables_type
+        self.metrics_score = metrics_score
+        self.target = target
+        self.test_size = test_size
+        self.random_state = random_state
+        self.quantiles = quantiles
+
+        # transformed variables
+        self.target_enc_dict = {}
```
solegalli, line 89: these attributes should be initialised in the fit method, as they will be populated there and be followed by a trailing underscore, so that sklearn can recognise that the trasnformer has been fit.

Sklearn looks for attributes with trailing underscores to check that transformers were fit.

Also, I don't think we need to store the transformer columns. And I also think we should capitalise on feature-engine discretisers to perform the discretisation, instead of adding code manually. We have that functionality already.

feature_engine/selection/target_mean_encoding.py
```diff
@@ -0,0 +1,400 @@
...
+        self.binned_vars = []
+        
+        # dataframes
+        self.X_train, self.X_test = None, None
+        self.y_train, self.y_test = None, None
+        self.X_train_enc, self.X_test_enc = None, None
+        
+        # output
+        self.output = None
+
```
solegalli, line 100: we don't need lines 94 to 99

feature_engine/selection/target_mean_encoding.py
```diff
@@ -0,0 +1,400 @@
...
+
+    def __target_mean_encoding(self, df_train_, df_test_, var_cols=None):
+        """
+        Performs
+        --------
+            For each variable:
+                + Performs the target mean encoding on the train set.
```
solegalli, line 107: to determine target mean encoding, we can use the MeanEncoder from feature-engine. Could you experiment with that? I don't think we need to write this functionality from scratch.

If I remember correclty, the MeanEncoder will encode categories with the target Mean, and for numerical variables, they need to be discretised first, making sure we set the return_object=True and then we can run the MeanEncoder at the back

feature_engine/selection/target_mean_encoding.py
```diff
@@ -0,0 +1,400 @@
...
+            self.target_enc_dict[col] = __target_mean_enc
+
+        # drop target from the updated datasets
+        df_train.drop([self.target], axis=1, inplace=True)
+        df_test.drop([self.target], axis=1, inplace=True)
+
+        return df_train, df_test 
+
+    def __validate_variables(self, X):
+        """
```
solegalli, line 159: we have a lot of code to validate variables in the variable_manipulation.py. Can we use those functions instead?

solegalli (changes requested):
Hi @ttungl thank you very much for such an amount of work. This is very impressive.

I think we could simplify this transformer quite a bit leveraging code and transformers that already exist in Feature-engine.

For example, to perform discretisation we have the EqualWidhtDiscretiser and EqualFrequencyDiscretiser, I think we should use those, and give the user the possibility to choose which one they want to use for numerical variables, instead of adding discretisation code in this transformer.

Regarding target mean encoding, in the encoders folder we have the MeanEncoder, which does exactly that, replaces categories by the mean of the target. The MeanEncoder can be used straighaway with categorical variables, and can also be used with discretised numerical variables, if we set in the discretisaation transformers the parameter return_object=True.

So in the fit method, I think we should build a small pipeline with discretisation +meanEncoder for numerical variables and use the mean encoder for categorical variables.

See for example here for a feature-engine transformer that uses pipelines:
https://github.com/solegalli/feature_engine/blob/master/feature_engine/encoding/decision_tree.py#L133

It would be great if you could simplify the transformer in the main script by leveraging existing transformers (or tell my why it would not work, or what other problems arise). And then we take it from there?

I am also thinking if we can use scikit-learn cross_validate instead of splitting in train and test, but maybe we leave this for a second step.

Thanks a lot and let me know what you think

Cheers

feature_engine/selection/target_mean_encoding.py
ttungl: Thanks for your reference. I fixed it.

feature_engine/selection/target_mean_encoding.py
ttungl: Your suggested name sounds good to me. Thanks.

feature_engine/selection/target_mean_encoding.py
ttungl: I still not understand your point. Would you mind to share if we don't specify the target in the init method, where should it be initialized in order to pass it to the __target_mean_encoding method?

feature_engine/selection/target_mean_encoding.py
ttungl: Thanks, Sole, that got fixed.

feature_engine/selection/target_mean_encoding.py
ttungl: Those variables have been removed.

feature_engine/selection/target_mean_encoding.py
ttungl: Hi Sole,
I still not figure out what it would be if we don't have this parameter. What scenarios should be covered here? I could have several cases as follows.
1. No variables at input: the output will be the categorical variables for feature selection.
2. With variables: 
a. check if the inputs are numerical variables then calculate using the numerical process, otherwise, calculate them using the categorical process. 
b. check if the inputs are mixed, then calculate the default mode using the categorical process.
Is it okay to do this approach? Please let me know your thought. Thanks Sole!

feature_engine/selection/target_mean_encoding.py
ttungl: Can you please specify that should we use the single trailing underscore (a_) or double trailing underscore (__a__)? Sorry for this confusion. Thanks Sole!

feature_engine/selection/target_mean_encoding.py
ttungl: Yes, I will use the MeanEncoder instead of the __target_mean_encoding method. Thanks Sole!

feature_engine/selection/target_mean_encoding.py
ttungl: Yes, I will use those functions for validation. Thanks Sole!

feature_engine/selection/target_mean_encoding.py
solegalli: pandas can detect the type of the variables. So we should have this functionality as part of the transformer and not ask the user to enter it. Thus, we do not need this parameter. We will do the variable type identification as part of the process.

In fact, if we use a combination of the transformers I suggested, they should automatically find numerical and categorical variables. So I am thinking that we do not need to add this functionality at all for this transformer, as it is already part of the other 2

feature_engine/selection/target_mean_encoding.py
solegalli: single trailing underscore `a_`

solegalli (commented):
@ttungl 

Hey, thanks for the work. I think I've answered all your questions. Let me know when you push the new code and would like me to review it.

Cheers

ttungl:
> @ttungl
> 
> Hey, thanks for the work. I think I've answered all your questions. Let me know when you push the new code and would like me to review it.
> 
> Cheers

I have pushed my new update, please review it. Thanks Sole!

feature_engine/selection/target_mean_encoding.py
```diff
@@ -0,0 +1,264 @@
...
+from sklearn.model_selection import train_test_split
+from sklearn.metrics import roc_auc_score
+from sklearn.metrics import r2_score
+from feature_engine.variable_manipulation import _find_all_variables
+from feature_engine.variable_manipulation import _find_categorical_variables
+from feature_engine.variable_manipulation import _find_numerical_variables
+from feature_engine.variable_manipulation import _define_variables
+from feature_engine.dataframe_checks import _is_dataframe
+from feature_engine import categorical_encoders as ce
+from feature_engine import discretisers as dsc
```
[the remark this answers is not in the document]
solegalli: we follow pep 8 to order the imports at the top, please have a look here:
https://pep8.org/#imports

feature_engine/selection/target_mean_encoding.py
```diff
@@ -0,0 +1,264 @@
...
+
+    """
+
+    def __init__(self, variables=None, scoring='roc_auc_score'):
+        self.variables = _define_variables(variables)
+        self.scoring = scoring
+        self.X_test_enc = None
+        self.y_test = None
+
+    def fit(self, X, y, quantiles=5, test_size=0.3, random_state=0):
```
solegalli, line 64: the fit method must only take the parameters X and y, so it remains compatible with the Scikit-learn pipeline. We need to send all the other params to init.

feature_engine/selection/target_mean_encoding.py
```diff
@@ -0,0 +1,264 @@
...
+
+        # check the target
+        if y is None:
+            raise ValueError('Please provide a target y for this encoding method.')
+
+        # check input dataframe
+        X = _is_dataframe(X)
+
+        # find all variables or check those entered are in the dataframe
+        self.variables = _find_all_variables(X, self.variables)
```
solegalli, line 114: I don't think we need this line, the functions below will automatically find categorical and numerical variables

feature_engine/selection/target_mean_encoding.py
```diff
@@ -0,0 +1,264 @@
...
+        
+        # find categorical variables or check that those vars entered by the user
+        # are of type object
+        cat_variables = _find_categorical_variables(X[self.variables] if self.variables else X)
+        num_variables = _find_numerical_variables(X[self.variables] if self.variables else X)
```
solegalli, line 119: I think it might be better to rename these as:

```
variables_categorical
variables_numerical
```

and we need to make them part of self.

so:

```
self.variables_categorical
self.variables_numerical
```

feature_engine/selection/target_mean_encoding.py
```diff
@@ -0,0 +1,264 @@
...
+                                                random_state = random_state)
+
+        ## filling missing values.
+        for col in cat_variables:
+            X_train[col].fillna('M', inplace=True)
+            X_test[col].fillna('M', inplace=True)
+
+        for col in num_variables:
+            X_train[col].fillna(0, inplace=True)
+            X_test[col].fillna(0, inplace=True)
```
solegalli, line 138: it is the user's responsibility to make sure their variables do not contain missing values. Dataframes used with this transformer should be complete (that is, no missing values).

The categorical encoder and the discretiser would check that the data contains no na, and if it does, it would raise an error. So we are good to delete lines 131 to 138.

feature_engine/selection/target_mean_encoding.py
```diff
@@ -0,0 +1,264 @@
...
+
+        # categorical variables
+        if cat_variables:
```
solegalli, line 141: the MeanCategoricalEncoder can handle all variables at the same time, so we don't need a loop.

Simply `MeanCategoricalEncoder(variables=self.variables_categorical)`

feature_engine/selection/target_mean_encoding.py
```diff
@@ -0,0 +1,264 @@
...
+            # transform
+            X_train_cat_enc, X_test_cat_enc = encoder.transform(X_train[cat_variables]), encoder.transform(X_test[cat_variables])
+
+            # filling missing values on the results
+            X_train_cat_enc, X_test_cat_enc = X_train_cat_enc.fillna(0), X_test_cat_enc.fillna(0)
+
+        # numerical variables
+        self.numerical_binned_ = dict()
+        if num_variables: 
+            # Equal-frequency discretizer
```
solegalli, line 155: same here, the discretiser can handle all variables at the same time, so no need to make a loop

feature_engine/selection/target_mean_encoding.py
```diff
@@ -0,0 +1,264 @@
...
+            # mean encoding
+            encoder.fit(X_train_t[num_variables], y_train)
+            X_train_num_enc, X_test_num_enc = encoder.transform(X_train_t[num_variables]), encoder.transform(X_test_t[num_variables])
+            X_train_num_enc.columns = [i+'_binned' for i in X_train_num_enc.columns]
+            X_test_num_enc.columns = [i+'_binned' for i in X_test_num_enc.columns]
+
+            self.numerical_binned_ = disc.binner_dict_
+            # filling missing values on the results
+            X_train_num_enc, X_test_num_enc = X_train_num_enc.fillna(0), X_test_num_enc.fillna(0)
+
```
solegalli, line 174: no need to remove na, data should be na free.

feature_engine/selection/target_mean_encoding.py
```diff
@@ -0,0 +1,264 @@
...
+
```
solegalli, line 175: For numerical variables, we should include both transformers into a pipeline, similar to what is done here:
https://github.com/solegalli/feature_engine/blob/master/feature_engine/encoding/decision_tree.py#L133

solegalli (changes requested):
Hi @ttungl thank you so much for refactoring the transformer.

It still needs a bit of work. I made some suggestions, see if you could factor those in, and we take it from there?

Don't worry too much about the docstrings at the moment I would say. We do that last.

Thank you and apologies for the delay in the response.
Had a few deadlines at work.

Cheers

Issue #137, opened 2020-10-02
feature_selection: select by individual feature performance with a ML algo


solegalli:
A transformer that implements this:

https://github.com/solegalli/feature-selection-for-machine-learning/blob/master/06-Filter-other-metrics/06.1-Univariate-roc-auc.ipynb

Should be able to take pretty much any ML algorithm

rupeshpalwadi:
can you assign this to me?

solegalli:
With pleasure :)

Please checkout the **develop branch** (not master) and make the PR to that branch, as I am currently refactoring code, and the modules have changed their names.

This transformer should allow the user to decide which ML model they want to use, and which metric (if possible). So we need to think a bit how to play with this, because some metrics are better when they are smaller, and some when they are bigger.

Look forward to the PR :)

solegalli:
@rupeshpalwadi any news on the development of this feature?

solegalli:
@NicoGalli 

this one is more complicated than correlated, but there is a template that we could use to create the transformer. See for example this:
https://github.com/solegalli/feature_engine/blob/master/feature_engine/selection/shuffle_features.py

The idea is:

for feature in X.variables:
 (so feature per feature)

train a machine learning model using the cross_validate function from sklearn:
https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.cross_validate.html

cross_validate takes any machine learning model available in sklearn as input, any metric available in sklearn as input, a cross validation fold (please do a quick search to understand what this is) and it returns trained machine lerning model if return_estimator=True. and the model performance.

https://github.com/solegalli/feature_engine/blob/master/feature_engine/selection/shuffle_features.py
you can see how we call all of these

So, as I was saying, feature per feature, using cross_validate, train a machine learning model (user enters this) with the single feature, against the target). Determine the performance and store it

the fit attributes should be a dictionary of feature:performance pairs and the selected features. Selected features are those which performance is above a threshold, the threshold is defined by the user.

The code should be a simpler version of the shuffle_features.

I would say, make an initial PR with whatever you understand from all of what I am saying and we take it from there :)

NicoGalli:
OK @solegalli I will give it a try. Would you please assign the issue to me?
Thanks,
N.-

PR #173, opened 2020-11-02
feature selection initial commit #137

NicoGalli:
feature_selection.py Class with some initial code


feature_engine/selection/feature_selection.py
```diff
@@ -0,0 +1,176 @@
...
+    _check_input_matches_training_df,
+)
+from feature_engine.variable_manipulation import (
+    _define_variables,
+    _find_numerical_variables,
+)
+
+
+class FeatureSelection(BaseEstimator, TransformerMixin):
+    """
```
solegalli, line 19: could we please rename the class to UnivariatePerformanceSelection() or SignleFeaturePerformanceSelection() not sure which one is best. Maybe the second?

feature_engine/selection/feature_selection.py
```diff
@@ -0,0 +1,176 @@
...
+
+        self
+        """
+
+        # check input dataframe
+        X = _is_dataframe(X)
+
+        # reset the index
+        X = X.reset_index(drop=True)
+        y = y.reset_index(drop=True)
```
solegalli, line 118: we don't need to reset indeces. Can we remove?

feature_engine/selection/feature_selection.py
```diff
@@ -0,0 +1,176 @@
...
+        self.selected_features_ = []
+
+        # train model with all features and cross-validation
+        for feature in self.variables:
+            model = cross_validate(
+                self.estimator,
+                X[feature].to_frame(),
+                y,
+                cv=self.cv,
+                return_estimator=True,
```
solegalli, line 133: we don't need to recover the trained machine learning model in this selector. We only need the performance. So we can set this parameter to False.

feature_engine/selection/feature_selection.py
```diff
@@ -0,0 +1,176 @@
...
+                scoring=self.scoring,
+            )
+
+            if model["test_score"].mean() > self.threshold:
+                self.selected_features_.append(feature)
+
```
solegalli, line 139: this is perfect. The only thing I would like to add is an attribute called feature_importance_ to the fit method with a dictionary of the feature:model["test_score"].mean() 

Would you be able to add that please?

feature_engine/selection/feature_selection.py
```diff
@@ -0,0 +1,176 @@
+import numpy as np
+from sklearn.base import BaseEstimator, TransformerMixin
+from sklearn.ensemble import RandomForestClassifier
+from sklearn.metrics import get_scorer
```
solegalli, line 4: we are not using get_scorer, no? can we remove? also, are we using np? otherwise please also remove from theimports

feature_engine/selection/feature_selection.py
```diff
@@ -0,0 +1,176 @@
...
+        """
+
+        # check if fit is performed prior to transform
+        check_is_fitted(self)
+
+        # check if input is a dataframe
+        X = _is_dataframe(X)
+
+        # reset the index
+        X = X.reset_index(drop=True)
```
solegalli, line 171: we don't need to reset the index, please remove

solegalli (commented):
@NicoGalli really, really well done. That captures everything.

If you could add some tests, that would be great, we would need tests with different machine learning models, for regression and classification. Maybe we can try, RandomForests(), LinearRegression(), and a nearestneighbour for classification or regression? we will need different dataframes for the tests, which you can get from sklearn.

feature_engine/selection/single_feature_performance_selection.py
```diff
@@ -0,0 +1,171 @@
...
+                X[feature].to_frame(),
+                y,
+                cv=self.cv,
+                return_estimator=False,
+                scoring=self.scoring,
+            )
+
+            if model["test_score"].mean() > self.threshold:
+                self.selected_features_.append(feature)
+                self.feature_importance_[feature] = model["test_score"].mean()
```
solegalli, line 136: this line should be outside the if, because we want to store the importance for every feature, whether they are selected or not. So that the user can have a look at why a feature may have been selected.

tests/test_selection/test_single_feature_performance_selection.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+@pytest.fixture(scope="module")
+def load_diabetes_dataset():
+    # Load the diabetes dataset from sklearn
+    diabetes_X, diabetes_y = load_diabetes(return_X_y=True)
+    X = pd.DataFrame(diabetes_X)
+    y = pd.DataFrame(diabetes_y)
+    return X, y
+
```
solegalli, line 39: Another contributor, is going to move these functions to a config inside the test_selection folder, we need to coordinate and rebase master when this is done.

tests/test_selection/test_single_feature_performance_selection.py
```diff
@@ -0,0 +1,136 @@
...
+
+def test_default_parameters(df_test):
+    X, y = df_test
+    sel = SignleFeaturePerformanceSelection(RandomForestClassifier(random_state=1))
+    sel.fit(X, y)
+
+    # expected result
+    Xtransformed = pd.DataFrame(X["var_7"].copy())
```
solegalli, line 47: did you check that the only selected variable is 7?

tests/test_selection/test_single_feature_performance_selection.py
```diff
@@ -0,0 +1,136 @@
...
+        "var_9",
+        "var_10",
+        "var_11",
+    ]
+    assert sel.threshold == 0.5
+    assert sel.cv == 3
+    assert sel.scoring == "roc_auc"
+    # test fit attrs
+    assert sel.selected_features_ == ["var_7"]
+    # test transform output
```
solegalli, line 69: we should also test the dictionary with the feature:importances pairs.

tests/test_selection/test_single_feature_performance_selection.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+def test_regression_cv_3_and_r2(load_diabetes_dataset):
+    #  test for regression using cv=3, and the r2 as metric.
+    X, y = load_diabetes_dataset
+    sel = SignleFeaturePerformanceSelection(estimator=LinearRegression(), scoring="r2", cv=3)
+    sel.fit(X, y)
+
+    # expected output
+    Xtransformed = pd.DataFrame(X[[1, 2, 3, 4, 5, 8]].copy())
```
solegalli, line 80: same here, did you test that these are the selected variables?

solegalli (changes requested):
@NicoGalli this is in very good shape thank you!

I think the tests are failing because we need to have a look at how the transformer assigns the importances to the variables and which ones are selected, so that we can incorporate the right values to the test.

I can pick up this, later in the week. Not sure I have time in the weekend. Do you want me to pick this up? or would you like to try to get the notebooks to work?

Also, we need the following docs:

- [ ] docs/index.rst: add the name of the transformer
- [ ] docs/selection/index.rst: same
- [ ] docs/selection/new_rst_file_with_example.rst

Thanks a lot!!!

tests/test_selection/test_single_feature_performance_selection.py
NicoGalli: Ok, that makes completelly sense

tests/test_selection/test_single_feature_performance_selection.py
NicoGalli: Fixed

tests/test_selection/test_single_feature_performance_selection.py
NicoGalli: Ok, that makes totally sense.

tests/test_selection/test_single_feature_performance_selection.py
NicoGalli: Done!

tests/test_selection/test_single_feature_performance_selection.py
NicoGalli: Fixed!

solegalli:
closed in #173

PR #177, opened 2020-11-14
roll out type hint and test automation with circleci

solegalli:


solegalli:
@ChristopherGS 

mypy is returning error on base_transformers.py and variable_manipulation.py.

Could you have a quick look at the error and steer me in the right direction? I attach a copy:

errormypy

feature_engine/variable_manipulation.py
```diff
@@ -9,7 +9,7 @@
 
 def _define_variables(
     variables: Union[str, Optional[List[str]]]
-) -> Optional[List[str]]:
+) -> Union[str, Optional[List[str]]]:
```
solegalli, line 12: @ChristopherGS but the problem here is that this function does not return a string. It returns a list or None, That's why I don't understand the mypy error.

Okroshiashvili:
@solegalli there is type annotations mismatch. That's why there is an error. These functions are someway buggy because sometimes they return a string and sometimes they return None or a list of string.

solegalli:
Hi @ChristopherGS @Okroshiashvili 

I get this recurrent error, across all transformers, that tells me that the expression `variables`, which comes from the` init` and has Union of None, str, int, and lists is incompatible with `variables` as it is redefined during `fit`, which can only be a list of ints or strs. See last yellow box for an example.

With my meager experience on typehints, I can see 2 ways to "solve" this:
1) #type: ignore so that mypy ignores
2) in fit, instead of over-writing variables, I could assign another attribute variables_ (con: it will break backwards compatibility)

What's your views?

errormypy

feature_engine/variable_manipulation.py
```diff
@@ -2,14 +2,14 @@
 # License: BSD 3 clause
 # functions shared across transformers
 
-from typing import List, Optional, Union
+from typing import List, Optional, Union, Any
 
 import pandas as pd
 
+Variables = Union[None, int, str, List[Union[str, int]]]
```
ChristopherGS, line 9: @solegalli note that you can use typehint aliases like this (bit neater) https://docs.python.org/3/library/typing.html#type-aliases

feature_engine/variable_manipulation.py
```diff
@@ -2,14 +2,14 @@
...
 
-def _check_input_parameter_variables(
-    variables: Union[None, int, str, List[Union[str, int]]]
-) -> Union[None, int, str, List[Union[str, int]]]:
+# set return value typehint to Any here to avoid issues with the base transformer fit methods
+def _check_input_parameter_variables(variables: Variables) -> Any:
```
ChristopherGS, line 12: this is the key change I've made - not ideal, but I think it's more ideal than covering the code base in ignores and/or some kind of notable refactor.

PR #193, opened 2020-12-08
encoding jupyter notebooks in examples folder closes #59

### LINKED ISSUE #59: improve categorical encoding notebooks
Separate each categorical encoder into a single notebook, within the encoders folder. 

Check, if new parameters or functionality has been added and include examples within the notebooks.

ashok49473:


solegalli:
@ashok49473 thanks for the PR. This is looking really good. 

The only thing is that in the latest version of Feature-engine, which you will find in master but has not been released yet, the imports and class names are slightly different.

Now we import from `encoding` instead of `categorical_encoders`, and the classes don't take the word "Categorical" any more, so it would be MeanEncoder instead of MeanCategoricalEncoder

In order to get the lastest version, you need to clone the repo, cd into the repo folder and do  `pip install -e .`

More details on the installation here:
https://feature-engine.readthedocs.io/en/latest/contributing/github_quick_start.html

It would be great if you could update the notebooks to the latest version, which we will release next week :)

Alternatively, we can wait till the release next week, then you can install from pypi as usual and you will also have the documentation available as well. Up to you.

thank you!

ashok49473:
Hello @solegalli mam,
I updated the jupyter notebooks with the latest version

solegalli:
Hi @ashok49473 this is such a quick turn-around, thank you so much! 

Could I ask you the following:

1. I think it would be very useful for users to highlight that if they want to treat **numerical** variables, like **pclass** as if they were categorical, they would specifically need to **cast it as object** (as we do when we load the dataset), otherwise the transformer will  either send an error (if we pass it as argument) or not pick it up (if we leave `variables=None`). Could you add a line that explains / highlight this, in all notebooks? Perhaps in cell 6 of all notebooks where we have this `code: X[['cabin', 'pclass', 'embarked']].dtypes`  ? 

2. The Ordered ordinal encoder, MeanEncoder, WoE and PRatio have the characteristic that return monotonic variables, that is, encoded variables which values increase as the target increases. Could we add in each notebook a plot of the encoded variable (I would say either pclass or embarked, vs the target mean per category to show the monotonic relationship? Something like this:

image

3. The rare label encoder has new functionality since I made the notebook: 1 the user can change the string from 'Rare' to something else. 2. The user can choose to retain only the most popular categories with the argument max_n_categories. Could we add a demo for each of these in the notebook?
https://feature-engine.readthedocs.io/en/latest/encoders/RareLabelCategoricalEncoder.html

ashok49473:
@solegalli 
I updated the encoding jupyter notebooks with the above mentioned changes  :-)

I would like to add notebooks for other transformers also.

solegalli:
@ashok49473 you are awesome! thanks for the quick responses.

If you want, go ahead and add notebooks for other transformers (imputers, discretisers, whichever you like), and make a separate PR for each group of notebooks, while I review this code later today :)

Thanks a lot!

PR #204, opened 2020-12-27
update imputation notebooks closes #58

### LINKED ISSUE #58: improve imputation notebooks
Separate each missing data imputer into a single notebook, within the imputer folder. Check if new parameters or changes have been done to the imputers, and try to include a demo in the notebook.

pradumna123:


solegalli (changes requested):
Hi @pradumna123 

First of all, thank you very much for your contribution. And second, sorry for the delay in the review. I took a few days off over Christmas and also had to travel. It was madness.

I have a few suggestions and requests for this PR:

**General comments**

- In the import cell, instead of` import feature_engine.imputation as mdi,` could we just import the transformer we are going to use? say for example `from feature_engine.imputation import AddMissingIndicator`?  It is tidier, and we avoid making imports we do not need. This needs changing in all notebooks.
- could we remove the statement that prints the version? because this notebooks should be useful for future versions as well, so don't want to confuse readers
- in every notebook, in the top cell, where we have the description, could we turn around the content, so that first we describe what the transformer (imputer) is suppossed to do, and then we add a Note to say where the data comes from and how to obtain it? See for example this [notebook](https://github.com/solegalli/feature_engine/blob/master/examples/discretisation/EqualFrequencyDiscretiser.ipynb).
- whenever a description of the transformer is missing in the top cell (for example EndTailImputer), could you add a few lines explaining what the transformer is doing?


**AddNaNBinaryImputer.ipynb:**

- Could you please rename the notebook to the current name of the transformer? and also fix the text at the top of the notebook with the current name of the transformer?
- Could we also show the functionality of the argument missing_only? At the moment it is not there, and the idea of these notebooks is to show the user the full functionality of the classes

**ArbitraryNumberImputer.ipynb**

- this transformer has now the functionality to take 1 single value and impute all variables with it (which is shown in the notebook), and also take a dictionary of variable to number pairs, and impute each variable with the value in the dictionary. Could we add this demo to the notebook?
- also could you please in the import import only the class to use instead of mdi?

**CategoricalImputer.ipynb**
- this transformer, also allows the user to input the string with which they would like to replace NA, using the parameter fill_value. Could we add a demo, where the user inputs with a specific string, like for example "Other".

**EndTailImputer.ipynb**

- could you add a heading saying "Gaussian, right tail** before the implementation of the gaussian imputation?
- in the heading with the max, left tail, could we remove left tail? when the imputation_method is max, the argument tail is ignored. Thus, could we also remove the parameter tail from the instance where you create the transformer for the max imputation? (cell 20)

**RandomSampleImputer**
- could you add a bit of information about what the seed is doing under the heading Setting a "seed on variables, at observation level"? 

Thanks a lot for your support!

pradumna123:
I made those changes, Let  me know if I missed any.

solegalli:
@pradumna123 thanks for the changes. I made a few changes and a PR to your repo. If you merge that PR, then the files will be updated here as well and I can close this issue.

Thanks a lot!

solegalli:
Quick question, your name is Pradumna Suryawanshi? so I can add your contribution to the changelog?

pradumna123:
Yes thats right

> Quick question, your name is Pradumna Suryawanshi? so I can add your contribution to the changelog?

Yes that's right.

PR #212, opened 2020-12-31
update tansformations notebooks closes #62

### LINKED ISSUE #62: improve variable transformers jupyter notebooks
Separate each transformer into a new jupyter notebook. 

Check if new functionality has been added and include it in the notebook.

pradumna123:
I have added transformer notebooks.
Please let me know if  any changes are needed.


solegalli (changes requested):
Hi @pradumna123 

thank you so much for your help and for such quick response. I have some suggestions:

**general suggestions**
- in each notebook, could you add a plot of the variable before the transformation and after the transformation, so the user can see the effect before and after? For example, in the BoxCox notebook that there is no plot before the transformation.
- could you add a tiny comment in every cell before the command explaining what the command is going to do?
- can we change the headline "Selecting all variables" to "Automatically select numerical variables"

**BoxCoxTransformer**

- could you add a plot of the variable before the transformation, so the user can see the effect before and after?
- cell 5, before fit, could you add a comment: # in fit, the transformer learns the optimal parameters for the transformation
- cell 7: add a comment: # we now transform the variables
- I think the practice of adding +1 to the variables is a bit dangerous. I think it would be better for the user to apply yeojohnson instead of adding 1 to make it suitable for the box cox transformation. So, instead of working with the entire dataset, could we just work with a slice of the dataset that contains only positive values, and do the automatic selection there?
- also could you add a line to show the variables attribute that now contains all numerical variables and also the transformer_dict_ for this part as well?

**LogTransformer**
- same here, could we instead of adding 1 to all variables just work with a slice of the data that contains only positive variables?
- in the first part, could you add the parameter `base` to indicate whether it is log e or log 10? and maybe add a comment that we can change this to apply a different log. 
- In fact, could you apply log e when you pass those 2 variables and log 10 instead when you automatically select all variables?
- also, please add a comment in every cell before the command with a short explanation of what is the command going to do

**PowerTransformer**

- could you add the argument `exp`, so the user knows what exponent is being used?
- could you use different exp when passing the variables or automatically selecting the variables?
- could you implement the special cases of square root and cube root? exp 1/2 and 1/3?

**YeoJohnsonTransformer**
- there is a warning in cell 11, could you investigate what is raising it and if possible have  it removed?

Thanks a lot!

pradumna123:
I have added the changes for the notebooks can you have a look and let me know if something is missing

PR #215, opened 2021-01-02
adds drop NA transformer closes #183

### LINKED ISSUE #183: imputation: complete case analysis - drop observations with na
We need a transformer that would drop observations from the train / test set if they have na, in any or a subset of variables, as specified by the user.

Transformer works for both categorical and numerical variables

Useful to catch errors when deployed if na values appear.

Transformer is a bit dangerous, when deployed, because it will not allow users to know when variables come with na. Need to add a warning in the docs

pradumna123:
I have created a skeleton code for the NA transformer.
I have also added an example notebook in the imputation folder(in examples)

I have not added any comments yet. Let me know what you think.
 I can add comments and finalize the code.

And we can decide on the name too.


solegalli (commented):
Hi @pradumna123 

Thanks a lot for this contribution. The transformer looks pretty much done. Looks great.

I am thinking that a better name could be DropMissingData, could we change?

Could we add an additional method to just retain the data with NA? This might be useful when the model is in production, because sometimes you want to know which data is being dropped and not passed to the model?

Not sure how to call this method, maybe `retrieve_na`? Feel free to get more creative.

You probably know this, we need docstrings and tests. 

And also, it would be great if you could add the transformer in the:
- docs/index.rst
- an rst file inside docs/imputation for this transformer
- the rst file to the docs/imputation/index.rst

Finally, for the style test to pass, it needs to pass flake8, so if you install black and run black over the .py files you generate, that should sort it.

Thanks a lot!

feature_engine/imputation/na_transformer.py
```diff
@@ -0,0 +1,70 @@
...
+
+        return self
+
+
+    def transform(self, X: pd.DataFrame) -> pd.DataFrame:
+
+        X = self._check_transform_input_and_state(X)
+
+
+        X = X.copy()
```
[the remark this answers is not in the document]
pradumna123: I have changed the code  not create any more copies

pradumna123:
Regarding the "an additional method to just retain the data with NA .." The problem would be if the production env has a test/train dataset. We will have to save both test and train dataset in the transformer. 

What do you think should be done?
My thought save the latest transformed dataset

feature_engine/imputation/na_transformer.py
```diff
@@ -0,0 +1,79 @@
...
+
+        if not isinstance(missing_only, bool):
+            raise ValueError("missing_only takes values True or False")
+
+        self.variables = _check_input_parameter_variables(variables)
+        self.missing_only = missing_only
+
+
+    
+    def return_na(self):
```
solegalli, line 35: maybe call this method `return_dropped_data`

This method should be identical to transform(), except that instead of returning the data without NA, it would return the data with NA. So in essence it does the reciprocal of transform. 

So something like this:

```
def return_dropped_data(self, X: pd.DataFrame) -> pd.DataFrame:

        X = self._check_transform_input_and_state(X)

        id=pd.isnull(X[self.variables_]).any(1)
        id=id[id==True]
        return X[X.index.isin(id.index)]
```

solegalli: but actually, maybe instead of isin, use pandas loc, or iloc.

solegalli: is there not a way to retain observations with na with pandas, without the need to create an intermediate index?

feature_engine/imputation/na_transformer.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+    def transform(self, X: pd.DataFrame) -> pd.DataFrame:
+
+        X = self._check_transform_input_and_state(X)
+
+
+        ser=pd.isnull(X[self.variables_]).any(1)
+        new=ser[ser==True]
+        self.dropped_data_=X[X.index.isin(new.index)].copy()
```
solegalli, line 73: I would remove this 3 lines from here.

solegalli:
> 
> 
> Regarding the "an additional method to just retain the data with NA .." The problem would be if the production env has a test/train dataset. We will have to save both test and train dataset in the transformer.
> 
> What do you think should be done?
> My thought save the latest transformed dataset

I think we should not save anything, that saves memory. The user would pass the dataset with NA from which they wish to retain the data with NA. and the method just returns the data as transform does.

pradumna123:
I have added changes for the remove data method if everything looks good iI can start with documentation and test cases

solegalli:
> ave added changes for the remove data method if every

yes, the transformer looks good. please change the name of the class to DropMissingData. 

thanks a lot

pradumna123:
I am not sure why the style checks fail. When I run the black formatting tool on the feature engine directory it doesn't change any files.
Yet the style checks fail. I am not sure why.
Can you help me with this.

docs/index.rst
```diff
@@ -102,6 +102,7 @@ Missing Data Imputation: Imputers
 - :doc:`imputation/CategoricalImputer`: replaces missing data in categorical variables with the string 'Missing' or by the most frequent category
 - :doc:`imputation/RandomSampleImputer`: replaces missing data with random samples of the variable
 - :doc:`imputation/AddMissingIndicator`: adds a binary missing indicator to flag observations with missing data
+- :doc:`imputation/DropMissingDataImputer`: removes rows conaining NA values from dataframe
```
solegalli, line 105: should be `DropMissingData`, like the class name. Without the word `imputer`.

feature_engine/dataframe_checks.py
```diff
@@ -29,8 +29,9 @@ def _is_dataframe(X: pd.DataFrame) -> pd.DataFrame:
     """
```
solegalli, line 29: could you please remove this file from the PR.

This change is causing all tests to fail. There is no reason why this file should be changed.

feature_engine/dataframe_checks.py
```diff
@@ -29,8 +29,9 @@ def _is_dataframe(X: pd.DataFrame) -> pd.DataFrame:
...
 
     if not isinstance(X, pd.DataFrame):
-        raise TypeError("X is not a pandas dataframe. The dataset should be a "
-                        "pandas dataframe.")
+        raise TypeError(
+            "X is not a pandas dataframe. The dataset should be a " "pandas dataframe."
+        )
```
solegalli, line 34: also, please checkout the version of this file in master, it should work fine.

feature_engine/imputation/__init__.py
```diff
@@ -8,6 +8,7 @@
 from .end_tail import EndTailImputer
 from .missing_indicator import AddMissingIndicator
 from .random_sample import RandomSampleImputer
+from .na_transformer import Na_transformer
```
solegalli, line 11: please change the import so that it matches the .py and the class name

feature_engine/imputation/na_transformer.py
```diff
@@ -0,0 +1,192 @@
+# Authors: Pradumna Suryawanshi <pradumnasuryawanshi@gmail.com>
+# License: BSD 3 clause
+
+from typing import Optional, List, Union
+
+import pandas as pd
+
+from feature_engine.dataframe_checks import _is_dataframe
+from feature_engine.imputation.base_imputer import BaseImputer
+from feature_engine.parameter_checks import _define_numerical_dict
```
solegalli, line 10: you are not using this import I think. This will make style_checks fail. Please remove.

feature_engine/imputation/na_transformer.py
```diff
@@ -0,0 +1,192 @@
...
+from feature_engine.variable_manipulation import (
+    _check_input_parameter_variables,
+    _find_or_check_numerical_variables,
+)
+
```
solegalli, line 15: please double check that all imports are used. otherwise style checks fail.

feature_engine/imputation/na_transformer.py
```diff
@@ -0,0 +1,192 @@
+# Authors: Pradumna Suryawanshi <pradumnasuryawanshi@gmail.com>
```
solegalli, line 1: please change name of script to `drop_missing_data.py`

feature_engine/imputation/na_transformer.py
```diff
@@ -0,0 +1,192 @@
...
+
+class DropMissingData(BaseImputer):
+    """
+    The DropMissingData() will delete rows containing NA values.
+
+    DropMissingData() will delete rows containing NA values from the variables
+    indicated by the user, or variables with NA values in the train set.
+
+    The DropMissingData() works for both numerical and categorical variables.
+    The user can pass a list with the variables for which the missing indicators
```
solegalli, line 25: fix docstring. We are not adding missing indicators

feature_engine/imputation/na_transformer.py
```diff
@@ -0,0 +1,192 @@
...
+        Raises
+        ------
+        TypeError
+            If the input is not a Pandas DataFrame
+
+        Returns
+        -------
+        self.variables_ : list
+            The list of variables for which the rows with NA will be deleted.
+        """
```
solegalli, line 91: please remove this docstrings. we don't need docstrings after __init__

feature_engine/imputation/na_transformer.py
```diff
@@ -0,0 +1,192 @@
...
+        self.missing_only = missing_only
+
+    def return_dropped_data(self, X: pd.DataFrame) -> pd.DataFrame:
+        """
+        Returns the subset of the dataframe which contains rows with NA values.
+
+        Parameters
+        ----------
+        X : pandas dataframe of shape = [n_samples, n_features]
+            The dataset to be cleaned(Delete rows with NA).
```
solegalli, line 106: please remove (Delete rows with NA)

feature_engine/imputation/na_transformer.py
```diff
@@ -0,0 +1,192 @@
...
+
+        Raises
+        ------
+        TypeError
+            If the input is not a Pandas DataFrame
+
+        Returns
+        -------
+        X : pandas dataframe.
+            The cleaned version of dataset that does not contain rows with NA values.
```
solegalli, line 116: I would remove the word "cleaned"

feature_engine/imputation/na_transformer.py
```diff
@@ -0,0 +1,192 @@
...
+
+        return self
+
+    def transform(self, X: pd.DataFrame) -> pd.DataFrame:
+        """
+        Remove rows with NA values.
+
+        Parameters
+        ----------
+
```
solegalli, line 177: remove line between heading and  args

feature_engine/imputation/na_transformer.py
```diff
@@ -0,0 +1,192 @@
...
+        X : pandas dataframe of shape = [n_samples, n_features]
+            The dataframe to be transformed.
+
+        Returns
+        -------
+
```
solegalli, line 183: remove space

solegalli (changes requested):
Hey @pradumna123 

Hope my comments are helpful. All checks are failing now. Don't think it is a big issue. Please see my comments to get it sorted.

If the style checks continue to fail, I will have a look after you add the tests. All the rest looks really good.

Cheers
Sole

PR #234, opened 2021-02-02
Date-time transformer initial commit

pradumna123:
please review the code for date-time transformer.
I have an initial skeleton of code if that is fine I can add other features in the code.

### DIFF
--- a/feature_engine/transformation/date_transformer.py
+++ b/feature_engine/transformation/date_transformer.py
@@ -1,0 +1,49 @@
+from typing import List, Optional, Union
+
+import numpy as np
+import pandas as pd
+
+from feature_engine.base_transformers import BaseNumericalTransformer
+from feature_engine.variable_manipulation import _check_input_parameter_variables
+
+
+class date_feature_extractor(BaseNumericalTransformer):
+    def __init__(
+        self,
+        features: Union[None, int, str, List[Union[str, int]]] = None,
+        variables: Union[None, int, str, List[Union[str, int]]] = None,
+    ) -> None:
+
+        if not fearures:
+            raise ValueError(
+                "Feature list cannot be empty.Enter features to extract from dates"
+            )
+
+        self.variables = _check_input_parameter_variables(variables)
+        self.fearures = set(fearures)
+
+    def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
+
+        X = super().fit(X)
+
+        # check if all the variables are of type datetime.
+        if not all([pd.is_datetime64_any_dtype(X[col]) for col in X.columns]):
+            raise ValueError("Some variables are not datetime cannot transform data")
+
+        self.input_shape_ = X.shape
+
+        return self
+
+    def transform(self, X: pd.DataFrame) -> pd.DataFrame:
+        X = super().transform(X)
+        for col in self.variables:
+            if "week" in self.fearures:
+                X[col + "week"] = X[col].dt.week
+            if "month" in self.fearures:
+                X[col + "month"] = X[col].dt.month
+            if "quarter" in self.fearures:
+                X[col + "quarter"] = X[col].dt.quarter
+            if "semester" in self.fearures:
+                X[col + "semester"] = np.where(X[col].dt.quarter.isin([1, 2]), 1, 2)
+
+        return X


feature_engine/transformation/date_transformer.py
```diff
@@ -0,0 +1,49 @@
+from typing import List, Optional, Union
+
+import numpy as np
+import pandas as pd
+
+from feature_engine.base_transformers import BaseNumericalTransformer
+from feature_engine.variable_manipulation import _check_input_parameter_variables
+
+
+class date_feature_extractor(BaseNumericalTransformer):
```
solegalli, line 10: would it be better to call the class `DatetimeFeatureExtractor` ?

feature_engine/transformation/date_transformer.py
```diff
@@ -0,0 +1,49 @@
+from typing import List, Optional, Union
+
+import numpy as np
+import pandas as pd
+
+from feature_engine.base_transformers import BaseNumericalTransformer
```
solegalli, line 6: I don't think I would inherit this class. It has a check for numerical variables in the fit method that is not suitable here. And also, I would allow na, and make it optional for the user to say whether they want those to be raised or ignored.

feature_engine/transformation/date_transformer.py
```diff
@@ -0,0 +1,49 @@
...
+from feature_engine.variable_manipulation import _check_input_parameter_variables
+
+
+class date_feature_extractor(BaseNumericalTransformer):
+    def __init__(
+        self,
+        features: Union[None, int, str, List[Union[str, int]]] = None,
```
solegalli, line 13: could we rename this parameter to `features_to_extract`?

feature_engine/transformation/date_transformer.py
```diff
@@ -0,0 +1,49 @@
...
+        variables: Union[None, int, str, List[Union[str, int]]] = None,
+    ) -> None:
```
solegalli, line 15: could we add the following parameters?

- `keep_original = False`, which indicates if user wants to drop datetime variables after the transformation or not.
- `missing_values = 'ignore'`, which indicates if missing values should be ignored or alternatively, use 'raise' to throw an error when dataframe has Na

feature_engine/transformation/date_transformer.py
```diff
@@ -0,0 +1,49 @@
...
+
+        if not fearures:
```
solegalli, line 17: mind the typo all throughout the code, should be features not fearures. Also, please change name as indicated above.

This error should check that if the user passes a value different from None, the value is one of:
  "year",  "semester", "quarter", "month", "week",  "day", "day_of_week", "is_weekend", "hr", "min", "sec".

The trasnformer also should be able to return all of those features. The problem that we face is, to discriminate if the datetime variable has time and date component or only 1 of the 2.

feature_engine/transformation/date_transformer.py
```diff
@@ -0,0 +1,49 @@
...
+    def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
+
+        X = super().fit(X)
+
+        # check if all the variables are of type datetime.
+        if not all([pd.is_datetime64_any_dtype(X[col]) for col in X.columns]):
+            raise ValueError("Some variables are not datetime cannot transform data")
+
+        self.input_shape_ = X.shape
+
```
solegalli, line 34: this looks good, also maybe this fit methods works as inspiration/ template:

https://github.com/solegalli/feature_engine/blob/master/feature_engine/creation/combine_with_reference_feature.py#L198

feature_engine/transformation/date_transformer.py
```diff
@@ -0,0 +1,49 @@
...
+        self.variables = _check_input_parameter_variables(variables)
+        self.fearures = set(fearures)
+
+    def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
+
+        X = super().fit(X)
+
+        # check if all the variables are of type datetime.
+        if not all([pd.is_datetime64_any_dtype(X[col]) for col in X.columns]):
```
solegalli, line 30: is this the only datetime format allowed / understood by pandas?

solegalli: Also, it should check only the type of self.variables and not X.columns. We only want those to be transformed to be datetime

feature_engine/transformation/date_transformer.py
```diff
@@ -0,0 +1,49 @@
...
+            raise ValueError("Some variables are not datetime cannot transform data")
+
+        self.input_shape_ = X.shape
+
+        return self
+
+    def transform(self, X: pd.DataFrame) -> pd.DataFrame:
```
solegalli, line 37: look at the checks in this class for inspiration:
https://github.com/solegalli/feature_engine/blob/master/feature_engine/creation/combine_with_reference_feature.py#L257

feature_engine/transformation/date_transformer.py
```diff
@@ -0,0 +1,49 @@
...
+        for col in self.variables:
+            if "week" in self.fearures:
+                X[col + "week"] = X[col].dt.week
+            if "month" in self.fearures:
+                X[col + "month"] = X[col].dt.month
+            if "quarter" in self.fearures:
+                X[col + "quarter"] = X[col].dt.quarter
+            if "semester" in self.fearures:
+                X[col + "semester"] = np.where(X[col].dt.quarter.isin([1, 2]), 1, 2)
+
```
solegalli, line 48: can we expand to add all the other features?

solegalli (commented):
Hi @pradumna123 

thanks a lot for the contribution. I made a few comments throughout the code. 

The tests are failing, can it be that your master is behind the master of the repo? could you rebase uppstream master to the branch from which you are making the PR? not sure why those errors come up.

In this PR: #111 there is a script with some tests. See if those are useful, maybe you don't need to write them from scratch.

Thanks a lot.

Sole

feature_engine/transformation/date_transformer.py
solegalli: Also, could we place the script in a separate folder called datetime? not within the folder transformers?

PR #235, opened 2021-02-28
added outlier demo

ashok49473:
Added example jupyter notebooks for Winsorizer and  ArbitraryOutlierCapper


ashok49473:
Some checks were not successful. Can I know the reasons?

solegalli:
> 
> 
> Some checks were not successful. Can I know the reasons?

There is an difference in the last decimal of a check in one of the tests. This happens because python floats are pseudo-random, so, we should check rounded values instead of the entire float as we do in most tests. This needs fixing. 

To have this test pass, on line 141 of the test_quantile_capping_both_tails_with_fold_15_percent in test_winsorizer.py
[this link](https://github.com/solegalli/feature_engine/blob/master/tests/test_outliers/test_winsorizer.py#L141)

we need to replace: 
`assert transformer.right_tail_caps_ == {"var": 0.11823196128033647}`

by:
`assert np.round(transformer.right_tail_caps_["var":], 5) == np.round(0.11823196128033647, 5)`

and make sure we import numpy at the top if not already. Can you add this please?

solegalli (changes requested):
Hi @ashok49473 

thank you so much for the notebooks! Could I ask for some changes?

Could you drop the variables 'name', 'ticket' from the loading function in cell 2 instead of from train_test_split in cell 3 in all notebooks?

could you add comments before each line of code, indicating what the code will do? 

could you add some spaces between lines of code and between parameters and values? I think the function autopep8 or prettify from the notebook can do this automatically for you, if you have them installed
Also, outlier trimmer and winzorizer are not the same, the trimmer removes / deletes the observations with outliers, so that they will no longer be in the dataset. The winsorizer replaces the values by the cap. In the trimmer notebook, we need to show that the size of the dataset changes.

Thanks a lot

ashok49473:
Hello @solegalli, Is it necessary to pass all the checks?

solegalli (changes requested):
Hi @ashok49473 great work!!! This is almost ready.

Can I ask 2 tiny changes to merge?

1) test style is failing because the lane 141 in test_winzoriser is too long. It has to contain a maximum of 88 characters, and now it contains 95. I think you can shorten it by introducing the following change:
instead of:
`assert np.round(transformer.right_tail_caps_["var"], 5) == np.round(0.11823196128033647, 5)`
change to:
` assert np.round(transformer.right_tail_caps_["var"], 5) == np.round(0.1182319, 5)`

Finally, in the notebooks of winsorizer and outlierTrimmer, in the percentile section (cells 23 and 21, respectively), you write:

percentiles or quantiles:

The limits are given by the percentiles.

    right tail: 95th percentile
    left tail: 5th percentile

but then in the class you use fold=0.02. 

For a 95th and 5th percentile, folds need to be 0.05. If fold is 0.02, then you are doing 98th, 2th percentile.

Can you fix the writing to:

percentiles or quantiles:

The limits are given by the percentiles.

    right tail: **98**th percentile
    left tail: **2**th percentile

In both notebooks?

With these changes, it should be ready to go. Thanks a lot!

solegalli:
@ashok49473 all done, thank you very much!!!

Issue #230, opened 2021-01-24
Ciclycal Features.


hectorpatino:
Hi @solegalli, I was practicing on kaggle the cat_on_dat_II challenge and I found that some features may be ciclycal, it was the first time I read that term. So according to  [this blog](http://blog.davidkaleko.com/feature-engineering-cyclical-features.html#:~:text=Hours%20of%20the%20day%2C%20days,each%20other%20and%20not%20far.) It is a way to treat some numerical variables when the max number of a variable is close to the min number and it is a very common mistake. I checked the issues but didn´t see something related to it, so it will be nice to have that kind of transformer.
I'm not that good in programming but if you think this is worth doing I may work a while on it.

solegalli:
Hi @hectorpatino curiously, I also saw a Kaggle notebook last week that worked with cyclical features and took a mental note to have a look at some point.

Happy for you to give it a go. I link here a couple of Kaggle notebooks for reference. Do you know of any article that discusses this encoding?

Not sure in which folder to add this transformer. I don't think it belongs to transformation, maybe a separate module? thoughts

https://www.kaggle.com/mansipriya/handling-cyclic-data?scriptVersionId=31108111

https://www.kaggle.com/danofer/cyclical-weather-features-for-deep-learning

https://www.kaggle.com/avanwyk/encoding-cyclical-features-for-deep-learning

https://datascience.stackexchange.com/questions/5990/what-is-a-good-way-to-transform-cyclic-ordinal-attributes

hectorpatino:
Well @solegalli I haven't do a serious research about it, but I will check your links and look some academical papers. 
I think the folder should be transformation as there are Log, Reciprocal and other mathematical transformations. In this case any variable will be mapped into 2 new features being: var_sin, var_cos. So according to this, I think it is the best option.

solegalli:
great. If you would like to give it a go, make a PR with some code and we take it from there. Probably the easiest would be to take the skeleton of the other transformers as template. Cheers

PR #239, opened 2021-03-17
Ciclycal transformer

hectorpatino:
Hi this is the solution I propose for Issue #230 I tried to keep all your conventions. Let me know what you think of it.


feature_engine/transformation/ciclycal.py
```diff
@@ -0,0 +1,122 @@
...
+    """
+
+    def __init__(
+            self, variables: Union[None, int, str, List[Union[str, int]]] = None
+    ) -> None:
+        self.variables = _check_input_parameter_variables(variables)
+
+    def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
+        """
+        This transformer does not learn any parameter. But keeps a map of the max values
```
solegalli, line 59: Does this not learn the maximum value to be able to convert the features into cyclical?

feature_engine/transformation/ciclycal.py
```diff
@@ -0,0 +1,122 @@
...
+        self.input_shape_ = X.shape
+
+        # check for nans
+        self.max_values_ = {}
+        for variable in self.variables:
+            if X[variable].isna().sum() > 0:
+                raise ValueError(f'The transformer CiclycalTransformer '
+                                 f'does not allow to have NaN values, please '
+                                 f'check the column {variable}')
+            self.max_values_[variable] = X[variable].max()
```
solegalli, line 95: Instead of the loop, could we change the presence of NA, with a feature engine function: 
_`check_contains_na(X, self.variables)`?

Also, can we make a dictionary with 1 line instead:
`self.max_values_ = X[self.variables].max().to_dict()` ?

Finally, I see in the article that it substracts 1 to months because their scale is 1-12, but it does not subtract 1 to hr, because they are already starting in 0, so 0-23, instead of 1-24. 

How does this affect the cyclical transformation?

Should we not also check for the minimum value and if it is not 0, subtract to reindex in 0?

feature_engine/transformation/ciclycal.py
```diff
@@ -0,0 +1,122 @@
...
+            If some of the columns contains NaNs
+        """
+        X = super().transform(X)
+
+        for variable in self.variables:
+            max_value = self.max_values_[variable]
+            X[f'{variable}_sin'] = np.sin(X[variable] * (2. * np.pi / max_value))
+            X[f'{variable}_cos'] = np.cos(X[variable] * (2. * np.pi / max_value))
+            X.drop(columns=variable, inplace=True)
+
```
solegalli, line 121: instead of dropping the original features by default, could we add a parameter in init called `drop_original=False`, where the user has the chance to decide if they want to drop or not?

To be honest, I have used months and day of weeks in my predictions, so even if these features add prediction power, I would be hesitant to discard them straightaway.

tests/conftest.py
```diff
@@ -174,3 +174,30 @@ def df_normal_dist():
...
+        ],
+        'months': [
+            3,
+            7,
+            9,
+            12,
+            4,
+            6,
+            12,
+        ]
```
solegalli, line 199: following my comment on whether to index from 1 or 0, or how to work with that... not sure this example needs modifying

solegalli (commented):
Hi @hectorpatino 

Thank you so much for this PR, this is looking really good.

I have a question regarding the technique mostly, in the blog and kaggle kernels, they subtract 1 from months and nothing from time, because the first one starts at 1, whereas the second at 0.

Should features start at 0? why or why not?

thinking if we should offer another param called, reset_scale=False to reset the variables at 0

Also, the transformer as it stands, assumes that in the train set, we find all possible values of the feature, and thus, the max is the cycle period. What if this is not the case? What if in the train set there are only 11 months? should we offer the user the possibility of entering the values for the cycle? 

Finally, I think this transformer works best in the creation folder, because we are in essence creating 2 new features from existing 1, and offer the user the possibility to not drop the original variables. Could we move the file there?

I also made a few comments in the code

Thanks a lot!

feature_engine/transformation/ciclycal.py
```diff
@@ -0,0 +1,122 @@
+from typing import List, Optional, Union
```
solegalli, line 1: `cyclical.py` instead of ciclycal.py as file name?

feature_engine/transformation/ciclycal.py
hectorpatino: Yeah you are right I forgot to change docstrings after i implemented the maximum value.

feature_engine/transformation/ciclycal.py
hectorpatino: I will check this. Not sure how does it affects the ciclycal transformation.

feature_engine/transformation/ciclycal.py
hectorpatino: `drop_original` was at the beggining, but I removed it (to be honest I don´t remember by I said no). But Sure! I can bring it to live again.

feature_engine/transformation/ciclycal.py
hectorpatino: OK!

hectorpatino:
> Hi @hectorpatino
> 
> Thank you so much for this PR, this is looking really good.
> 
> I have a question regarding the technique mostly. In the blog and kaggle kernels, they subtract 1 from months and nothing from time, because the first one starts at 1, whereas the second at 0.
> 
> Should features start at 0? why or why not?
> 
> thinking if we should offer another param called, reset_scale=False to reset the variables at 0
> 
> Also, the transformer as it stands, assumes that in the train set, we find all possible values of the feature, and thus, the max is the cycle period. What if this is not the case? What if in the train set there are only 11 months? should we offer the user the possibility of entering the values for the cycle?
> 
> Finally, I think this transformer works best in the creation folder, because we are in essence creating 2 new features from existing 1, and offer the user the possibility to not drop the original variables. Could we move the file there?
> 
> I also made a few comments in the code
> 
> Thanks a lot!

I've been thinking in the 'why or why not?' part. I think you are right but maybe we are looking at a very small example (datetime) for this. In GIS (which is my area of expertise) there is something called [aspect](https://desktop.arcgis.com/en/arcmap/10.3/tools/spatial-analyst-toolbox/how-aspect-works.htm) that identifies the downslope direction of the maximum rate of change in value from each cell to its neighbors. In hour and aspect cases the values begin at 0, but not in the case of months, nevertheless, there maybe a plethora of examples unknown for as that do not follow this condition, so I don´t know if rescaling them its a good idea. 

On the other hand I think that the transformer needs the mapping of the max cycle period. What you think of a parameter like this one on `__init__` `max_values_mapping={'day': 23, 'month':12}`

solegalli:
Hi @hectorpatino 

Thinking it further, I agree with you that re-scaling may not be needed. Sounds like an overkill. So let's leave that one out. Thanks for the further thoughts on this.

Regarding the additional parameter, we could call it `max_values` and if left to `None`, then `self.max_values_` automatically detects the maximums. Alternatively, the user can pass a dictionary of `variable name: max value`.

How does that sound?

Thanks a lot!

Sole

hectorpatino:
Hi @solegalli

This are the implementations of your reviews.

feature_engine/creation/cyclical.py
```diff
@@ -0,0 +1,140 @@
...
+    """
+
+    def __init__(
+            self, variables: Union[None, int, str, List[Union[str, int]]] = None,
+            max_values: Optional[Dict[str, Union[int, float]]] = None,
+            drop_original: Optional[bool] = False
+    ) -> None:
+        self.variables = _check_input_parameter_variables(variables)
+        self.max_values = max_values
+        self.drop_original = drop_original
```
solegalli, line 66: these 2 parameters need checks that they receive the allowed values

feature_engine/creation/cyclical.py
```diff
@@ -0,0 +1,140 @@
...
+            - If some of the mapping keys are not present in variables.
+
+        Returns
+        -------
+        self
+        """
+
+        # check input dataframe
+        X = super().fit(X)
+        _check_contains_na(X, self.variables)
```
solegalli, line 96: BaseNumericalTransformer already checks for na, so this line is duplicated / not needed:
https://github.com/solegalli/feature_engine/blob/master/feature_engine/base_transformers.py#L62

feature_engine/creation/cyclical.py
```diff
@@ -0,0 +1,140 @@
...
+        """
+        X = super().transform(X)
+
+        for variable in self.variables:
+            max_value = self.max_values_[variable]
+            X[f'{variable}_sin'] = np.sin(X[variable] * (2. * np.pi / max_value))
+            X[f'{variable}_cos'] = np.cos(X[variable] * (2. * np.pi / max_value))
+            if self.drop_original:
+                X.drop(columns=variable, inplace=True)
+
```
solegalli, line 139: could we remove the indentation and remove all variables in one go, instead of inside the loop?

tests/conftest.py
```diff
@@ -174,3 +174,30 @@ def df_normal_dist():
     df.columns = ["var"]
 
     return df
+
+
+@pytest.fixture(scope="module")
+def df_ciclycal_trans():
```
solegalli, line 180: These dataframes will solely be used for the cyclical transformer, so instead of in the config, let's put them in the same test.py file with the tests.

solegalli, line 180: cyclical instead of ciclycal?

tests/test_creation/test_cyclical_transformer.py
```diff
@@ -0,0 +1,189 @@
+import pandas as pd
+import pytest
+
+from feature_engine.creation import CyclicalTransformer
+
+
+def test_general_transformation_without_dropping_values(df_ciclycal_trans):
```
solegalli, line 7: dropping variables? cyclical?

tests/test_creation/test_cyclical_transformer.py
```diff
@@ -0,0 +1,189 @@
...
+    assert ciclycal.input_shape_ == (7, 2)
+    assert ciclycal.max_values_ == {
+        'day': 7,
+    }
+
+    # test transform output
+    pd.testing.assert_frame_equal(X, transf_df)
+
+
+def test_general_transformation_dropping_original_values(df_ciclycal_trans):
```
solegalli, line 47: variables instead of values?

solegalli (commented):
Hi @hectorpatino 

The code is almost ready. Just a few suggestions to remove or re-accomodate some lines of code, and the tests contain a few typos, if you could have a look at those.

Would you also be able to add some docs? We need these:

In this folder:
https://github.com/solegalli/feature_engine/tree/master/docs/creation

create an rst file with a demo for these transformer, and add the transformer name to the index rst in the same folder.

Also add the transformer name and a very short description to the main page of the documentation here:
https://github.com/solegalli/feature_engine/blob/master/docs/index.rst

And if you are feeling in the mood, add the release of a new version here:
https://github.com/solegalli/feature_engine/blob/master/docs/whats_new/v1.rst

with the name of the new transformer and yourself as a contributor.

Would these be possible?

Thanks a lot!

hectorpatino:
Gonna work on this a little bit this weekend. I have never done .rst files so I would seek for help if something happens. Is there a command to execute for them?

hectorpatino:
I´m feeling a little bit lost here. I've finished all your code reviews and I was working on documentation, but as this is the very first time I try to contribute to some project I have some doubts. I was trying to do the same kind of doc that `ReciprocalTransformer()` but I saw it uses a dataframe which reads `houseprice.csv`. I didnt found that file anywhere. I used pycharm find in path-> in project and it didn´t found anything. So I guess Im missing something here.

solegalli:
> 
> 
> I´m feeling a little bit lost here. I've finished all your code reviews and I was working on documentation, but as this is the very first time I try to contribute to some project I have some doubts. I was trying to do the same kind of doc that `ReciprocalTransformer()` but I saw it uses a dataframe which reads `houseprice.csv`. I didnt found that file anywhere. I used pycharm find in path-> in project and it didn´t found anything. So I guess Im missing something here.

Hi @hectorpatino, the houseprices.csv is the data set from the house prices competition from Kaggle. I used this originally, but looking forward I am not sure it is the best option, because for the user to be able to run the code, they would have to go to kaggle, download and rename the dataset first. Which is quite annoying.

I think, the best is to try and use one of the datasets already available in sklearn, but I am not sure they have cyclical features, so as an alternative, you can create a toy dataframe for the demo.

Are you in the mood of giving the docs a go? or would you rather I do it?

solegalli:
> 
> 
> Gonna work on this a little bit this weekend. I have never done .rst files so I would seek for help if something happens. Is there a command to execute for them?

In this link you find info on how to run the docs:
https://feature-engine.readthedocs.io/en/latest/contribute/building_the_docs.html

hectorpatino:
> > I´m feeling a little bit lost here. I've finished all your code reviews and I was working on documentation, but as this is the very first time I try to contribute to some project I have some doubts. I was trying to do the same kind of doc that `ReciprocalTransformer()` but I saw it uses a dataframe which reads `houseprice.csv`. I didnt found that file anywhere. I used pycharm find in path-> in project and it didn´t found anything. So I guess Im missing something here.
> 
> Hi @hectorpatino, the houseprices.csv is the data set from the house prices competition from Kaggle. I used this originally, but looking forward I am not sure it is the best option, because for the user to be able to run the code, they would have to go to kaggle, download and rename the dataset first. Which is quite annoying.
> 
> I think, the best is to try and use one of the datasets already available in sklearn, but I am not sure they have cyclical features, so as an alternative, you can create a toy dataframe for the demo.
> 
> Are you in the mood of giving the docs a go? or would you rather I do it?



> > Gonna work on this a little bit this weekend. I have never done .rst files so I would seek for help if something happens. Is there a command to execute for them?
> 
> In this link you find info on how to run the docs:
> https://feature-engine.readthedocs.io/en/latest/contribute/building_the_docs.html

No I'm gonna finish this PR from top to bottom as I said this is my first ever PR and I would like to be envolve in all the cycle; just 1 question. Where do I save the toy dataset?

solegalli:
Hi @hectorpatino 

I like the attitude :)

Toy datasets are usually small, so you don't really need to save it anywhere, it is created and stored in memory as you proceed with the example. 

It can be something as simple as this:
https://scikit-learn.org/stable/modules/impute.html#univariate-feature-imputation

So I think the simplest is if you expand a little bit the datasets that you used in the test script, and in the rst include as well some outputs of the transformed and un-transformed variables.

solegalli:
Hi @hectorpatino 

How are you? hope all is well and your are healthy.

We are almost there with this PR. Would you know when you would find the time to finish it of?

Thank you!

hectorpatino:
Hi @solegalli I´ve been really busy this two weeks, can you wait until weekend?

hectorpatino:
@solegalli Ayudame Sole!!! Im not able to see what the error is accordding to circleci `/home/circleci/project/docs/whats_new/v1.rst:20:Title level inconsistent:` this is the error but I just don´t see it. What am I doing wrong?

docs/whats_new/v1.rst
```diff
@@ -1,3 +1,17 @@
+Version 1.0.3
+=============
+
+Deployed: 17th April 2021
+
+Contributors
+____________
+    - Hector Patino
```
solegalli, line 8: Can it be that the underline needs to be ------- instead of the continuous line? rst is quite adamant about what you use to mark titles and subtitles @hectorpatino

solegalli: > 
> 
> @solegalli Ayudame Sole!!! Im not able to see what the error is accordding to circleci `/home/circleci/project/docs/whats_new/v1.rst:20:Title level inconsistent:` this is the error but I just don´t see it. What am I doing wrong?

Hice un PR a tu repo con algunos cambios :)

PR #241, opened 2021-04-02
round numbers in test_winsorizer.py

indymnv:
I rounded numbers for test_winsorizer.py, for each one I use 3 decimals and I added for all comparisons. I use tox and pytest and passed successfully. If you think that is ok, I can apply the same rule to the rest of the test files.


solegalli:
Hi @indymnv 

Yes, this looks great. Thank you! Happy for you to apply to the rest of the test files.

The only thing, it seems that in the commit there are a bunch of files called 0.11.1, 0.22.2 etc, those should not be part of the commit. Not sure what they are. Could you please remove them?

Also, if you are feeling brave, the style test is failing, probably because the lines are too long (more than 88 char) or there are more than the allowed spaces between numbers and =.

You can test the errors with flake8, and automatically correct them with black. Both flake8 and black are python packages that can be installed from pip).

Thanks a lot!

indymnv:
Sure, for now, I will do another commit with the rest of the test files.  Later I will check the style test

Issue #233, opened 2021-01-31
allow  MathematicalCombination to take in null values


solegalli:
introduce parameter to be able to select whether to ignore or raise the missing values

AgustinFi:
Hi @solegalli 
I have time to work on this request.

solegalli:
That's great! Thanks!.

If you want, go ahead and make a PR with the suggested changes.

Cheers
Sole

PR #243, opened 2021-04-11
#233 MathematicalCombination to take in null values

AgustinFi:
Optional parameter _**[skipna: Optional[bool] = True]**_ introduced to allow users the ability to provide variables containing nulls. If _skipna_ is set to True, the following must be taken into account:

1. It won't raise a value error.
2. Pandas will try to perform operations according to its internal logic. No extra null handling logic has been implemented. Results might be numbers or NaN. Ex: Sum(1,NaN) = 1 while Std(1, NaN) = NaN

### DIFF
--- a/feature_engine/creation/mathematical_combination.py
+++ b/feature_engine/creation/mathematical_combination.py
@@ -86,7 +86,7 @@ class MathematicalCombination(BaseEstimator, TransformerMixin):
         If `new_variable_names = None`, the transformer will assign an arbitrary name
         to the newly created features starting by the name of the mathematical
         operation, followed by the variables combined separated by -.
-        
+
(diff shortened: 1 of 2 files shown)


solegalli:
Hi @AgustinFi 

Thanks for the contribution!!

`fit `and `transform` methods can only take parameters `X` and `y`, to make the class compatible with the sklearn pipeline. 

So the `skipna` method should be passed onto the `init`. And it will apply both to fit and transform, so it should be stored as an attribute.

Could you remake the files with this change? Also, we have this parameter in other classes (see for example outlier classes). Could we use the same names? that makes it easier to users.

Thanks a lot!

Sole

solegalli:
@AgustinFi any thoughts on this?

Issue #191, opened 2020-12-07
one hot encoder has the option to automatically drop the second binary variable for binary categorical variables


solegalli:
We need to add one parameter to the init method of the one hot encoder, called for example automate_drop_last_binary that when True, will automatically return 1 binary variable if the original variable was binary.

Details:
- automate_drop_last_binary should default to False not to break backwards compatibility
- automate_drop_last_binary would only work for binary variables
- automate_drop_last_binary would only work if drop_last=False

When automate_drop_last_binary=True, binary variables would result in the addition of just one dummy instead of 1. So for the variable colour, with categories black and white, instead of returning the binary colour_black and colour_white, which are redundant, it would only return colour_black

In addition, we should add an attribute in the fit method called for example binary_variables_ where the transformer returns a list with the names of the binary variables.

To finish of this PR, we need to add 1-2 tests to make sure that the behaviour is the expected

solegalli:
I would recommend starting the chances in a new PR

PR #247, opened 2021-04-17
Check one hot encoder for variables with 2 categories. #191

vasusuren:
#191 Fixed the bug that creates two dummy variables when there are two categories. Added test case and a new test dataframe

### DIFF
--- a/feature_engine/encoding/one_hot.py
+++ b/feature_engine/encoding/one_hot.py
@@ -112,10 +112,10 @@ class OneHotEncoder(BaseCategoricalTransformer):
     """
 
     def __init__(
-        self,
-        top_categories: Optional[int] = None,
-        variables: Union[None, int, str, List[Union[str, int]]] = None,
-        drop_last: bool = False,
+            self,
+            top_categories: Optional[int] = None,
+            variables: Union[None, int, str, List[Union[str, int]]] = None,
+            drop_last: bool = False,
     ) -> None:
 
         if top_categories and not isinstance(top_categories, int):
@@ -176,10 +176,10 @@ class OneHotEncoder(BaseCategoricalTransformer):
                     self.encoder_dict_[var] = [
                         x
                         for x in X[var]
-                        .value_counts()
-                        .sort_values(ascending=False)
-                        .head(self.top_categories)
-                        .index
+                            .value_counts()
+                            .sort_values(ascending=False)
+                            .head(self.top_categories)
+                            .index
                     ]
                 elif len(X[var].unique()) == 2:
                     self.encoder_dict_[var] = [
--- a/feature_engine/encoding/one_hot.py
+++ b/feature_engine/encoding/one_hot.py
@@ -169,17 +169,27 @@ class OneHotEncoder(BaseCategoricalTransformer):
                     category_ls = [x for x in X[var].unique()]
                     self.encoder_dict_[var] = category_ls[:-1]
                 else:
-                    self.encoder_dict_[var] = X[var].unique()
+                    self.encoder_dict_[var] = X[var].unique() if len(X[var].unique()) > 2 else X[var].unique()[0]
 
             else:
-                self.encoder_dict_[var] = [
-                    x
-                    for x in X[var]
-                    .value_counts()
-                    .sort_values(ascending=False)
-                    .head(self.top_categories)
-                    .index
-                ]
+                if self.top_categories > 2:
+                    self.encoder_dict_[var] = [
+                        x
+                        for x in X[var]
+                        .value_counts()
+                        .sort_values(ascending=False)
+                        .head(self.top_categories)
+                        .index
+                    ]
+                else:
+                    self.encoder_dict_[var] = [
+                        x
+                        for x in X[var]
+                            .value_counts()
+                            .sort_values(ascending=False)
+                            .head(1)
+                            .index
+                    ]
 
         self._check_encoding_dictionary()
 
--- a/feature_engine/encoding/one_hot.py
+++ b/feature_engine/encoding/one_hot.py
@@ -169,27 +169,27 @@ class OneHotEncoder(BaseCategoricalTransformer):
                     category_ls = [x for x in X[var].unique()]
                     self.encoder_dict_[var] = category_ls[:-1]
                 else:
-                    self.encoder_dict_[var] = X[var].unique() if len(X[var].unique()) > 2 else X[var].unique()[0]
+                    self.encoder_dict_[var] = X[var].unique() \
+                        if len(X[var].unique()) > 2 \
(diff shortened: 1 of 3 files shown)


feature_engine/encoding/one_hot.py
```diff
@@ -169,18 +169,27 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
...
-                self.encoder_dict_[var] = [
-                    x
-                    for x in X[var]
-                    .value_counts()
-                    .sort_values(ascending=False)
-                    .head(self.top_categories)
-                    .index
-                ]
-
+                if self.top_categories >= 2 and len(X[var].unique()) > 2:
```
solegalli, line 177: can we not do here the same as above?  without the if and elif simply:

```
self.encoder_dict_[var] = [
                    x
                    for x in X[var]
                    .value_counts()
                    .sort_values(ascending=False)
                    .head(self.top_categories)
                    .index
                ] \
 if X[var].nunique() >2 \
else ...
```

?

Binary variables should always be encoded as 1 dummy, regardless of whether the user chooses 2 as top categories. The second variable is entirely redundant.

in my humble opinion :)

solegalli (commented):
Hi @vasusuren thanks a lot for the contribution, and sorry for the delayed response. 

Nicely done, thanks!

Would you mind having a look at my comments / questions?

feature_engine/encoding/one_hot.py
```diff
@@ -112,10 +112,10 @@ class OneHotEncoder(BaseCategoricalTransformer):
...
     def __init__(
-        self,
-        top_categories: Optional[int] = None,
-        variables: Union[None, int, str, List[Union[str, int]]] = None,
-        drop_last: bool = False,
+            self,
+            top_categories: Optional[int] = None,
+            variables: Union[None, int, str, List[Union[str, int]]] = None,
+            drop_last: bool = False,
     ) -> None:
```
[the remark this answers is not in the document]
vasusuren: That was not on purpose. I suppose it would have leaked in due to automatic formatting settings per pep-8 in pycharm (my IDE). I can de-indent it. I noticed it after the push and since it passed the flake8 check I left it as is, sorry about that.

feature_engine/encoding/one_hot.py
```diff
@@ -169,18 +169,27 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
                     category_ls = [x for x in X[var].unique()]
                     self.encoder_dict_[var] = category_ls[:-1]
                 else:
-                    self.encoder_dict_[var] = X[var].unique()
+                    self.encoder_dict_[var] = X[var].unique() \
+                        if len(X[var].unique()) > 2 \
```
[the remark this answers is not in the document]
vasusuren: Yes, that sounds and reads a lot better. I learnt something new method (nunique) today.

feature_engine/encoding/one_hot.py
vasusuren: My initial feeling was also similar. But when I thought about it I felt it would add more information to have 2 columns when 2 top categories are selected.
Column 1 --> Category 1 encoded as 1, rest 0
Column 2 --> Category 2 encoded as 1, rest 0
If we go ahead with one dummy both category 1 and category 2 will be coded as 1 and rest as 0. Eventually, we lose information which rows has category 1 and which has category 2. I felt this would be an useful information when someone is purposefully requesting top 2 categories. However, the code should create only one dummy for column that contains only two categories instead of more than 2.

Please let me know what you think. I will wait for your comments before making changes above.

solegalli: Hello again,

Thinking out loud:

From a mathematical perspective, if a variable has 2 categories, say male and female, then one will be 1 and the other 0, and it does not really matter which is which.

If we derive 2 dummies from a categorical variable with 2 categories, say male and female, the 2 dummies will be 100% (negatively) correlated. Thus, the second variable does not add absolutely anything respect to the first one.

Because of these, I think that binary categorical variables should be encoded as 1 dummy.

We will preserve the information because the variable will be called say `gender_male`, so we know that male takes 1 and female takes 0. If it was the other way around, say `gender_female`, then female takes 1 and male takes 0. So we do know which one is which, after second thought.

There is 1 risk, thinking it further...

what if in the training set the variable has 2 categories, say male and female, and in the test set, there is a third category, say non_binary? 

In the above case, non_binary will be encoded as female by default by the way the transformer is set up. Whereas if we have 2 dummies, it will be 0 in both, so it will be captured.

Can we tackle this somehow? I wonder whether it makes sense to add 1 additional parameter in the init, to ask the user if it wants 1 or 2 dummies from binary categorical variables?

vasusuren: Hello again,
Yes, I agree with you 100%. 

When the variable contains only 2 categories (i.e., binary categories0 - it will be represented by one dummy variable - this is taken care of in the "elif" part of code.

However, if the variable has more than 2 categories and the user purposefully requests top two categories to be hot-coded, I prefer to have two dummies. One for the first category, Two for the second category, and together for the rest of the category(ies). This part is handled in the "IF" the user requests top two categories and the variable has more than two categories. Please note: I added few additional properties to code in further commits.

Regarding your comment on test data having more than 2 categories is a different animal and we can handle that in a different pull request to make sure we are making incremental changes. In this pull request, we are making sure if we have a binary categorical variable, we create only one dummy variable.

solegalli:
Hi @vasusuren did you see this PR to your repo? https://github.com/vasusuren/feature_engine/pull/1

I made a few changes on the docstrings.

Not sure how straightforward the merge would be

feature_engine/encoding/one_hot.py
solegalli: The `top_categories` argument is intended **only** to discern if the binary variables should be derived for all categories or only for the most frequent ones. So we shouldn't add "extra" or "different" functionality here, respect to when it is set to `None`.

I think we have the following dilemma:

With OHE, categorical variables with 2 categories will be encoded into 2 dummy variables, the second one being completely redundant, except when in the test set, new "unseen" categories appear, in which case we need the 2 dummies to carry a 0.

In the current version of the transformer, the user can take care of removing the second dummy variable by setting `drop_last=True`.

This PR is intended to automate the dropping of the second variable for all categorical variables with only 2 categories.

The plus:
it will automatically remove redundant variables

The con:
potentially unseen labels in the test set will be encoded as the second category.

The question is:
is the plus big enough to compensate the con? or should we leave the removal of the second variable to the user?

if the plus is big enough, then we should go ahead with this transformer and see how we can prevent the unadverted user of the risk

if the plus is not big enough, then we should not add this functionality

Thoughts?

PR #254, opened 2021-04-30
modify np.float to float

AndrewTanQB:
### Description

This is a simple PR to change the use of numpy aliases for builtins like `float`.

It eliminates deprecation warnings for numpy >= 1.20, e.g.
```
***/feature_engine/encoding/count_frequency.py:122: DeprecationWarning: `np.float` is a deprecated alias for the builtin `float`. To silence this warning, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here.
  Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
    n_obs = np.float(len(X))
```

See numpy 1.20 release notes on deprecations

https://numpy.org/doc/stable/release/1.20.0-notes.html

### DIFF
--- a/feature_engine/encoding/count_frequency.py
+++ b/feature_engine/encoding/count_frequency.py
@@ -119,7 +119,7 @@ class CountFrequencyEncoder(BaseCategoricalTransformer):
                 self.encoder_dict_[var] = X[var].value_counts().to_dict()
 
             elif self.encoding_method == "frequency":
-                n_obs = np.float(len(X))
+                n_obs = float(len(X))
                 self.encoder_dict_[var] = (X[var].value_counts() / n_obs).to_dict()
 
         self._check_encoding_dictionary()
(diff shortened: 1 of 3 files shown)


solegalli:
Hi @AndrewTanQB 

Thank you so much for this PR! 

The test_style is failing because numpy, which now is not used any more because we changed np.float by float, needs to be removed from the imports, see:

```
feature_engine/encoding/count_frequency.py:7:1: F401 'numpy as np' imported but unused
feature_engine/selection/drop_constant_features.py:3:1: F401 'numpy as np' imported but unused
```

Could you check the files you committed to see if the imports are still there and remove them?

Thanks a lot!

AndrewTanQB:
yes - i should have checked the build :) have fixed!

PR #256, opened 2021-05-03
[MRG] add check_estimator tests closes #178

### LINKED ISSUE #178: make transformers compatible with check_estimator from sklearn
With the idea of making feature-engine part of sklearn contrib, we need to make our transformers pass the test check_estimator.

There are a number of modifications that we need, 1 is being able to take other data types apart from dataframes, and the init params should be assigned directly, without modification. So we need to move some transformations that currently take place in init to the fit method.

### LINKED ISSUE #257: remove self.input_shape which will be replaced by n_features_in

### LINKED ISSUE #258: add automatic linting to circle ci

solegalli:
addresses and closes #178
closes #258 
closes #257 

This PR aims to add check_estimator tests to all feature-engine transformers. 

template for tests here:
https://github.com/scikit-learn-contrib/project-template/tree/master/skltemplate/tests

tags and more info:
https://scikit-learn.org/stable/developers/develop.html

In order to pass the tests, we needed to modify some parts of the code quite a bit, and change the names of some parameters or the behaviour of some parameters.

Particularly the params in init, and the attributes in fit

Sklearn introduced the n_features_in, which is also tested in check_estimators, thus we needed to change input_shape by this new attribute.

Also sklearn suggests not to over-write init parameters, so on some ocassions we needed to add attributes to over-write the defo parameters with the suitable variables.

This led to changes in the documentation docstrings. So I took the opportunity to update and expand the documentation.


solegalli:
encoders will not pass the check_estimators test, because they fail when no variable is of type object. The tests are done with numpy arrays, so they will fail most tests.

A work around would be to add a parameter `ignore_variable_type=False`. If False, which is the default, then the transformer works as usual. If true, instead of trasforming only object variables, it will encode as well numerical.

Thoughts?

solegalli:
check_estimator added succeasfully to all transformers.

The transformers that work with categorical variables fail, because the tests are done with numpy array numeric values. These are the encoders, the categorical imputer and the mean_target_selector

Need to decide if we add a work around to allow the transformers to operate on numeric variables, or accept that they will fail the tests

solegalli:
closes #258

solegalli:
closes #257

solegalli:
> 
> 
> TODO:
> 
>     * [x]  add ignore_format to categorical encoders
> 
>     * [x]  add check_estimator to categorical encoders
> 
>     * [x]  try the tag requires_positive_X for the log, reciprocal and etc transformers. Link below
> 
>     * [ ]  see if categorical imputer can be modified to pass the check_estimator
> 
>     * [ ]  target_mean_selector: modify to pass, needs some re-thinking
> 
>     * [ ]  remove input_shape
> 
>     * [ ]  update doc examples
> 
>     * [ ]  review the documentation once it is all done (new attributes added all along)
> 
>     * [ ]   update jupyter notebook examples
> 
>     * [ ]   add additional important tests like fit().fit()
> 
> 
> https://scikit-learn.org/stable/developers/develop.html

the tag requires_positive_X only ensures that an additional test is run to determine if a proper warning is raised, but does not create positive arrays, so the tests keep on failing.

PR #273, opened 2021-06-22
add inverse_transform method to transformer classes

### LINKED ISSUE #248: add inverse_transform method to variable transformer classes
Add to the log, power and reciprocal transformers (https://github.com/solegalli/feature_engine/tree/master/feature_engine/transformation) an inverse_transform method that performs the inverse operation. 

So for log it would be exp, for exp it would be log, for 1/x it would be 1/x

BoxCox and Yeojohnson, would continue without the inverse_transform. Not sure these operations can be reversed.

NicoGalli:
we have a new PR for this #283 

closes #248

### DIFF
--- a/feature_engine/transformation/log.py
+++ b/feature_engine/transformation/log.py
@@ -13,7 +13,7 @@ from feature_engine.variable_manipulation import _check_input_parameter_variable
 
 class LogTransformer(BaseNumericalTransformer):
     """
-    The LogTransformer() applies the natural logarithm or the base 10 logarithm to
+    The LogTransformer() applies the natural logarithm, the base 10 logarithm or exponetial function to
     numerical variables. The natural logarithm is logarithm in base e.
 
     The LogTransformer() only works with numerical non-negative values. If the variable
@@ -48,6 +48,8 @@ class LogTransformer(BaseNumericalTransformer):
         Transforms the variables using log transformation.
     fit_transform:
         Fit to data, then transform it.
+    inverse_transform:
+        Transforms the variables using exp transformation.
     """
 
     def __init__(
@@ -145,6 +147,45 @@ class LogTransformer(BaseNumericalTransformer):
             X.loc[:, self.variables_] = np.log(X.loc[:, self.variables_])
         elif self.base == "10":
             X.loc[:, self.variables_] = np.log10(X.loc[:, self.variables_])
+
+        return X
+    
+    def inverse_transform(self, X: pd.DataFrame) -> pd.DataFrame:
+        """
+        Transforms the variables using exp transformation.
+
+        Parameters
+        ----------
+        X: Pandas DataFrame of shape = [n_samples, n_features]
+            The data to be transformed.
+
+        Raises
+        ------
+        TypeError
+            If the input is not a Pandas DataFrame
+        ValueError
+            - If the variable(s) contain null values
+            - If the df has different number of features than the df used in fit()
+            - If some variables contains zero or negative values
+
+        Returns
+        -------
+        X: pandas dataframe
+            The dataframe with the transformed variables.
+        """
+
+        # check input dataframe and if class was fitted
+        X = super().transform(X)
+
+        # check contains zero or negative values
+        if (X[self.variables_] <= 0).any().any():
+            raise ValueError(
+                "Some variables contain zero or negative values, can't apply log"
+            )
+
+        # inverse_transform
+        X.loc[:, self.variables_] = np.exp(X.loc[:, self.variables_])
+
 
         return X
 
--- a/feature_engine/transformation/log.py
+++ b/feature_engine/transformation/log.py
@@ -13,8 +13,9 @@ from feature_engine.variable_manipulation import _check_input_parameter_variable
 
 class LogTransformer(BaseNumericalTransformer):
     """
-    The LogTransformer() applies the natural logarithm, the base 10 logarithm or exponetial function to
-    numerical variables. The natural logarithm is logarithm in base e.
+    The LogTransformer() applies the natural logarithm, the base 10 logarithm or
+    exponetial function to numerical variables.
+    The natural logarithm is logarithm in base e.
 
     The LogTransformer() only works with numerical non-negative values. If the variable
     contains a zero or a negative value, the transformer will return an error.
@@ -49,7 +50,7 @@ class LogTransformer(BaseNumericalTransformer):
     fit_transform:
         Fit to data, then transform it.
     inverse_transform:
-        Transforms the variables using exp transformation.
+        Transforms the variables using exponential transformation.
     """
 
     def __init__(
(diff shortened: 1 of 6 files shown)


feature_engine/transformation/log.py
```diff
@@ -13,7 +13,7 @@
 
 class LogTransformer(BaseNumericalTransformer):
     """
-    The LogTransformer() applies the natural logarithm or the base 10 logarithm to
+    The LogTransformer() applies the natural logarithm, the base 10 logarithm or exponetial function to
```
solegalli, line 16: Could we revert to the original line in the docstring? Normally we describe the transformers by their functionality in the transform method and not in the inverse_transform. It is a convention :)

feature_engine/transformation/log.py
```diff
@@ -48,6 +48,8 @@ class LogTransformer(BaseNumericalTransformer):
         Transforms the variables using log transformation.
     fit_transform:
         Fit to data, then transform it.
+    inverse_transform:
+        Transforms the variables using exp transformation.
```
solegalli, line 52: could we say exponential instead of exp?

feature_engine/transformation/log.py
```diff
@@ -147,6 +149,45 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
             X.loc[:, self.variables_] = np.log10(X.loc[:, self.variables_])
 
         return X
+    
+    def inverse_transform(self, X: pd.DataFrame) -> pd.DataFrame:
+        """
+        Transforms the variables using exp transformation.
```
solegalli, line 155: could we say exponential instead of exp? Could we also add "reverses the logarithm transformation"

feature_engine/transformation/log.py
```diff
@@ -147,6 +149,45 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+            The data to be transformed.
+
+        Raises
+        ------
+        TypeError
+            If the input is not a Pandas DataFrame
+        ValueError
+            - If the variable(s) contain null values
+            - If the df has different number of features than the df used in fit()
+            - If some variables contains zero or negative values
```
solegalli, line 169: contain instead of contain**s**

feature_engine/transformation/log.py
```diff
@@ -147,6 +149,45 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+        X = super().transform(X)
+
+        # check contains zero or negative values
+        if (X[self.variables_] <= 0).any().any():
+            raise ValueError(
+                "Some variables contain zero or negative values, can't apply log"
+            )
+
+        # inverse_transform
+        X.loc[:, self.variables_] = np.exp(X.loc[:, self.variables_])
```
solegalli, line 187: this function reverses log natural (base e). But what if user used log10? Can you add this?

feature_engine/transformation/power.py
```diff
@@ -124,3 +124,34 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
         X.loc[:, self.variables_] = np.power(X.loc[:, self.variables_], self.exp)
 
         return X
+
```
solegalli, line 127: can we add a description of inverse_transform in the init docstrings?

feature_engine/transformation/reciprocal.py
```diff
@@ -45,6 +45,8 @@ class ReciprocalTransformer(BaseNumericalTransformer):
         Apply the reciprocal 1 / x transformation.
     fit_transform:
         Fit to data, then transform it.
+    inverse_transform:
+        Apply the inverse reciprocal 1 / x transformation.
```
solegalli, line 49: Can we replace by "Reverses reciprocal transformation"

feature_engine/transformation/reciprocal.py
```diff
@@ -135,6 +137,33 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
         X.loc[:, self.variables_] = np.reciprocal(X.loc[:, self.variables_])
 
         return X
+    
+    def inverse_transform(self, X: pd.DataFrame) -> pd.DataFrame:
+        """
+        Apply the inverse reciprocal 1 / x transformation.
```
solegalli, line 143: Can we replace by "Reverses reciprocal transformation"

tests/test_transformation/test_log_transformer.py
```diff
@@ -85,3 +85,21 @@ def test_non_fitted_error(df_vartypes):
     with pytest.raises(NotFittedError):
         transformer = LogTransformer()
         transformer.transform(df_vartypes)
+        
+        
+def test_inverse_log_plus_user_passes_var_list(df_vartypes):
```
solegalli, line 90: Nice. We would also need a test for base=10

tests/test_transformation/test_power_transformer.py
```diff
@@ -49,3 +49,23 @@ def test_non_fitted_error(df_vartypes):
...
+    X["Age"] = X["Age"].round().astype('int64')
+    X["Marks"] = X["Marks"].round(1)
+
+    # test init params
+    assert transformer.variables is None
+    # test fit attr
+    assert transformer.variables_ == ["Age","Marks"]
+    assert transformer.n_features_in_ == 5
+    # test transform output
+    pd.testing.assert_frame_equal(X, df_vartypes)
```
solegalli, line 71: Nice. Could we add a test for another exp parameter? different from the default? just to make sure, eg exp=100

Just testing this line would do:  pd.testing.assert_frame_equal(X, df_vartypes)

solegalli (commented):
Nicely done, thank you very much!

To fix the codestyle you can do in your console:
isort feature_engine tests
black8 feature_engine tests

and then run:
flake8 feature_engine tests

to make sure there are no errors

Alternatively, if you have tox installed, you can run from your console:
tox -e lint

If your branch is sync with master, this will modify only the files you worked on. If it modifies more files, make sure to commit only the ones you worked on and not additional ones!!

feature_engine/transformation/log.py
NicoGalli: Would you please clarify on this request? New transform funcionality is explained same way it used to be: The Class used to described about the logarithim part and I just added the exponential. New method is explained on the new method, same way old one. I mean, if I revert still "natural logarithm and base 10" will be there, which as you mentioned dont follow the convention.

feature_engine/transformation/log.py
NicoGalli: Already says "contain". 
Are you asking to add the "S"? What about line 127 in transform method?

feature_engine/transformation/log.py
NicoGalli: I feel like we discussed about this one in Telegram. There is not a function to inverse log10.

Do you think something like this will work?

elif self.base == "10":
            X.loc[:, self.variables_] = np.array(10 ** X.loc[:, self.variables_])

feature_engine/transformation/power.py
NicoGalli: It is already there, would you please clarify on this?

def inverse_transform(self, X: pd.DataFrame) -> pd.DataFrame:
        """
        Apply the power inverse transformation to the variables.

        Parameters ....

feature_engine/transformation/log.py
NicoGalli: Test case is passed, so looks like it should ...

solegalli: yes, that works. Also np.exp(10, variable) should work for variable in self.variables_

solegalli:
closed in#283

Issue #275, opened 2021-06-22
[Feat]: Mutual Information in SmartCorrelatedSelection method


TremaMiguel:
**Is your feature request related to a problem? Please describe.**

Mutual Information help identifies different type of relationships between variables, not only linear relationships like Pearson or monotonic relations like spearman. 

**Describe the solution you'd like**

[SmartCorrelatedSelection](https://github.com/solegalli/feature_engine/blob/master/feature_engine/selection/smart_correlation_selection.py#L229) calls `pandas.corr()` under the hood, this method can also receives a custom function implementing calculating the mutual information between two variables, this [Stackoverflow answer](https://stackoverflow.com/questions/63969194/how-to-calculate-pairwise-mutual-information-for-entire-pandas-dataset/67987713#67987713) shows an example. 

The Mutual information can be calculated through the method `mutual_info_regression` from sklearn (its already a required dependency) or a custom calculation.

solegalli:
This is actually and easy issue. 

We need to remove the init checks in [this line](https://github.com/solegalli/feature_engine/blob/master/feature_engine/selection/smart_correlation_selection.py#L146) for the SmartCorrelatedSelection and in [this line](https://github.com/solegalli/feature_engine/blob/master/feature_engine/selection/drop_correlated_features.py#L89) for the drop correlated features transformer and that would allow the user to pass a custom function.

We should also add 1 test for each class to make sure that it works. And finally modify the docstrings and ideally add a Jupyter notebook. So to summarize:

- [ ] remove the init checks in [this line](https://github.com/solegalli/feature_engine/blob/master/feature_engine/selection/smart_correlation_selection.py#L146) for the SmartCorrelatedSelection
- [ ] remove the init checks in [this line](https://github.com/solegalli/feature_engine/blob/master/feature_engine/selection/drop_correlated_features.py#L89) for the drop correlated features transformer
- [ ] add 1 test for each class to make sure that it can take a custom function like the mutual_info
- [ ] modify the docstrings
- [ ] modify existing jupyter notebook in /examples or make a new one to showcase the new functionality

fancy giving it a go? @TremaMiguel

TremaMiguel:
@solegalli, would it be extending the method parameter to accept any valid callable ? 

```
method: string or callable, default='pearson'
        Can take 'pearson', 'spearman', 'kendall' or any callable method. It refers 
        to the correlation method to be used to identify the correlated features.
        - pearson : standard correlation coefficient
[...]

PR #279, opened 2021-06-25
allow callables to pandas.corr in correlation selection transformers

TremaMiguel:
This PR closes issue #275.

## Describe Your Changes

[SmartCorrelatedSelection](https://github.com/TremaMiguel/feature_engine/blob/feat_smartcorrelated_mi/feature_engine/selection/smart_correlation_selection.py#L46) and [DropCorrelatedFeatures](https://github.com/TremaMiguel/feature_engine/blob/feat_smartcorrelated_mi/feature_engine/selection/drop_correlated_features.py#L37) take as method any callable.
 
Callable Method is required to have to arguments of kind `POSITIONAL_OR_KEYWORD`, TypeError in `pandas.corr()` is raised if function receives a different number of positional arguments. [Link](https://github.com/TremaMiguel/feature_engine/blob/feat_smartcorrelated_mi/feature_engine/selection/drop_correlated_features.py#L98)

## How has this been tested

- New test `test_callable_method` added for [SmartCorrelatedSelection](https://github.com/TremaMiguel/feature_engine/blob/feat_smartcorrelated_mi/tests/test_selection/test_smart_correlation_selection.py#L257) and [DropCorrelatedFeatures](https://github.com/TremaMiguel/feature_engine/blob/feat_smartcorrelated_mi/tests/test_selection/test_drop_correlated_features.py#L126).  
- Raising `TypeError` is also tested [1](https://github.com/TremaMiguel/feature_engine/blob/feat_smartcorrelated_mi/tests/test_selection/test_smart_correlation_selection.py#L295) and [2](https://github.com/TremaMiguel/feature_engine/blob/feat_smartcorrelated_mi/tests/test_selection/test_drop_correlated_features.py#L154)
- Two new fixtures passing a [single argument method](https://github.com/TremaMiguel/feature_engine/blob/feat_smartcorrelated_mi/tests/test_selection/conftest.py#L71) and [dummy method](https://github.com/TremaMiguel/feature_engine/blob/feat_smartcorrelated_mi/tests/test_selection/conftest.py#L61) to `test_callable_method`

## Examples

[Smart-Correlation-Selection](https://github.com/TremaMiguel/feature_engine/blob/feat_smartcorrelated_mi/examples/selection/Smart-Correlation-Selection.ipynb) and [Drop-Correlated-Features](https://github.com/TremaMiguel/feature_engine/blob/feat_smartcorrelated_mi/examples/selection/Drop-Correlated-Features.ipynb) notebooks explaining how to pass callable methods.


feature_engine/selection/drop_correlated_features.py
```diff
@@ -86,11 +89,27 @@ def __init__(
         missing_values: str = "ignore",
     ):
 
-        if method not in ["pearson", "spearman", "kendall"]:
+        if method not in ["pearson", "spearman", "kendall"] and not callable(method):
             raise ValueError(
-                "correlation method takes only values 'pearson', 'spearman', 'kendall'"
+                "correlation method takes only values 'pearson', 'spearman', "
```
solegalli, line 94: I know that I am to blame for having so many checks in the init file.

But Sklearn actually encourages [not to have param checks in the init](https://scikit-learn.org/stable/developers/develop.html), to have them instead in the fit. So as we make modifications/emhancements to the package, I 'd like to entertain the idea of maximising compatibility with sklearn.

pandas.corr() will work or return an error depending on whether the user passes the correct argument here (pearson, spearman, kendall or callable: https://github.com/pandas-dev/pandas/blob/v1.2.5/pandas/core/frame.py#L8325-L8412

So how about removing all param checks from init, and just leaving the work to pandas.corr() as per
https://github.com/pandas-dev/pandas/blob/v1.2.5/pandas/core/frame.py#L8325-L8412

This will make the code of this transformer simpler, and a tiny bit faster, by outsourcing the checks to pandas.

The tests that you added, should still pass? 

Thoughts?

feature_engine/selection/drop_correlated_features.py
```diff
@@ -86,11 +89,27 @@ def __init__(
...
+                + "'kendall' or callable."
             )
 
+        # check callable method takes two input arguments
+        if callable(method):
```
solegalli, line 99: as per the previous comment, we would also not need this bit of code

feature_engine/selection/smart_correlation_selection.py
```diff
@@ -143,11 +146,27 @@ def __init__(
...
             raise ValueError(
-                "correlation method takes only values 'pearson', 'spearman', 'kendall'"
+                "correlation method takes only values 'pearson', 'spearman', "
+                + "'kendall' or callable."
             )
 
+        # check callable method takes two input arguments
+        if callable(method):
+            sig = signature(method)
+            positional_or_keyword_args = sum(
```
ChristopherGS, line 158: Perhaps: 

```
positional_or_keyword_args = [param for param in sig.paramters.values() if param.kind == param.POSITIONAL_OR_KEYWORD]

if len(positional_or_keyword_args) != 2:
    ...
```

solegalli (commented):
Hi @TremaMiguel thank you so much for such an amazing contribution.

Would you mind taking a look at my comments on checking the arguments of the init parameter? that would be the main comment.

The rest is mostly to brain storm a tiny bit.

Thanks a lot!

feature_engine/selection/smart_correlation_selection.py
ChristopherGS: Also (sorry, I'm not that familiar with the context), do we want to catch callables with keyword args?

e.g.
```
def foo(a, *, b:int):
    pass
```
we wouldn't catch `b` with the above check, we'd need to also check for kind `KEYWORD_ONLY`

tests/test_selection/conftest.py
```diff
@@ -55,3 +55,27 @@ def df_test_num_cat():
     X = df[["var_A", "var_B", "var_C", "var_D"]]
     y = df["target"]
     return X, y
+
+
+@pytest.fixture(scope="module", name="dummy_method")
+def wrap_dummy_method():
+    def wrapper():
+        import random
```
solegalli, line 63: may I confess that I didn't know this pytest wrapping functionality? would anyone fancy enlightening me? :p

Thank you!

ChristopherGS: The fixture is returning a function, which is then returning another function.

Git is complaining about me checking out this branch (perhaps because it is a PR off a feature branch?), but my initial thoughts were:

- Do we need the triple nesting of functions? 
- Do we need to immediately invoke the returned function?

feature_engine/selection/smart_correlation_selection.py
TremaMiguel: @ChristopherGS  you're right we wouldn't be catching also 

```python
def foo2(*, a:int, b:int):
    pass
```

or catching more arguments than needed.

```
def foo3(x, y, *, z=1, w='something'):
    pass
```

the idea behind this check is to avoid a `TypeError` when calling the method as it only expects two input arguments. For example
```
def corrf(x):
    pass

corrf(1, 2)
>>> TypeError: foo4() takes 1 positional argument but 2 were given
```
 
I think it could be better if the error is directly raise by `pandas.corr()`. What are your thoughts ?

solegalli: > I think it could be better if the error is directly raise by `pandas.corr()`. What are your thoughts ?

agree

feature_engine/selection/drop_correlated_features.py
```diff
@@ -33,13 +34,15 @@ class DropCorrelatedFeatures(BaseSelector):
...
-        to be used to identify the correlated features.
+    method: string or callable, default='pearson'
+        Can take 'pearson', 'spearman', 'kendall' or callable. It refers to the
+        correlation method to be used to identify the correlated features.
 
         - pearson : standard correlation coefficient
         - kendall : Kendall Tau correlation coefficient
         - spearman : Spearman rank correlation
+        - callable : callable with input two 1d ndarrays
+            and returning a float.
```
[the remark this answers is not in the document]
TremaMiguel: we could also display the same message as `pandas.corr` 

Captura de pantalla (103)

tests/test_selection/test_drop_correlated_features.py
```diff
@@ -134,3 +153,8 @@ def test_non_fitted_error(df_correlated_single):
     with pytest.raises(NotFittedError):
         transformer = DropCorrelatedFeatures()
         transformer.transform(df_correlated_single)
+
```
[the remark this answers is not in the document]
TremaMiguel: yes, but to make it work with other kind of correlation measures we should adapt the method to the range of the measure (I think we can develop another PR than extends the threshold and selection criteria to a custom method). For example the mutual information can take values greater than one and normally one would discard features that have low value.

feature_engine/selection/drop_correlated_features.py
TremaMiguel: I agree on that, as this method call `pandas.corr` under the hood. However, one interesting thing about this approach is that user should go one step back to fix the method parameter.

```python
# Now: error raises here
transformer = DropCorrelatedFeatures(
    variables=None, method=dummy_method, threshold=0.6
)
# With change: error raises here, user goes one step back to fix method parameter
Xt = transformer.fit_transform(X)
```

TremaMiguel: Regarding tests, no major changes,  just need to remove the TypeError raising test.

tests/test_selection/conftest.py
TremaMiguel: well, here it injects a function to a test through a wrapper. As the output of fixtures can be passed to testing functions, is like a hack to pass a function to a test. Perhaps there's a better way of doing this.

TremaMiguel: @ChristopherGS triple nesting does not look nice, is there a way we can a pass a method to a test ? I think inside test we could import it from some testing_utils.

ChristopherGS: Would something like this do the trick?

```
@pytest.fixture()
def argument_printer():
    def _foo(*args, **kwargs):
        return (args, kwargs)

    return _foo


def test_example(argument_printer):
    first_case = argument_printer('a', 'b')
    assert first_case == (('a', 'b'), {})

    second_case = argument_printer('a', b=2, c='test')
    assert second_case == (('a',), {'b': 2, 'c': 'test'})

```

TremaMiguel: @ChristopherGS yes , thanks for the clarity.

feature_engine/selection/drop_correlated_features.py
solegalli: I know. I also find that why would you go one step back if you can avoid it. But this is how sklearn is designed. The initialisation of their transformers almost never fails, only when using fit(), the transformers check that the init params are allowed.

I think it is a design strategy, to keep init as clean as possible. And they also say to ensure compatibility with grid search and pipeline. I don't think those will fail with our checks. But still, I think for now, I would like to try and follow their functionality as closely as possible.

feature_engine/selection/smart_correlation_selection.py
solegalli: OK, so just to summarize, we will remove this functionality from the code. If anything else, for my clarity lol

feature_engine/selection/drop_correlated_features.py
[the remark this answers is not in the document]
solegalli: Not too sure how is this different from what is in the docstrings already? I thought it was quite similar to pandas.corr().

tests/test_selection/test_drop_correlated_features.py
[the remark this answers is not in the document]
solegalli: Thinking it further, the test for the callable with 2 arguments is doing enough. I got it later when I checked the functionality for pandas.corr(). So the tests we have already should do.

feature_engine/selection/drop_correlated_features.py
solegalli: Also not sure TypeError needs removing. pandas.corr() should not raise it?

solegalli:
@TremaMiguel thanks for the engagement and the comments! let's know when OK to review again :)

feature_engine/selection/smart_correlation_selection.py
TremaMiguel: ok, so we remove this from `DropCorrelatedFeatures` and `SmartCorrelatedSelection` and the error is raised by pandas when calling `fit` or `fit_transform`.

feature_engine/selection/drop_correlated_features.py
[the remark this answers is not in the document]
TremaMiguel: ok, so the message will be similar as `pandas.corr()`, except for `callable` we can only display 

`callable: callable with input two 1d ndarrays and returning a float.`

feature_engine/selection/smart_correlation_selection.py
```diff
@@ -43,13 +44,15 @@ class SmartCorrelatedSelection(BaseSelector):
...
-        to be used to identify the correlated features.
+    method: string or callable, default='pearson'
+        Can take 'pearson', 'spearman', 'kendall' or callable. It refers to the
+        correlation method to be used to identify the correlated features.
 
         - pearson : standard correlation coefficient
         - kendall : Kendall Tau correlation coefficient
         - spearman : Spearman rank correlation
+        - callable : callable with input two 1d ndarrays
+            and returning a float.
```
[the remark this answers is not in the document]
TremaMiguel: got it

feature_engine/selection/drop_correlated_features.py
TremaMiguel: sure, got it. Yes pandas.corr() will raise a `ValueError`, we were checking a `TypeError` for checking the arguments for the callable method, I've just removed that test and that check.

TremaMiguel:
@solegalli all ready for a second check please

TremaMiguel:
Hi @solegalli any update with the check ?

tests/test_selection/conftest.py
```diff
@@ -55,3 +55,13 @@ def df_test_num_cat():
     X = df[["var_A", "var_B", "var_C", "var_D"]]
     y = df["target"]
     return X, y
+
+
+@pytest.fixture(scope="module", name="dummy_method")
```
solegalli, line 60: @TremaMiguel this PR looks good to merge to me. The only request, could we change the name of this function to something more representative of what the function does? 

In case we want to use this function for other methods, we get an intuition through the name of what the function does. :)

tests/test_selection/conftest.py
```diff
@@ -55,3 +55,13 @@ def df_test_num_cat():
...
+def wrap_dummy_method():
+    import random
+
+    def _dummy_method(x, y):
+        return random.uniform(0.5, 1)
+
```
solegalli, line 66: @ChristopherGS @TremaMiguel 

Sorry, I am a bit ignorant on this front. I guess we need to wrap the method so that we do not need to pass X, y when calling it from the tests, is this correct? otherwise, why do we need to wrap the method?

would random be imported at every single evaluation of _dummy_method() ? would it not be better to import it at the top?

And as per previous message, could we give this block of code a more representative name? Something like "return_uniform_between_05_and_1" or something on that front.

solegalli: > 
> 
> Hi @solegalli any update with the check ?

Hey sorry for the belated response. I am on hols, back in the office on Monday. 

I think this PR is almost ready to merge. I made a PR to your repo @tremaMiguel with some minor edits to docstrings, and I have a question regarding the dummy_method, that would be great if you or @ChristopherGS could enlighten me with and/or make the changes if necessary.

Thanks a lot!

solegalli:
Also, could you please remove feature_engine version print out at the top of e`xamples/selection/Drop-Correlated-Features.ipynb`  ?

thank you :)

tests/test_selection/conftest.py
TremaMiguel: yes importing random at the top would be better in this case as several tests can call this method, thanks for the catchup.

TremaMiguel: Basically, we're wrapping it to make it accessible for the test function or inject it to the test function. 

The output of `wrap_dummy_method` is a callable `_dummy_method` (we do not pass any arguments to it yet) that it is passed as argument to `SmartCorrelatedSelection` in `test_callable_method`. 

Inside `SmartCorrelatedSelection`, the function `_dummy_method` is called with the provided `X, y` arguments from the user, basically inside `pandas.corr` is performed this calculation.

TremaMiguel: @solegalli added changes requested, let me know if anything else is needed? Do you know if it is possible to change the target branch to `master` ?

ChristopherGS (approved):

Issue #274, opened 2021-06-22
[Feat]: Log1p transformer 


TremaMiguel:
**Is your feature request related to a problem? Please describe.**

When you want to apply a logarithmic transformation and data contains a zero value, calculating `log(1+x)` is more convenient. 

**Describe the solution you'd like**

This could be another argument in [LogTransformer](https://github.com/solegalli/feature_engine/blob/master/feature_engine/transformation/log.py#L144) that checks for zeros and apply `log(1 + x)`

```
from feature_engine import transformation as vt

tf = vt.LogTransformer(variables = ['LotArea', 'GrLivArea'], check_zeros=True)
```
under the hood basically applies 
```
if check_zeros:
    log_1p = True if (X[self.variables_] == 0).any().any() else False
```

or it can be a separate transformer 

```
from feature_engine import transformation as vt

tf = vt.Log1PTransformer(variables = ['LotArea', 'GrLivArea'], check_zeros=True)
```

solegalli:
What would this transformer do if `check_zeroes` is `True`? add 1 to the variable and then apply the log?

I am aware that this is done quite frequently. I just think it is dangerous to add 1 to a variable without really inspecting the variable nature that I am kind of reluctant to encourage its use by implementing it in a transformer.

Also why +1 and not +10?

TremaMiguel:
I think it would be better for the user to decide which constant `C` to add and handle it as a separate transformer. For example, 

```
class LogCpTransformer(BaseNumericalTransformer):
    """
    The LogCpTransformer() applies the transformation log( x + C) where C is a positive constant to 
    the input variable. Adding a C constant helps apply log transformation to a variable having zero
    or negative values. 
    
    The LogTransformer() only works with numerical non-negative values. If the variable
    contains a zero or a negative value after adding a constant C, the transformer will return an error.
```

GLevV:
In `LogCpTransformer` we can just set C=0 by default and also add option 'auto' which will do something like that `C = abs(min(x)) + 1`. This will allow us to use log transformation on negative values.
Also this one is bad 
>If the variable contains a zero or a negative value after adding a constant C, the transformer will return an error.
because log(0+1) = 0, so you will always return error. The reason behind log1p is to avoid infs and nans that will arise in log(0) case, not in avoiding zeros.

solegalli:
> If the variable contains a zero or a negative value after adding a constant C, the transformer will return an error.

that line checks the variable **before** the transformation.

[...]

PR #281, opened 2021-06-27
feat: log c plus transformer

TremaMiguel:
This PR closes issue #274 

## Describe Your Changes

New transformer [LogCpTransformer](https://github.com/TremaMiguel/feature_engine/blob/feat_logc_transformer/feature_engine/transformation/log.py#L169)
 
As suggested by @GLevV, the `C` parameter can take an integer value or an `"auto"` option, in the second case the absolute value of the min is stored as `C`

```python
self.C = X[self.variables_].min(axis=0).abs() + 1
```

## How has this been tested

Same tests as Log Transformer added. [Link](https://github.com/TremaMiguel/feature_engine/blob/feat_logc_transformer/tests/test_transformation/test_logcp_transformer.py#L36)

## Examples

[LogCpTransformer](https://github.com/TremaMiguel/feature_engine/blob/feat_logc_transformer/docs/transformation/LogCpTransformer.rst) documentation includes example.

log_cp_example


tests/test_transformation/test_logcp_transformer.py
```diff
@@ -0,0 +1,112 @@
...
+
+@pytest.mark.parametrize("log_base, exp_age, exp_marks", _test_automatic_find_variables)
+def test_logcp_base_plus_automatically_find_variables(
+    log_base, exp_age, exp_marks, df_vartypes
+):
+    transformer = LogCpTransformer(base=log_base, variables=None)
+    X = transformer.fit_transform(df_vartypes)
+
+    # expected output
+    transf_df = df_vartypes.copy()
```
solegalli, line 44: The test file is a masterpiece. Thank you so much. 

Could we add tests for init param C and fit attribute C_?

either as part of tests 1 or 2 or separately if you find it better. Thank you!

docs/transformation/LogTransformer.rst
```diff
@@ -21,7 +21,7 @@ Example
 	from feature_engine import transformation as vt
 
 	# Load dataset
-	data = data = pd.read_csv('houseprice.csv')
+	data = pd.read_csv('houseprice.csv')
```
solegalli, line 24: thank you!

docs/transformation/LogCpTransformer.rst
```diff
@@ -0,0 +1,51 @@
...
+.. autoclass:: feature_engine.transformation.LogCpTransformer
+    :members:
+
+
+Example
+-------
+
+.. code:: python
+
+	import pandas as pd
```
solegalli, line 16: indentation

docs/transformation/LogCpTransformer.rst
```diff
@@ -0,0 +1,51 @@
...
+
+    # Load dataset
+    X, y = load_boston(return_X_y=True)
+    X = pd.DataFrame(X)
+
+    # Separate into train and test sets
+    X_train, X_test, y_train, y_test =  train_test_split(X, y, test_size=0.3, random_state=0)
+
+    # set up the variable transformer
+    tf = vt.LogCpTransformer(variables = [7, 12])
```
solegalli, line 31: not necessary but helpful if we do show param C in the init

docs/transformation/LogCpTransformer.rst
```diff
@@ -0,0 +1,51 @@
...
+
+    # transform the data
+    train_t= tf.transform(X_train)
+    test_t= tf.transform(X_test)
+
+	# un-transformed variable
+    X_train[12].hist()
+
+.. image:: ../images/logcpraw.png
+
```
solegalli, line 44: could we also display the output of C_? particularly useful when C is auto.

solegalli (commented):
Hi @TremaMiguel 

This PR is really good. Thank you so much! 

Aside of the comments below, could you consider the following:

- [ ] could we add the new transformer to the list to be checked by sklearn's `check_estimator ` [here](https://github.com/solegalli/feature_engine/blob/master/tests/test_transformation/test_check_estimator_transformers.py) and make sure tests pass? They will after we change the C to C_
- [ ] In PR #273 we are adding `inverse_transform` to all transformers from this module. Would you mind adding it to this one as part of this PR you are currently working with?
- [ ] if you have time and don't mind, would you be able to add a jupyter notebook with an example to the examples folder? not necessary to merge this PR though.

feature_engine/transformation/log.py
```diff
@@ -164,3 +164,161 @@ def _more_tags(self):
...
+    variables_:
+        The group of variables that will be transformed.
+
+    n_features_in_:
+        The number of features in the train set used in fit.
+
+    Methods
+    -------
+    fit:
+        This transformer does not learn parameters.
```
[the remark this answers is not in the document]
TremaMiguel: Sure, will be something like 
```
This transformer learns the constant C to add to log transformation.
```

feature_engine/transformation/log.py
```diff
@@ -164,3 +164,161 @@ def _more_tags(self):
...
+        if not isinstance(C, int) and not C == "auto":
+            raise ValueError("C can take only 'auto' or int")
+
+        self.variables = _check_input_parameter_variables(variables)
+        self.base = base
+        self.C = C
+
+    def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
+        """
+        This transformer does not learn parameters.
```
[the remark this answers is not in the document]
TremaMiguel: got it

feature_engine/transformation/log.py
```diff
@@ -164,3 +164,161 @@ def _more_tags(self):
...
+
+        Select the numerical variables and determines whether the logarithm
+        can be applied on the selected variables (it checks if the variables
+        are all positive).
```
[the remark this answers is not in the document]
TremaMiguel: got it

feature_engine/transformation/log.py
```diff
@@ -164,3 +164,161 @@ def _more_tags(self):
...
+        -------
+        self
+        """
+
+        # check input dataframe
+        X = super().fit(X)
+
+        # calculate C to add to each variable
+        if self.C == "auto":
+            self.C = X[self.variables_].min(axis=0).abs() + 1
```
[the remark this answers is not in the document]
TremaMiguel: Sure, I'll change that. Besides, I've frequently seen the use of trailing underscores, but I'm not all familiar with it, could you share me a reference to better understand it ?

feature_engine/transformation/log.py
```diff
@@ -164,3 +164,161 @@ def _more_tags(self):
...
+        The constant C to apply the transfomation log(x + C).
+
+        - If int, then log(x + C)
+        - If "auto", then C = abs(min(x)) + 1
+
+    Attributes
+    ----------
+    variables_:
+        The group of variables that will be transformed.
+
```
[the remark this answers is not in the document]
TremaMiguel: got it

tests/test_transformation/test_logcp_transformer.py
TremaMiguel: got it

docs/transformation/LogCpTransformer.rst
TremaMiguel: got it

docs/transformation/LogCpTransformer.rst
TremaMiguel: got it

docs/transformation/LogCpTransformer.rst
TremaMiguel: not sure why locally this indentation doesn't appear 

```python
.. code:: python
    
    import pandas as pd
    import matplotlib.pyplot as plt
```

TremaMiguel: fixed

TremaMiguel:
@solegalli please take a second look, I have addressed all your comments.

feature_engine/transformation/log.py
```diff
@@ -164,3 +164,206 @@ def _more_tags(self):
...
+    n_features_in_:
+        The number of features in the train set used in fit.
+
+    C_:
```
solegalli, line 206: could we put this attribute above n_features_in_ please?

feature_engine/transformation/log.py
```diff
@@ -164,3 +164,206 @@ def _more_tags(self):
...
+        if not isinstance(C, int) and not C == "auto":
+            raise ValueError("C can take only 'auto' or int")
+
+        self.variables = _check_input_parameter_variables(variables)
+        self.base = base
+        self.C = C
+
+    def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
+        """
+        This transformer learns the constant C to add to log transformation.
```
solegalli, line 238: **If auto,** this transformer learns the constant C to add to log transformation.

Could we add the above? otherwise it might be confusing?

feature_engine/transformation/log.py
```diff
@@ -164,3 +164,206 @@ def _more_tags(self):
...
+
+        Raises
+        ------
+        TypeError
+            - If the input is not a Pandas DataFrame
+            - If any of the user provided variables are not numerical
+        ValueError
+            - If there are no numerical variables in the df or the df is empty
+            - If the variable(s) contain null values
+            - If some variables contain zero or negative values after suming constant C
```
solegalli, line 261: could we replace **suming** by **adding** please?

feature_engine/transformation/log.py
```diff
@@ -164,3 +164,206 @@ def _more_tags(self):
...
+        # calculate C to add to each variable
+        if self.C == "auto":
+            self.C_ = X[self.variables_].min(axis=0).abs() + 1
+        else:
+            self.C_ = self.C
+
+        # check contains zero or negative values
+        if (X[self.variables_] + self.C_ <= 0).any().any():
+            raise ValueError(
+                "Some variables contain zero or negative values after suming"
```
solegalli, line 280: could we replace **suming** by **adding** please?

feature_engine/transformation/log.py
```diff
@@ -164,3 +164,206 @@ def _more_tags(self):
...
+            The data to be transformed.
+
+        Raises
+        ------
+        TypeError
+            If the input is not a Pandas DataFrame
+        ValueError
+            - If the variable(s) contain null values
+            - If the df has different number of features than the df used in fit()
+            - If some variables contains zero or negative values after suming constant C
```
solegalli, line 304: could we replace **suming** by **adding** please?

feature_engine/transformation/log.py
```diff
@@ -164,3 +164,206 @@ def _more_tags(self):
...
+                + "constant C, can't apply log"
+            )
+
+        self.n_features_in_ = X.shape[1]
+
+        return self
+
+    def transform(self, X: pd.DataFrame) -> pd.DataFrame:
+        """
+        Transforms the variables using log transformation.
```
solegalli, line 290: Transforms the variables using **the** log transformation. Or more directly "Transforms the variables with the logarithm" (probably better).

Probably also needs changing in the old logtransformer above.

feature_engine/transformation/log.py
```diff
@@ -164,3 +164,206 @@ def _more_tags(self):
...
+            The dataframe with the transformed variables.
+        """
+
+        # check input dataframe and if class was fitted
+        X = super().transform(X)
+
+        # check contains zero or negative values
+        if (X[self.variables_] + self.C_ <= 0).any().any():
+            raise ValueError(
+                "Some variables contain zero or negative values after suming"
```
solegalli, line 318: "adding"

feature_engine/transformation/log.py
```diff
@@ -164,3 +164,206 @@ def _more_tags(self):
...
+        if self.base == "e":
+            X.loc[:, self.variables_] = np.log(X.loc[:, self.variables_] + self.C_)
+        elif self.base == "10":
+            X.loc[:, self.variables_] = np.log10(X.loc[:, self.variables_] + self.C_)
+
+        return X
+
+    def inverse_transform(self, X: pd.DataFrame) -> pd.DataFrame:
+        """
+        Transforms the variables using exponential transformation.
```
solegalli, line 332: We try to use the sklearn docstring: "Convert the data back to the original representation."

See for example here:
https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.OrdinalEncoder.html#sklearn.preprocessing.OrdinalEncoder.inverse_transform

feature_engine/transformation/log.py
```diff
@@ -164,3 +164,206 @@ def _more_tags(self):
...
+    C_:
+        The constant C to transform data.
+
+    Methods
+    -------
+    fit:
+        This transformer learns the constant C to add to log transformation.
+    transform:
+        Transforms the variables using log transformation.
+    fit_transform:
```
solegalli, line 215: can we please add:
```
inverse_transform: 
      Convert the data back to the original representation.
```

https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.OrdinalEncoder.html#sklearn.preprocessing.OrdinalEncoder.inverse_transform

feature_engine/transformation/log.py
```diff
@@ -164,3 +164,206 @@ def _more_tags(self):
...
+            The data to be detransformed.
+
+        Raises
+        ------
+        TypeError
+            If the input is not a Pandas DataFrame
+        ValueError
+            - If the variable(s) contain null values
+            - If X has different number of features than X used in fit()
+            - If some variables contains zero or negative values
```
solegalli, line 346: I don't think we should raise this error, because log of a value >0 and <1 will be negative, and can return the inverse. Am I correct?

feature_engine/transformation/log.py
```diff
@@ -164,3 +164,206 @@ def _more_tags(self):
...
+        -------
+        X: Pandas dataframe
+            The dataframe with the transformed variables.
+        """
+
+        # check input dataframe and if class was fitted
+        X = super().transform(X)
+
+        # check contains zero or negative values
+        if (X[self.variables_] <= 0).any().any():
```
solegalli, line 358: this check is suitable for transform, but not for inverse_transform, values can be negative.

This was probably in the PR i linked before, but we fixed it there already and we made a new PR with the updates. Sorry

tests/test_transformation/test_logcp_transformer.py
```diff
@@ -0,0 +1,155 @@
+import pandas as pd
+import pytest
+from sklearn.exceptions import NotFittedError
+
+from feature_engine.transformation import LogCpTransformer
+
+_age_inverse_transform = [20.0, 21.0, 19.0, 18.0]
```
solegalli, line 7: If we add one variable with values >0 and < than 1 we would catch the bug on having the test for <0 in the inverse_transform

tests/test_transformation/test_logcp_transformer.py
```diff
@@ -0,0 +1,155 @@
...
+    # test error when data contains negative values
+    df_neg = df_vartypes.copy()
+    df_neg.loc[1, "Age"] = -df_vartypes["Age"].min() - 2
+
+    # when variable contains negative value, transform
+    with pytest.raises(ValueError):
+        transformer = LogCpTransformer()
+        transformer.fit(df_vartypes)
+        transformer.transform(df_neg)
+
```
solegalli, line 150: do we need a test for when user specifies a c and the variable + the c is still negative during fit()? We have the error catch already, it is just a matter of adding a test.

I anticipate that many users will just make c=1 without even looking at the variable values, so transform seems a bit late to pick up this error.

solegalli (commented):
@TremaMiguel 

Thank you so much for the quick turnaround and for such good and complete code. I love it :)

Just very tiny modifications requests to the docstrings and something that I believe could be a bug in the inverse_transform. And 1 or 2 small test addition requests :)

feature_engine/transformation/log.py
```diff
@@ -164,3 +164,206 @@ def _more_tags(self):
...
+        -------
+        self
+        """
+
+        # check input dataframe
+        X = super().fit(X)
+
+        # calculate C to add to each variable
+        if self.C == "auto":
+            self.C_ = X[self.variables_].min(axis=0).abs() + 1
```
solegalli, line 273: This is perhaps a painful request because it might slow things down, but I'll make it anyhow and we can discuss about it.

All Feature-engine transformers store learn parameters when they differ per variable as a dictionary, so that the user can easily see which value corresponds to each variable. And this is a key difference with sklearn arrays, where you can't really understand without some data manipulation what value corresponds to what.

I find this very useful. So I would like C_ to be a dictionary of variable: minimum +1 value pairs. Could we do that?

The display in rst would also need some changing after this modification

feature_engine/transformation/log.py
TremaMiguel: got it, additionally as the variable is already in log scale, the exponential function accepts negative and positive values, so there's no need for this error, thanks for catching it.

tests/test_transformation/test_logcp_transformer.py
TremaMiguel: ok, so if removing the check in inverse transform, no need for testing this, right?

tests/test_transformation/test_logcp_transformer.py
TremaMiguel: yes, good catch, when `C` is set manually then it is possible to raise this error. I'll add a test for that

feature_engine/transformation/log.py
TremaMiguel: will add this also as input option for `C` param 

```python
C: "auto", int or dict, default="auto"
    The constant C to apply the transfomation log(x + C).

      - If int, then log(x + C)
      - If "auto", then C = abs(min(x)) + 1
      - If dict, dictionary mapping the constant C to apply
        to each variable.
```

TremaMiguel:
@solegalli please take another look at changes

tests/test_transformation/test_logcp_transformer.py
solegalli: it doesn't hurt to test in my opinion :)

feature_engine/transformation/log.py
solegalli: makes total sense, great idea

feature_engine/transformation/log.py
TremaMiguel: I've not seen this one, gonna make the change

tests/test_transformation/test_logcp_transformer.py
TremaMiguel: so if we add an "Score" variable here

```python
import pandas as pd
import numpy as np

data = {
        "Name": ["tom", "nick", "krish", "jack"],
        "City": ["London", "Manchester", "Liverpool", "Bristol"],
        "Age": [20, 21, 19, 18],
        "Marks": [0.9, 0.8, 0.7, 0.6],
        "Score": [0.15, 0.007, 0.32, 0.21],
        "dob": pd.date_range("2020-02-24", periods=4, freq="T"),
    }

df_vartypes = pd.DataFrame(data)
df_vartypes
```

what should we test for ?, considering we remove the check in `inverse_transform` 

```python 
# check contains zero or negative values
if (X[self.variables_] + self.C_ <= 0).any().any():
    raise ValueError(
        "Some variables contain zero or negative values after adding"
        + "constant C, can't apply log"
    )
```

solegalli: You could use the "Marks" variable directly. That one should have catched the error raising the variables should not be negative in the inverse transform.

solegalli: Oh, I understand the question now. Sorry. Applying the inverse_transform on a variable with values >-1, ensures the transformer works not only for variables >1 but also >0 and <1. And protects us automatically for introducing an error in the future. The error does not exist today, but we may accidentally introduce it in the future if we forget why we removed it, or if we modify some other masterclass from which this transformer inherits.

feature_engine/transformation/log.py
```diff
@@ -164,3 +164,205 @@ def _more_tags(self):
...
+        if self.base == "e":
+            X.loc[:, self.variables_] = np.log(X.loc[:, self.variables_] + self.C_)
+        elif self.base == "10":
+            X.loc[:, self.variables_] = np.log10(X.loc[:, self.variables_] + self.C_)
+
+        return X
+
+    def inverse_transform(self, X: pd.DataFrame) -> pd.DataFrame:
+        """
+        Transforms the variables using the exponential transformation.
```
solegalli, line 338: The method should read "Convert the data back to the original representation." as per sklearn convention for inverse_transform.

tests/test_transformation/test_logcp_transformer.py
TremaMiguel: yes, thanks for clarifying, this is an easy addition, might be similar to `test_inverse_transform`

feature_engine/transformation/log.py
TremaMiguel: got it

TremaMiguel:
@solegalli added tests for float and int for constant `C`, think it is ready to be merged. Do you think we should change the name of the transformer ?

solegalli:
> 
> 
> @solegalli added tests for float and int for constant `C`, think it is ready to be merged. Do you think we should change the name of the transformer ?

@TremaMiguel  thank you so much for the final changes! Yes, this is looking ready to merge.

Not sure about the name change, at the beginning, when I saw LogCpTransformer I did not quite understand what the p stands for. Only when I read the docs, I saw that the transformation is plus (p) a Constant (C). And then it made sense.

Ideally, we want the user to understand what the transformer does, directly from the transformer name. Not sure it this is possible in 100% of cases.

I think, I am happy with the name for the time being. Unless you have a suggestion?

PS: sorry for the delayed response, I am actually on holidays for the week :p

TremaMiguel:
@solegalli ok ok, added your PR, let me know if anything else is needed.

PR #298, opened 2021-08-04
Added `get_feature_names` API to encoders

janrito:
+ added checks for feature name output on tests

This allows the use of  things like [eli5](https://eli5.readthedocs.io/en/latest/) alongside feature engine. 

Important links:
https://scikit-learn.org/stable/glossary.html#term-get_feature_names

### DIFF
--- a/feature_engine/base_transformers.py
+++ b/feature_engine/base_transformers.py
@@ -2,7 +2,7 @@
 classes. Provides the base functionality within the fit() and transform() methods
 shared by most transformers, like checking that input is a df, the size, NA, etc.
 """
-from typing import Dict, List, Optional
+from typing import Dict, List, Optional, Union
 
 import pandas as pd
 from sklearn.base import BaseEstimator, TransformerMixin
@@ -147,7 +147,7 @@ class BaseNumericalTransformer(BaseEstimator, TransformerMixin):
 
     def get_feature_names(
         self,
-        input_features: List[str] = None,
+        input_features: Optional[List[Union[str, int]]] = None,
     ) -> List:
         """
         Return feature names for output features.
--- a/feature_engine/base_transformers.py
+++ b/feature_engine/base_transformers.py
@@ -2,7 +2,7 @@
 classes. Provides the base functionality within the fit() and transform() methods
 shared by most transformers, like checking that input is a df, the size, NA, etc.
 """
-from typing import Dict, Optional
+from typing import Dict, List, Optional
 
 import pandas as pd
 from sklearn.base import BaseEstimator, TransformerMixin
@@ -144,3 +144,26 @@ class BaseNumericalTransformer(BaseEstimator, TransformerMixin):
     # for the check_estimator tests
     def _more_tags(self):
         return _return_tags()
+
+    def get_feature_names(
+        self,
+        input_features: List[str] = None,
+    ) -> List:
+        """
+        Return feature names for output features.
+
+        Parameters
+        ----------
+        input_features : list of str of shape (n_features,)
+            String names for input features if available. By default,
+            self.variables_ is used
+
+        -------
+        feature_names : list of str of shape (n_output_features,) of feature names
+        """
+        check_is_fitted(self)
+
+        if input_features is None:
+            input_features = self.variables_
+
+        return [feature for feature in input_features if feature in self.variables_]
--- a/feature_engine/base_transformers.py
+++ b/feature_engine/base_transformers.py
@@ -97,6 +97,9 @@ class BaseNumericalTransformer(BaseEstimator, TransformerMixin):
         # find or check for numerical variables
         self.variables_ = _find_or_check_numerical_variables(X, self.variables)
 
+        # save input features
+        self.input_features_: List[Union[str, int]] = X.columns.tolist()
+
         # check if dataset contains na or inf
         _check_contains_na(X, self.variables_)
         _check_contains_inf(X, self.variables_)
@@ -145,25 +148,13 @@ class BaseNumericalTransformer(BaseEstimator, TransformerMixin):
     def _more_tags(self):
         return _return_tags()
 
-    def get_feature_names(
-        self,
-        input_features: Optional[List[Union[str, int]]] = None,
-    ) -> List:
+    def get_feature_names(self) -> List:
         """
         Return feature names for output features.
-
-        Parameters
-        ----------
-        input_features : list of str of shape (n_features,)
-            String names for input features if available. By default,
-            self.variables_ is used
 
         -------
         feature_names : list of str of shape (n_output_features,) of feature names
         """
         check_is_fitted(self)
 
-        if input_features is None:
-            input_features = self.variables_
-
-        return [feature for feature in input_features if feature in self.variables_]
+        return self.input_features_
--- a/feature_engine/encoding/decision_tree.py
+++ b/feature_engine/encoding/decision_tree.py
@@ -262,7 +262,7 @@ class DecisionTreeEncoder(BaseCategoricalTransformer):
 
     def get_feature_names(
         self,
-        input_features: List[str] = None,
+        input_features: Optional[List[Union[str, int]]] = None,
     ) -> List:
         """
         Return feature names for output features.
--- a/feature_engine/encoding/decision_tree.py
+++ b/feature_engine/encoding/decision_tree.py
@@ -260,24 +260,13 @@ class DecisionTreeEncoder(BaseCategoricalTransformer):
         """inverse_transform is not implemented for this transformer."""
         return self
 
-    def get_feature_names(
-        self,
-        input_features: Optional[List[Union[str, int]]] = None,
-    ) -> List:
+    def get_feature_names(self) -> List:
         """
         Return feature names for output features.
-
-        Parameters
-        ----------
-        input_features : list of str of shape (n_features,)
-            String names for input features if available. By default,
-            self.variables_ is used
 
         -------
         feature_names : list of str of shape (n_output_features,) of feature names
         """
         check_is_fitted(self)
 
-        return self.encoder_.named_steps.tree_discretiser.get_feature_names(
-            input_features
-        )
+        return self.encoder_.named_steps.tree_discretiser.get_feature_names()
--- a/feature_engine/encoding/decision_tree.py
+++ b/feature_engine/encoding/decision_tree.py
@@ -5,6 +5,7 @@ from typing import List, Optional, Union
 
 import pandas as pd
 from sklearn.pipeline import Pipeline
+from sklearn.utils.validation import check_is_fitted
 
 from feature_engine.discretisation import DecisionTreeDiscretiser
 from feature_engine.encoding.base_encoder import BaseCategoricalTransformer
@@ -258,3 +259,25 @@ class DecisionTreeEncoder(BaseCategoricalTransformer):
     def inverse_transform(self, X: pd.DataFrame):
         """inverse_transform is not implemented for this transformer."""
         return self
+
+    def get_feature_names(
+        self,
+        input_features: List[str] = None,
+    ) -> List:
+        """
+        Return feature names for output features.
+
+        Parameters
+        ----------
+        input_features : list of str of shape (n_features,)
+            String names for input features if available. By default,
+            self.variables_ is used
+
+        -------
+        feature_names : list of str of shape (n_output_features,) of feature names
+        """
+        check_is_fitted(self)
+
+        return self.encoder_.named_steps.tree_discretiser.get_feature_names(
+            input_features
+        )
--- a/feature_engine/encoding/one_hot.py
+++ b/feature_engine/encoding/one_hot.py
@@ -6,6 +6,8 @@ from typing import List, Optional, Union
 import numpy as np
 import pandas as pd
 
+from sklearn.utils.validation import check_is_fitted
+
 from feature_engine.encoding.base_encoder import BaseCategoricalTransformer
 from feature_engine.variable_manipulation import _check_input_parameter_variables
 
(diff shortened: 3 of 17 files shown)


solegalli:
Hey @janrito thank you for the PR.

Could you please add an example or a link showing how the `get_feature_names` attribute is needed to work with eli5?

Of the top of my head I don't remember that sklearn transformers have this attribute. How does eli5 work with them?

Thank you

janrito:
Sure! I'll come back with an example, but here are a few references of the API:

https://scikit-learn.org/stable/modules/generated/sklearn.feature_extraction.text.TfidfVectorizer.html#sklearn.feature_extraction.text.TfidfVectorizer.get_feature_names

https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.OneHotEncoder.html#sklearn.preprocessing.OneHotEncoder.get_feature_names

janrito:
Here is an example with  toy model:

```python 
import eli5
import pandas as pd

from feature_engine.encoding import OneHotEncoder
from sklearn.linear_model import LogisticRegression

# artificial data
dat = pd.DataFrame.from_dict(
    {
        "var1": ["a"] * 10 + ["b"] * 4,
        "var2": ["c"] * 2 + ["d"] * 7 + ["e"] * 5,
        "target": [1] * 7 + [0] * 7,
    }
)

dat
```
Out:
```tsv
var1	var2	target
0	a	c	1
1	a	c	1
2	a	d	1
3	a	d	1
4	a	d	1
5	a	d	1
6	a	d	1
7	a	d	0
8	a	d	0
9	a	e	0
10	b	e	0
11	b	e	0
12	b	e	0
13	b	e	0
```

Setup a simple model pipeline 

```python
model = Pipeline(
    [
        ("enc", OneHotEncoder()),
        ("clf", LogisticRegression()),
    ]
)
# fit the artificial data
model.fit(dat[['var1', 'var2']], dat['target'])
```

ELI5 helps us explain what are the features that have positive impact on the model:

```python
eli5.explain_weights(model)
```
Screenshot 2021-08-04 at 17 20 48

We can also use it to explain a single prediction - although the API here is a bit clunkier:

```python
eli5.show_prediction(
    model.named_steps.clf,
    model.named_steps.enc.transform(
        pd.DataFrame.from_dict([{"var1": "a", "var2": "c"}])
    ),
    feature_names=model.named_steps.enc.get_feature_names(),
)

```
Screenshot 2021-08-04 at 17 22 21


This is just one example of the use of `get_feature_names`, I'm sure there are more. It's also worth noting that there is currently an effort to improve this API in order to make introspecting into pipelines and understanding models easier, see here: https://github.com/scikit-learn/scikit-learn/pull/18444

In the case of `feature_engine`'s encoders, there are some aspects that can cause issues in the future and are worth thinking about -  for example, every encoder takes a `variable` parameter, which subselects the columns that will be modified by the encoder, leaving the rest of the columns untouched. This means that the number and order of features on the output of the encoder can vary, which can cause problems down the line. You will find that some of my test include the line: 

```python
assert encoder.get_feature_names() == X.columns.tolist()
```
and others don't, this has to do with whether the encoder (or transformer) is only partially modifying the input.

In any case, I'm happy to contribute further

solegalli:
Thank you @janrito that was very enlightening. 

So if I understand this correctly, `get_feature_names` should capture **all the feature names** of the output (transformed) dataset, including the new added ones (if any) and the ones that were not modified, **in the order** that they appear in the (output) dataset.

Is this correct?

**Food for thought:**
For consistency with sklearn, should we not add as well the method `get_feature_names_out`? As per [this comment](https://github.com/scikit-learn/scikit-learn/issues/20581#issuecomment-884417824). Not too sure how this 2 methods are going to play together yet, or if get_features_names will be deprecated. Would it be worth waiting and seeing how those methods are finally incorporated into sklearn?


> In the case of `feature_engine`'s encoders, there are some aspects that can cause issues in the future and are worth thinking about - for example, every encoder takes a `variable` parameter, which subselects the columns that will be modified by the encoder, leaving the rest of the columns untouched. This means that the number and order of features on the output of the encoder can vary, which can cause problems down the line. You will find that some of my test include the line:

This is also true for all sklearn transformers, am I correct? I was aware of this issue, but I thought that users might be aware since in sklearn feature order, or feature names for that matter,  are not controlled currently. So I thought we could live with it for the time being. But I agree that it is very dangerous and prone to serious errors.

Regarding this code:
> ```python
> feature_names=model.named_steps.enc.get_feature_names(),
> ```

Instead of the above, could one not directly pass the dataframe column names to eli?

```
feature_names=model.named_steps.enc.transform(
        pd.DataFrame.from_dict([{"var1": "a", "var2": "c"}])
    ).columns,
```
should do the trick. We would only need the names of the final data transformation, just prior to entering the model.

Or in other words, where / how else would the method `get_feature_names`  be used / valuable?

feature_engine/base_transformers.py
```diff
@@ -144,3 +144,26 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
     # for the check_estimator tests
     def _more_tags(self):
         return _return_tags()
+
+    def get_feature_names(
+        self,
+        input_features: List[str] = None,
```
solegalli, line 150: currently, feature-engine also supports integers as column names, see for example [this](https://github.com/solegalli/feature_engine/blob/master/feature_engine/imputation/mean_median.py#L64).

feature_engine/base_transformers.py
```diff
@@ -144,3 +144,26 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+    ) -> List:
+        """
+        Return feature names for output features.
+
+        Parameters
+        ----------
+        input_features : list of str of shape (n_features,)
+            String names for input features if available. By default,
+            self.variables_ is used
```
solegalli, line 159: self.variables_ contains the variables that are modified by the transformer. Not all the variables in the dataframe. Why would this be the default?

feature_engine/base_transformers.py
```diff
@@ -144,3 +144,26 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+
+        -------
+        feature_names : list of str of shape (n_output_features,) of feature names
+        """
+        check_is_fitted(self)
+
+        if input_features is None:
+            input_features = self.variables_
+
+        return [feature for feature in input_features if feature in self.variables_]
```
solegalli, line 169: I don't think this would capture the names of all features in the final transformed dataframe. Is this the desired output of this method? maybe I did not understand the aim of the function?

solegalli (commented):
Went briefly over the code, not too sure I understand the aim of get_features_names, or if the implementation matches the desired output. Would you be so kind to clarify that for me?

Also, the code seems almost identical for every transformer, I wonder if we can capture that in a function in dataframe checks or variable manipulation and then implement it across classes.

feature_engine/base_transformers.py
janrito: Thanks for the reference, I've fixed it here: c784f6d311a4e5dafd5d18ea5a40ceefa2702e2c.

I'm following the implementation in scikit's OneHotEncoder as an example https://github.com/scikit-learn/scikit-learn/blob/efede4b1864492c18238f60ffc39e9729672382c/sklearn/preprocessing/_encoders.py#L658. 

From what I understand, the `input_features` argument servers to limit the desired output features by the ones that directly map from features included. For example, `feature_engine.OneHotEncoder` can be fitted to encode multiple columns - calling `get_feature_names()` on the fitted encoder would return all transformed columns (`column_1 * cats_in_column_1 + column_2 * cats_in_column_2 ...`); however calling `get_feature_names(input_features=['column_1'])` would just return the features derived from that column: `column_1 * cats_in_column_1 `.

This is not essential for compatibility with eli5, I don't think. My intention was to follow a reference implementation of `get_feature_names`

feature_engine/base_transformers.py
janrito: Yes, this is a problem. I didn't want to modify the internal behaviour of the encoders before I showed it to you. 

I could use a different variable to keep every column in the input on `fit`, and use that as the basis to determine the output features.

feature_engine/base_transformers.py
janrito: I thought that for the generic case this would capture output features. Am I missing something?

feature_engine/base_transformers.py
solegalli: I see. My understanding is that sklearn has this functionality, because as it returns numpy arrays, you more or less have to guess which feature comes from which original variable. There is similar behaviour with the PolynomialFeatures class. So to help the user they created this method.

Feature-engine does not have that problem. If new variables are added, their name starts with the original variable name. Or alternatively, the user provides the names. So you always know what you get.

So thinking it further, I am not even sure we need a method here. If we were to store the name and order of the original variables, as they appeared in the training set, we would just need to create a new attribute, `feature_names_` in all transformers, and add it to the doscstrings.

But I need more convincing that this is useful / urgent :p

feature_engine/base_transformers.py
solegalli: this particular transformer captures only numerical variables by default or checks that the variables entered by the user are of type numerical.

I think the methods wants to capture all variables in the dataset, so list(X.columns) should do. But as per my previous comment, I think an attribute is enough, we don't need a  method. And also, would like to understand more about the value of this attribute. Because we basically need to modify all transformers. It is some work

feature_engine/base_transformers.py
janrito: I don't think pre-calculating the feature names is a problem, or even storing it as a new attribute.  
I would only argue that `get_feature_names` is currently the expected API call for this - even if it is looking like it is going to change in the future. 

My use-case - eli5 - for example,  expects this as the way to retrieve the names of the features. But also pipelines are composable, so if one uses feature_engine encoders alongside ColumnTransformers - it is expected that features will be able to be retrieved in the same form: https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/compose/_column_transformer.py#L372

I'm not sure I can make an honest argument on urgency, based on my own usecase :)

I can point to about [100k code results](https://github.com/search?q=get_feature_names&type=code) for the function name, [~250 issues](https://github.com/search?q=get_feature_names&type=issues) across multiple projects, and a [5-year-old](https://github.com/scikit-learn/scikit-learn/issues/6425) effort to unify the behaviour in scikit-learn

solegalli: OK then, I won't oppose any further :)

Thanks a lot for the links and the initiative.

Would it be ok to make the method so that it takes no argument?

I think for most transformers list(X.columns) should do. 

Special treatment would be required for the transformers in the modules creation, selection and wrappers and for the one hot encoder, of the top of my head.

Shall I leave this with you?

janrito: 681bc8a

Some tests are failing on the one hot encoder. I think it's because there's an inconsistency on columns that should get dropped

feature_engine/base_transformers.py
janrito: d452ef2

janrito: I think these are looking quite simple now, it would be nice if we could add them to all transformer at once?

feature_engine/base_transformers.py
```diff
@@ -97,6 +97,9 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None) -> pd.DataFrame:
         # find or check for numerical variables
         self.variables_ = _find_or_check_numerical_variables(X, self.variables)
 
+        # save input features
+        self.input_features_: List[Union[str, int]] = X.columns.tolist()
```
solegalli, line 101: For some transformers, it would not be only the input features, but the input features plus the features that are added or removed. So I would use a more neutral name here to unify across classes like feature_names_

feature_engine/base_transformers.py
```diff
@@ -144,3 +147,14 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
     # for the check_estimator tests
     def _more_tags(self):
         return _return_tags()
+
+    def get_feature_names(self) -> List:
+        """
+        Return feature names for output features.
+
+        -------
```
solegalli, line 155: We need to use the doscstring convention that I use in all files. We can not have ---- without a title. Here it should be Parameters i think. Pls check other files.

feature_engine/encoding/decision_tree.py
```diff
@@ -258,3 +259,14 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
     def inverse_transform(self, X: pd.DataFrame):
         """inverse_transform is not implemented for this transformer."""
         return self
+
+    def get_feature_names(self) -> List:
```
solegalli, line 263: can this not be inherited from basecategorical transformer?

feature_engine/encoding/one_hot.py
```diff
@@ -284,3 +286,22 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+        """
+        Return feature names for output features.
+
+        -------
+        feature_names : list of str of shape (n_output_features,) of feature names
+        """
+        check_is_fitted(self)
+
+        feature_names = []
+        for feature in self.input_features_:
```
solegalli, line 300: Oh, I see... you want to have the parameter input_features_ in fit, and then add those features that are added or removed in the function get_feature_names(). Yes, I like this better.

feature_engine/base_transformers.py
solegalli: > 
> 
> I think these are looking quite simple now, it would be nice if we could add them to all transformer at once?

That would be great!

solegalli:
Hi @janrito 

I like the implementation of input_features_ in the fit method, and then get_feature_names() adding additional features if necessary. This is looking really good.

It would be great if you could expand this to all classes.

I will be on hols until the 29th, so I won't be able to check the PR until then. But I can make it as soon as I am back. 

I am not sure why the OHE test fails. For some reason is putting the target at the end. If you don't manage to resolve it, or if more tests fail, we can look at them together when I m back.

Thanks a lot!

PS: a student in one of my courses had a SelectFromModel class inside of a pipeline, and asked me how to find out which features were selected. And I realised that it is not easy, because X_train suffers a lot of transformations before select from model. So this functionality will be really useful. Thanks for the initiative :)

solegalli:
FYI:
https://scikit-learn.org/stable/auto_examples/release_highlights/plot_release_highlights_1_0_0.html#feature-names-support

sklearn has already implemented this to some of its transformers. Good idea to carry on with this issue @janrito :)

kylegilde:
it looks like the get_feature_names method is being deprecated and replaced with get_feature_names_out in sklearn

PR #308, opened 2021-09-13
Add match variables closes #132

thibaultbl:
Add transformer to ensure that columns in test dataset are similar to train dataset.


solegalli:
closes #132 

Check that PR for discussions, planning and code revisions

PR #311, opened 2021-09-28
Psi selector

### LINKED ISSUE #307: Selection based on PSI values
As a new feature, it would be nice to have a selection to remove all features having a PSI value higher than a given threshold. The Population Stability Index (PSI) is a widely used measure for population shift in a feature. If a shift has happened (i.e. if the PSI is above a given level), the feature is not fit for purpose and usually removed from the feature set.

A way to deal with it would be to have a new selector

```python
from feature_engine.selection import DropHighPSIFeatures

drop_psi = DropHighPSIFeatures(threshold=0.25)
drop_psi.fit(X_train, X_test)

X_train_adjusted = drop_psi.transform(X_train)
X_test_adjusted = drop_psi.transform(X_test)
``` 
Currently I have own implementation based on the PSI calculations from Probatus. 

Happy to pick-up the task if you believe it has added value for feature engine.

gverbock:
Sorry, I put the PR at the wrong place.

closes #307


gverbock:
This was the comment I put in the original PR
```
Hi I have here a first version of the code for dropping features with high PSI. I still need to polish the code and work on the documentation but I believe it would be great if you can have a first look at it.

I am also wondering if you use tools like black and mypy to perform automated check. This would ensure my code has a similar layout uses the same coding rules as you do.
```

I see you are using flake8 and typecheck. I'll do it too so that it pass the tests.

feature_engine/selection/drop_high_PSI_features.py
```diff
@@ -0,0 +1,460 @@
...
+            not isinstance(min_pct_empty_buckets, (float, int))
+            or min_pct_empty_buckets < 0
+        ):
+            raise ValueError("min_pct_empty_buckets must be larger or equal to 0")
+
+        if missing_values not in ["raise", "ignore", "include"]:
+            raise ValueError(
+                "missing_values can only be 'raise', 'ignore' or 'include'."
+            )
+        if method.lower() in ["equalwidth", "equal_width", "equal width"]:
```
solegalli, line 455: I think parameters should accept only one value, we should not try and interpret what the user tried to do. It is very nice of you, but I think in the long run it may cause problems, for example backwards compatibility. 

What values are allowed, are explained in the docstings and we could help with an error message here

solegalli:
> 
> 
> This was the comment I put in the original PR
> 
> ```
> Hi I have here a first version of the code for dropping features with high PSI. I still need to polish the code and work on the documentation but I believe it would be great if you can have a first look at it.
> 
> I am also wondering if you use tools like black and mypy to perform automated check. This would ensure my code has a similar layout uses the same coding rules as you do.
> ```
> 
> I see you are using tox which I didn't know. I'll install it and run it.

we do use mypy, check with flake8 and have black. And I have just seen that it is not in the docs, so I should probably add that asap.

If you install tox. then you can run `tox ` from the terminal and that will test all all of these, including running black to tidy the code. But I normally test that manually, on the files I am working with.

see for example here: https://github.com/feature-engine/feature_engine/blob/main/tox.ini#L45

feature_engine/selection/drop_high_PSI_features.py
```diff
@@ -0,0 +1,460 @@
...
+    In DropHighPSIFeatures two approaches are implemented though the
+    "basis" argument.
+
+    - If it is a pandas.DataFrame, the class will compare the distributions of the
+    X dataframe (argument of the fit method) and "basis". The two dataframes must
+    contain the same features (i.e. labels).
+
+    - If it is a dictionary, the X matrix of the fit method is split in two according
+    to time and the distribution between the two parts are compared using the PSI.
+    The dictionary must contain the label of a column with dates and the cut-off date.
```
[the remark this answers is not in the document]
gverbock: The issue we would then have it that PSI is usually performed between the in-time (train) and the out-of-time dataset. So this would mean that we need to pass the "full" dataset to the sklearn pipeline which does not fit with other approaches (like dropping correlated features) that are usually done on the in-time data set.

gverbock:
@solegalli Thanks for the review. I will have a look at the comment and solve them.

feature_engine/selection/drop_high_PSI_features.py
```diff
@@ -0,0 +1,460 @@
...
+)
+
+Variables = Union[None, int, str, List[Union[str, int]]]
+
+
+class DropHighPSIFeatures(BaseSelector):
+    """
+    DropHighPSIFeatures drops features with a Population Stability Index (PSI) above
+    a given threshold. The PSI of a feature is an indication of the shift in its
+    distribution; a feature with high PSI might therefore be seen as instable.
```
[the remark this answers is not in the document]
gverbock: I have been thinking about that also as it is not a pure selection such as using correlation. Could we have a module for feature stability?

feature_engine/selection/drop_high_PSI_features.py
```diff
@@ -0,0 +1,460 @@
...
+from feature_engine.selection.base_selector import BaseSelector
+from feature_engine.variable_manipulation import (
+    _check_input_parameter_variables,
+    _find_or_check_numerical_variables,
+)
+
+Variables = Union[None, int, str, List[Union[str, int]]]
+
+
+class DropHighPSIFeatures(BaseSelector):
```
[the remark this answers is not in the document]
gverbock: It is addressed as a drop because it is not about keeping the "best" but about removing the features that may perturb the model.

feature_engine/selection/drop_high_PSI_features.py
gverbock: Solved by keeping only the original names equal_width and equal_frequency

feature_engine/selection/drop_high_PSI_features.py
```diff
@@ -0,0 +1,460 @@
+from typing import List, Union
+
+import numpy as np
+import pandas as pd
+
+from feature_engine.dataframe_checks import _check_contains_na, _is_dataframe
+from feature_engine.discretisation.equal_frequency import EqualFrequencyDiscretiser
+from feature_engine.discretisation.equal_width import EqualWidthDiscretiser
```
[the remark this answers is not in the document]
gverbock: Adjusted according to your suggestion.

feature_engine/selection/drop_high_PSI_features.py
```diff
@@ -0,0 +1,460 @@
...
+
+    The comparison of the distribution is done through binning. Two strategies are
+    implemented: equal_frequency and equal_width. These labels refer to two
+    discretisation implementation from the present package.
+
+    References:
+    https://scholarworks.wmich.edu/cgi/viewcontent.cgi?article=4249&context=dissertations
+
+    DropHighPSIFeatures() works only with numerical variables. Categorical variables
+    will need to be encoded to numerical or will be excluded from the analysis.
```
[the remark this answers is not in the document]
gverbock: For nominal variable a Chi-square or Kolmogorov-Smirnov test would probably be more appropriate. For discrete variables it depends on the cardinality.
If you want to use PSI on a discrete variable with 5 features, you need to bin the variable using 5 equally spaced categories.

feature_engine/selection/drop_high_PSI_features.py
```diff
@@ -0,0 +1,460 @@
...
+    variables_:
+        The variables to consider for the feature selection.
+
+    n_features_in_:
+        The number of features in the train set used in fit.
+
+    psi:
+        Dataframe containing the PSI values for all features considered.
+
+    features_to_drop_:
```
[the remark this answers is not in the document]
gverbock: Good catch. I removed that piece of code.

feature_engine/selection/drop_high_PSI_features.py
```diff
@@ -0,0 +1,460 @@
...
+        Threshold above which the distribution of a feature has changed so much that
+        the feature will be dropped. The most common values are 0.25 (large shift)
+        and 0.10 (medium shift).
+
+    method: string or callable, default='equal_frequency'
+        Type of binning used to represent the distribution of the feature. In can be
+        either "equal_width" for equally spaced bins or "equal_frequency" for bins
+        based on quantiles.
+
+    n_nbins, int, default = 10
```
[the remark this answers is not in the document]
gverbock: Changed to bins.

feature_engine/selection/drop_high_PSI_features.py
```diff
@@ -0,0 +1,460 @@
...
+        If set to true the role of the two matrices involved in the PSI calculations
+        (basis and measurement) will be switch. This is an important option as the
+        PSI value is not symmetric (i.e. PSI(a, b) != PSI(b, a)).
+
+    threshold: float, default = 0.25.
+        Threshold above which the distribution of a feature has changed so much that
+        the feature will be dropped. The most common values are 0.25 (large shift)
+        and 0.10 (medium shift).
+
+    method: string or callable, default='equal_frequency'
```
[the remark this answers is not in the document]
gverbock: Done.

feature_engine/selection/drop_high_PSI_features.py
```diff
@@ -0,0 +1,460 @@
...
+    psi:
+        Dataframe containing the PSI values for all features considered.
+
+    features_to_drop_:
+        List of features that are dropped because of PSI values above threshold.
+
+    Methods
+    -------
+    fit:
+        Find correlated features.
```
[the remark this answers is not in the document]
gverbock: That's a mistake. I change the comment

feature_engine/selection/drop_high_PSI_features.py
```diff
@@ -0,0 +1,460 @@
...
+    transform:
+        Remove correlated features.
+    fit_transform:
+        Fit to the data. Then transform it.
+    _compute_psi:
```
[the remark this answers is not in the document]
gverbock: I didn't know that. Thanks for explaining

feature_engine/selection/drop_high_PSI_features.py
```diff
@@ -0,0 +1,460 @@
...
+        missing_values: str = "include",
+        switch_basis=False,
+        threshold: float = 0.25,
+        n_bins=10,
+        method="equal_frequency",
+        min_pct_empty_buckets=0.0001,
+    ):
+
+        # Set all arguments (except self and variables as attributes)
+        for name, value in vars().items():
```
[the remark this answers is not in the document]
gverbock: Yes, It takes all variables (except self and variables) and assign it to a "self equivalent". This avoids to do it manually as you mention. If you prefer I can switch to manually assignment.

feature_engine/selection/drop_high_PSI_features.py
```diff
@@ -0,0 +1,460 @@
...
+            switch_basis,
+            threshold,
+            n_bins,
+            method,
+            min_pct_empty_buckets,
+        )
+
+    def fit(self, X: pd.DataFrame, y: pd.Series = None):
+        """
+        Perform all steps needed for the the PSI calculations of all features.
```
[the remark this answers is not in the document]
gverbock: That indeed vague. I changed to 
```
Assess the features that needs to be dropped because of high PSI values. 
```

feature_engine/selection/drop_high_PSI_features.py
```diff
@@ -0,0 +1,460 @@
...
+        """
+        # check input dataframe
+        X = _is_dataframe(X)
+
+        # find all variables or check those entered are present in the dataframe
+        self.variables_ = _find_or_check_numerical_variables(X, self.variables)
+
+        if self.missing_values == "raise":
+            # check if dataset contains na
+            _check_contains_na(X, self.variables_)
```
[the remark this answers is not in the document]
gverbock: Done

feature_engine/selection/drop_high_PSI_features.py
```diff
@@ -0,0 +1,460 @@
...
+        return self
+
+    def _compute_PSI(self, df_basis, df_meas, bucketer):
+        """
+        Compute the PSI by first bucketing (=binning) and then call the psi calculator
+        for each feature.
+
+        Parameters
+        ----------
+        df_basis : pandas dataframe of shape = [n_1, n_features]
```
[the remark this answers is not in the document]
gverbock: that is an arbitrary value. I replaced it by n_rows for the first dataframe and p_rows for the second dataframe to show the two dataframes do not need to have the same number of observations.

feature_engine/selection/drop_high_PSI_features.py
```diff
@@ -0,0 +1,460 @@
...
+            Proxy for the measurement distribution.
+
+        Returns
+        -------
+        psi_value: float.
+            PSI value.
+        """
+        # Perform the binning for all features.
+        basis_binned = (
+            self.bucketer.fit_transform(series_basis.dropna()).value_counts().fillna(0)
```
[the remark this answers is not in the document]
gverbock: No. Missing values need to be removed from the binning anyway. The fillna(0) is to enforce an empty category is seen has having 0 observation in it.

feature_engine/selection/drop_high_PSI_features.py
```diff
@@ -0,0 +1,460 @@
...
+
+
+class DropHighPSIFeatures(BaseSelector):
+    """
+    DropHighPSIFeatures drops features with a Population Stability Index (PSI) above
+    a given threshold. The PSI of a feature is an indication of the shift in its
+    distribution; a feature with high PSI might therefore be seen as instable.
+
+    In fields like Credit Risk Modelling, the elimination of features with high PSI
+    is frequent and usually required by the Regulator.
```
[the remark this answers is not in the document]
gverbock: Thanks ;-)

feature_engine/selection/drop_high_PSI_features.py
```diff
@@ -0,0 +1,460 @@
...
+
+        missing_values: str
+            Takes values 'raise' and 'ignore'. Whether the missing values should
+            be raised as error or ignored when determining correlation.
+
+        Returns:
+            None
+
+        """
+        if not isinstance(n_bins, int) or n_bins <= 1:
```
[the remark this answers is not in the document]
gverbock: I rewrote as 

```
def _check_init_values(
        self, basis, variables, missing_values, switch_basis, threshold, bins, strategy, min_pct_empty_buckets,
    ):
        """
        Raise an error if one of the arguments is not of the expected type or 
        has an inadequate value.
```

and moved the checks before the assignment

feature_engine/selection/drop_high_PSI_features.py
```diff
@@ -0,0 +1,460 @@
...
+        returns the basis and the measurement dataframes needed for PSI calculations
+    _check_init_values:
+        Perform a series of check on the attributes.
+    """
+
+    # TODO: Implement the check on the types of the cut-off dates and the date
+    # column that need ot be the same.
+    def __init__(
+        self,
+        basis,
```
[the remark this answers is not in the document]
gverbock: Thanks. Maybe this is an option.
- Change the name basis into split, the default value is 0.5
- Have a column to sort by, default would be the index.
- If there are repeated values in the column used for sorting, computed the 'most natural split' to ensure that all observations with the same value in the sorted column belong to the same split.

For example:

split = 0.5 on the date columns for this dataframe

| Index   | Date   | Feature value |
|:-------|:------|:--------|
| 0         | 1 Jan 2020 | 27 |
| 1         | 2 Jan 2020 | 13 |
| 2         | 3Jan 2020 | 97 |
| 3         | 3 Jan 2020 | 71 |
| 4         | 4 Jan 2020 | 54 |
| 5         | 5 Jan 2020 | 32 || 

would lead to 

| Index   | Date   | Feature value |
|:-------|:------|:--------|
| 0         | 1 Jan 2020 | 27 |
| 1         | 2 Jan 2020 | 13 |
| 2         | 3Jan 2020 | 97 |
| 3         | 3 Jan 2020 | 71 |

and 

| 4         | 4 Jan 2020 | 54 |
| 5         | 5 Jan 2020 | 32 || 

Having the two values for 3 Jan in the same 'split'

feature_engine/selection/drop_high_PSI_features.py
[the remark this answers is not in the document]
solegalli: what else would go in a module called feature stability? a module should be something that groups related techniques. A module for 1 technique would be a bit of an overkill.

I wonder if this would fit in a module called "analysis"? l

feature_engine/selection/drop_high_PSI_features.py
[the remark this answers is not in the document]
solegalli: In this case, we should probably add somewhere in the docs that this method is suitable for continuous variables only. Or something on those lines. This sentence makes it sound like if I encode my categories, then I am good to go with the psi

solegalli:
Hi @gverbock 

Just to make sure I don't miss anything, let me know when you would like me to review again :)

cheers

feature_engine/selection/drop_high_PSI_features.py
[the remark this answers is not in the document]
gverbock: Analysis would definitely be a good option.

gverbock:
Hi @solegalli I am closing on the changes I wanted to make. Then I will review the docstring and will notify you when it is ready for a second round.

feature_engine/selection/drop_high_PSI_features.py
```diff
@@ -0,0 +1,460 @@
...
+    ):
+
+        # Set all arguments (except self and variables as attributes)
+        for name, value in vars().items():
+            if name not in ("self", "variables"):
+                setattr(self, name, value)
+
+        # Check the input is in the correct format.
+        self.variables = _check_input_parameter_variables(variables)
+        self._check_init_values(
```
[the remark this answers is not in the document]
gverbock: I moved it into a function because I found the code easier to read. But I have no problem in putting it directly into the init.

gverbock:
Dear @solegalli I am done with the changes I wanted to do so I guess it is ready for a second review round from your side.

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,537 @@
...
+    missing_values: str, default=ignore
+        Takes values 'raise' and 'ignore'. Whether the missing values should be raised
+        as error or ignored when determining correlation.
+
+    Attributes
+    ----------
+    features_to_drop_:
+        Set with the correlated features that will be dropped.
+
+    correlated_feature_sets_:
```
solegalli, line 122: this is probable a left over and should be removed?

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,537 @@
...
+        Groups of correlated features. Each list is a group of correlated features.
+
+    variables_:
+        The variables to consider for the feature selection.
+
+    n_features_in_:
+        The number of features in the train set used in fit.
+
+    psi:
```
solegalli, line 131: all attributes learned during fit need to have a trailing underscore, to follow sklearn conventions, so psi_

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,537 @@
...
+        Find features with high PSI values.
+    transform:
+        Remove features with high PSI values.
+    fit_transform:
+        Fit to the data. Then transform it.
+    """
+
+    def __init__(
+        self,
+        split_col: str = "use_df_index",
```
solegalli, line 146: I would default this to None, in which case it would use the index

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,537 @@
...
+        split_frac: float = 0.5,
+        split_distinct_value: bool = False,
+        variables: Variables = None,
+        missing_values: str = "include",
+        switch_basis: bool = False,
```
solegalli, line 151: could we call this just switch?

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,537 @@
...
+            variables,
+            missing_values,
+            switch_basis,
+            threshold,
+            bins,
+            strategy,
+            min_pct_empty_buckets,
+        )
+        # Set all arguments (except self and variables as attributes)
+        for name, value in vars().items():
```
solegalli, line 170: we should check that the parameters receive expected values before assignment

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,537 @@
...
+
+        Returns
+        -------
+        results_df: pandas dataframe.
+            Dataframe containing the PSI for each feature.
+        """
+        # Initialize a container for the results.
+        results = {}
+
+        # Compute the PSI for each feature excluding the column used for split.
```
solegalli, line 249: why would the columns used for splitting be part of variables_?

variables_ should contain only the variables that will be modified /analysed, to keep convention with other transformers

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,537 @@
...
+            Proxy for the measurement distribution.
+
+        Returns
+        -------
+        psi_value: float.
+            PSI value.
+        """
+        # Perform the binning for all features.
+        basis_binned = (
+            self.bucketer.fit_transform(series_basis.dropna()).value_counts().fillna(0)
```
solegalli, line 280: series_basis.dropna() is dropping the na from the original data. But we are giving the user options to raise include or ignore, should not capture that functionality here?

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,537 @@
...
+            or will be spit into a measurement and a basis dataframe.
+
+        Returns
+        -------
+        pandas dataframe with value below the
+        basis pandas dataframe
+        """
+        # Identify the values according to which the split must be done.
+        if self.split_col == "use_df_index":
+            reference = pd.Series(X.index.to_list()).dropna()
```
solegalli, line 372: can the index have na?

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,537 @@
...
+        else:
+            reference = X[self.split_col].dropna()
```
solegalli, line 374: should we allow the user to split a dataframe on a column that is not complete? what if the user passes a column with 30 or 50% NA? is the class going to work?

I am inclined to not allow columns with NA to split.

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,537 @@
...
+        ref : (np.array, pd.Series).
+            Series for which the nth quantile must be computed.
+
+        Returns
+        -------
+        cut_off: (float, int, str, object).
+            value for the cut-off.
+        """
+        # Ensure the argument is a pd.Series
+        if not isinstance(ref, pd.Series):
```
solegalli, line 415: why would it not be a pd.Series?

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,537 @@
...
+            ref = pd.Series(ref)
+
+        # If the value is numerical, use numpy functionalities
+        if isinstance(ref.iloc[0], (int, float)):
```
solegalli, line 419: when would the user pass a numerical column to divide the dataframe? instead of dates I mean? just the index? or anything else?

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,537 @@
...
+)
+
+Variables = Union[None, int, str, List[Union[str, int]]]
+
+
+class DropHighPSIFeatures(BaseSelector):
+    """
+    DropHighPSIFeatures drops features with a Population Stability Index (PSI) above
+    a given threshold. The PSI of a feature is an indication of the shift in its
+    distribution; a feature with high PSI might therefore be seen as instable.
```
solegalli, line 28: unstable

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,537 @@
...
+    is frequent and usually required by the Regulator.
+
+    The calculation of the PSI requires to compare two distributions.
+    In DropHighPSIFeatures the approach implemented is the following:
+
+    - The dataframe is split in two parts and the PSI calculations are performed by
+    comparing the distribution of the features in the two parts.
+
+    - The user defines the columns of the dataframe according to which the dataframe
+    will de split. The column can contain numbers, dates and strings.
```
solegalli, line 40: I understand the concept of dividing based on dates: we want to determine if a feature changes its distribution in time. This is what is normally asked in finance.

I understand the concept of dividing based on a numerical column, only if the numerical column is the index, and assuming that the index is ordered somehow and not random. When else would we use a numerical column to divide the data?

When would we use a numerical column to divide the data? I understand that it is technically possible. But would anyone divide based on categories in a real financial setting?

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,537 @@
...
+
+    - The user defines a split fraction. This is used to define the sizes of the two
```
solegalli, line 42: this would only be used if the split column is numerical? If dates, then it does not apply, right?

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,537 @@
...
+    dataframe. Is the split fraction is 50%, the two dataframe will have about the same
+    size. It is not guaranteed that the size will exactly match because the observations
+    are grouped by their label (i.e. their values in the split columns) so all
+    observations with the same label will automatically belong to the same dataframe.
+
+    - If the option split_distinct_value is activated, the number of distinct values
```
solegalli, line 48: I guess this would only apply if the column is categorical, but then, would one really use a categorical column to split the dataframe? thinking of finance?

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,537 @@
...
+    is used to make the split. For example a feature with values: [1, 2, 3, 4, 4, 4]
+    a 50% split without split_distinct_value will split into two parts [1, 2, 3] and
+    [4, 4, 4] while if the option is set on, the split will be [1, 2] and [3, 4, 4, 4].
+
+    The PSI calculations are not symmetric. The switch_basis argument allows to
```
solegalli, line 53: Somewhere in the docs, probably at the beginning, we should explain what the PSI is used for, and provide the formula

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,537 @@
...
+    https://scholarworks.wmich.edu/cgi/viewcontent.cgi?article=4249&context=dissertations
+
+    DropHighPSIFeatures suited for numerical variables and for discrete variables with
+    high cardinality. For categorical variables differences in distribution can be
+    investigated using a Chi-square or Kolmogorov-Smirnov test.
+
+    Parameters
+    ----------
+
+    split_col: string.
```
solegalli, line 70: add default value

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,537 @@
...
+        Label of the column according to which the dataframe will be split.
+
+    split_frac: float.
```
solegalli, line 73: add default value

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,537 @@
...
+        Ratio of the observations (when split_distinct is not activated) that goes
+        into the sub-dataframe that is used to determine the reference for the feature
+        distributions. The second sub-dataframe will be used to compare its feature
+        distributions to the reference ones.
+
+    split_distinct_values: boolean.
+        If set on, the split fraction does not account for the number of observations
+        but only for the number of distinct values in split_col.
```
solegalli, line 81: If I understand this correctly, we are giving the user multiple ways in which they can split the data. 

My question is: would someone really split the data other than with a date? from what I understand from PSI, it is aimed at determining shifts in the distribution in time. So if we do not have a time measurement to split the data, it kind of defeats the point. Let me know if I am wrong.

My second thought on this front: simple is better, or less is more. Giving the users too many options makes the class harder to use. So I would keep it as simple as possible. The less parameters the better.

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,537 @@
...
+
+    min_pct_empty_buckets, float, default = 0.0001
+        Value to add to empty bucket (when considering percentages). If a bin is
+        empty the PSI value may jump to infinity. By adding a small number to empty
+        bins, this issue is avoided. If the value added is too large, it may disturb
+        the calculations.
+
+    missing_values: str, default=ignore
+        Takes values 'raise' and 'ignore'. Whether the missing values should be raised
+        as error or ignored when determining correlation.
```
solegalli, line 115: correlation?

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,537 @@
...
+        Fit to the data. Then transform it.
+    """
+
+    def __init__(
+        self,
+        split_col: str = "use_df_index",
+        split_frac: float = 0.5,
+        split_distinct_value: bool = False,
+        variables: Variables = None,
+        missing_values: str = "include",
```
solegalli, line 150: ignore

gverbock: changed

feature_engine/selection/drop_psi_features.py
gverbock: PSI is used to compute population shift and a popular approach is to use time; for example remove features that are shifted comparing the pre-Covid to the Covid period. However it can also be used in other cases, like when combining two similar portfolios to make a common model. An example I know is the combination of two "similar" portfolios to build a common Credit Risk  model. The combination is needed because of the low number of observation in the positive class. PSI is then used to remove features than have different distribution between the two portfolios.

gverbock: I agree with the idea of less is more. I guess the first we need to agree on is if we want to have only a split based on time or if we want to allow a more general approach (and maybe touch a larger audience).
I can live with both approaches as I am personally using mostly  split according to time so for me the choice is yours.

feature_engine/selection/drop_psi_features.py
gverbock: Done

feature_engine/selection/drop_psi_features.py
gverbock: Done

feature_engine/selection/drop_psi_features.py
gverbock: In the current case, I made it generic so it would be used in all cases. This would say for example split my time scale in two equal parts (or in 75%, 25%). This "replace" them a cut-off date.

feature_engine/selection/drop_psi_features.py
gverbock: Corrected

feature_engine/selection/drop_psi_features.py
gverbock: When, on row 378 we have:

```
if self.split_distinct_value:
            cut_off = self._get_cut_off_value(reference.unique())
```
it passes a numpy.ndarray. The idea to transform into a pd.Series it to have the same approach independently of the value for self.split_distinct_value.

feature_engine/selection/drop_psi_features.py
gverbock: Yes, technically it can. I can even have multiple NA's
image

feature_engine/selection/drop_psi_features.py
gverbock: Sure. I changed it.

feature_engine/selection/drop_psi_features.py
gverbock: Changed

feature_engine/selection/drop_psi_features.py
gverbock: Make totally sense!

feature_engine/selection/drop_psi_features.py
gverbock: Good point. In the current case, include does not make much sense. I will remove the 'include' option for missing.

feature_engine/selection/drop_high_PSI_features.py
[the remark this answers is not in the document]
solegalli: Yes, I agree, let's keep it in selection, and we just make it very clear in the documentation, what the aim of this selector is, and how it is most commonly used.

feature_engine/selection/drop_psi_features.py
solegalli: Nice. I wasn't sure about this

feature_engine/selection/drop_psi_features.py
solegalli: Yes, makes sense, I see the point. 

Thank you for taking the time to explain. 

Let's do as you suggest. 

We will need to add a lot of info in the docs, probably with these examples you mention here to help users understand how to use this class. The examples are nice and clear.

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,566 @@
...
+
+        # Switch base and measurement dataframe if required.
+        if self.switch:
+            measurement_df, basis_df = basis_df, measurement_df
+
+        # Compute the PSI
+        self.psi_ = self._compute_PSI(basis_df, measurement_df, self.bucketer)
+
+        # Select features below the threshold
+        self.features_to_drop_ = self.psi_[
```
solegalli, line 240: we should probably make this part of the loop I suggest in my previous comment. Not to loop again.

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,566 @@
...
+            self.psi_.value >= self.threshold
+        ].index.to_list()
+
+        self.n_features_in_ = X.shape[1]
+
+        return self
+
+    def _compute_PSI(self, df_basis, df_meas, bucketer):
+        """
```
solegalli, line 249: I would remove this function

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,566 @@
...
+        Returns
+        -------
+        results_df: pandas dataframe.
+            Dataframe containing the PSI for each feature.
+        """
+        # Initialize a container for the results.
+        results = {}
+
+        # Compute the PSI for each feature excluding the column used for split.
+        for feature in self.variables_:
```
solegalli, line 275: with the code as it is, we are training 1 discretizer per variable, when we could apply the discretizer to the entire dataframe. I would change

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,566 @@
...
+                right_index=True,
+                left_index=True,
+                how="outer",
+            )
+            .fillna(0)
+        )
+        binning.columns = ["basis", "meas"]
+
+        # Compute the PSI value.
+        psi_value = self._psi(binning.basis.to_numpy(), binning.meas.to_numpy())
```
solegalli, line 326: I would return the 2 distributions ready as input of the psi function and leave that calculation to the following function, so we have functions with clear names, and thus clear delimited functionality

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,566 @@
...
+            or will be spit into a measurement and a basis dataframe.
+
+        Returns
+        -------
+        pandas dataframe with value below the
+        basis pandas dataframe
+        """
+        # Identify the values according to which the split must be done.
+        if self.split_col is None:
+            reference = pd.Series(X.index.to_list())
```
solegalli, line 397: why do we need to_list()?

Is the index of a dataframe not a pd.Series already?

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,566 @@
...
+                "Na's are not allowed in the columns used to split the dataframe."
+            )
+
+        # Define the cut-off point based on quantile.
+        if self.split_distinct_value:
+            cut_off = self._get_cut_off_value(reference.unique())
+        else:
+            cut_off = self._get_cut_off_value(reference)
+
+        self.quantile = {self.split_col: cut_off}
```
solegalli, line 413: Questions:

I guess self.quantile is because we want to inform the user where the cut-off is made? Do we really need to expose this value?

If it is learned during fit, the name has to have a trailing underscore to abide by sklearn convention. 

Also, is quantile the best name? 

We only allow 1 cut-of column, so I don't think we need to return a dictionary, I would just return the value. 

We are not giving the user the option to divide the data based on a certain date. We should probably add that, and we should allow the user to pass a string with a date within the init

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,566 @@
...
+
+        # Split the original dataframe in two parts: above and below cut-off
+        is_above_cut_off = reference > cut_off
```
solegalli, line 416: this also works for categorical?

feature_engine/selection/drop_psi_features.py
solegalli: You did, thank you for taking the time to explain :)

feature_engine/selection/drop_psi_features.py
solegalli: could we not do `cut_off = self._get_cut_off_value(pd.Series(reference.unique()))` instead?

Just thinking out loud, if I got confused, others might as well.

solegalli (commented):
Hi @gverbock 

Thanks a lot for the enormous amount of code and functionality. Much appreciated.

There is a lot going on in this class, so it took me a while to digest.

Thank you so much for thinking so carefully on all the options in which a user may want to divide the dataset. And also for taking the time to explain them to me :)

I added some suggestions throughout.

Let me know what you think

Cheers!

feature_engine/selection/drop_psi_features.py
solegalli: I got up thinking about this today: at the moment, we have functionality to split data based on proportions. But thinking of the examples you gave me, we should probably also add:

data ==> cut when df[split_col] > data
numerical value ==> cut when df[split_col] > number (thinking of the contract ID, I may want to split after a certain ID
categorical value ==> cut when df[split_col].isin([category list]), thinking on a column that says portfolio1, portfolio2 and portfolio3, I may want portfolios 2 and 3 in the meas df.

Do these make sense?

gverbock:
@solegalli Thanks a lot for the comments. I will have a look at it can come back to you.

feature_engine/selection/drop_psi_features.py
gverbock: We could if there are no missing values in the dataframe. Otherwise we need to either use .dropna() or impute the missings. In the two cases we change the distribution of the features which then affects the values of the PSI.
In my first draft I was using the discretizer on the full dataframe but I changed it back to train a single discretizer per variable... maybe there is a way around this but I haven't find it (yet).

In banking industrie, the number of missing values can be very large. This happens for example looking at customer's profiles. Most customers have a current account but only a small part of the customers has products like  personal loans or securities. The feature with the total balance / debt is missing for everyone that does not have the product.

feature_engine/selection/drop_psi_features.py
gverbock: The to list can be remove but the index is not a pd.Series it is a (one of the many) index object.

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,566 @@
...
+        )
+        binning.columns = ["basis", "meas"]
+
+        # Compute the PSI value.
+        psi_value = self._psi(binning.basis.to_numpy(), binning.meas.to_numpy())
+
+        return psi_value
+
+    def _psi(self, d_basis, d_meas):
+        """
```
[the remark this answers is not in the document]
gverbock: Done. Hope the new structure makes it easier to follow what is happening.

feature_engine/selection/drop_psi_features.py
gverbock: I restructured the code to make it simpler

feature_engine/selection/drop_psi_features.py
solegalli: I am having second thoughts on whether to call the param split_frac or test_size. What 's your view?

tests/test_selection/test_drop_high_psi_features.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+def test_variable_definition(df):
+
+    select = DropHighPSIFeatures(variables=["var_1", "var_3", "var_5"], split_frac=0.01)
+    transformed_df = select.fit_transform(df)
+
+    assert transformed_df.columns.to_list() == ["var_0", "var_2", "var_4"]
+
+
```
solegalli, line 238: Could we include some tests for the following (some of them are there already):

1. that the split column is not assigned to the variables when those are detected automatically
2. when missing_values is raised and user passes df with na, an error is raised
3. test that _observation_frequency_per_bin is working as expected
4. test for splitting on index and on column
5. test error is raised if split_col has na
6. test split based on date, int and float entered by user in cut_off (both index and col)
7. test split based on fraction with or without unique values for floats and strings, would it also work if datetime format?

gverbock: I was definitely planning to add more tests as we extended the possibilities to split the dataframe. I'll have a good look at the functionalities that have not yet been tested and at the comments you put.

feature_engine/selection/drop_psi_features.py
```diff
@@ -451,11 +445,11 @@ def _split_dataframe(self, X):
                 "dataframe."
             )
 
-        # If cut_off is not pre-defined, compute it.
+        # If cut_off is not pre-defined, compute it and set it as attribute.
         if not self.cut_off:
-            cut_off = self._get_cut_off_value(reference)
-        else:
-            cut_off = self.cut_off
+            self.cut_off = self._get_cut_off_value(reference)
```
solegalli, line 450: we can't override parameters set by the user. That is, by sklearn convention. The transformer would not pass the check_estimator tests like this. 

This is why I had a temporal variable cut_off.

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,523 @@
...
+
+        Returns
+        -------
+        distribution.basis: pd.Series.
+            Basis Pandas Series with percentage of observations per bin.
+
+        distribution.meas: pd.Series.
+            Test Pandas Series with percentage of observations per bin.
+        """
+        # TODO: were the fillna in previous version needed?
```
solegalli, line 389: is this still outstanding?

tests/test_selection/test_drop_high_psi_features.py
```diff
@@ -0,0 +1,410 @@
...
+        class_sep=2,
+        random_state=1,
+    )
+
+    # transform array into pandas df
+    colnames = ["var_" + str(i) for i in range(6)]
+    X = pd.DataFrame(X, columns=colnames)
+
+    # Add drifted features that will be dropped during transformation.
+    X["drift_1"] = [itel for itel in range(X.shape[0])]
```
solegalli, line 29: what does itel stand for?

solegalli:
Hi @gverbock 

Thank you so much for the amount of work!

I made a PR to your repo with some changes: https://github.com/gverbock/feature_engine/pull/2

Let me know what you think.

gverbock:
@solegalli Thank you for the support. I did not see that you had a PR before my last push so I will have a look at it and merge it to the last version.

feature_engine/selection/drop_psi_features.py
gverbock: I learned that the hard way. Now I created a self.cut_off_ attribute so that the value of the cut_off is available to the user when the cut_off is not explicitly available.

feature_engine/selection/drop_psi_features.py
gverbock: I removed it.

tests/test_selection/test_drop_high_psi_features.py
gverbock: That's a bad habit I have to call itel in an iteration. I changed it to number.

feature_engine/selection/drop_psi_features.py
solegalli: A quick note, if the class has the new attribute `self.cut_off_`, it also needs  to exist when the user defines cut_off themselves, in which case I guess, it would be the same value. In other words, the class should always have that parameter and not just when cut_off is None.

gverbock: Yes, that is howe is it currently designed. self.cut_off_ is either assigned to the cut_off defined in the init or computed.
image

gverbock:
Hi @solegalli, I made some changes in the tests. I believe it is ready for review

tests/test_selection/test_check_estimator_selectors.py
```diff
@@ -24,7 +24,7 @@
         DropConstantFeatures(),
         DropDuplicateFeatures(),
         DropCorrelatedFeatures(),
-        DropHighPSIFeatures(bins=5),
+        DropHighPSIFeatures(bins=4),
```
solegalli: nicely caught! thank you

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,539 @@
...
+        If `split_distinct` is True, the indicated fraction may not be achieved exactly.
+        See parameter `split_distinct` for more details.
+
+        If `cut_off` is not None, `split_frac` will be ignored and the data split based
+        on the `cut_off` value.
+
+    split_distinct: boolean, default=False.
+        If True, unique values in `split_col` will go to either basis or test data sets
+        but not both. For example, if `split_col` is [0, 1, 1, 1, 2, 2], `split_frac` is
+        0.5 and `split_distinct` is False, the data will be divided ind [0, 1, 1] and
```
solegalli, line 128: in instead of ind

feature_engine/selection/drop_psi_features.py
```diff
@@ -0,0 +1,539 @@
...
+    variables_:
+        The variables to evaluate.
+
+    psi_values_:
+        Dictionary containing the PSI value per feature.
+
+    n_features_in_:
+        The number of features in the train set used in fit.
+
+    cut_off_:
```
solegalli, line 193: cut_off_ should go before n_features_in_

tests/test_selection/test_drop_high_psi_features.py
```diff
@@ -0,0 +1,543 @@
...
+    "split_distinct": False,
+    "cut_off": None,
+    "switch": False,
+    "threshold": 0.25,
+    "bins": 10,
+    "strategy": "equal_frequency",
+    "min_pct_empty_bins": 0.0001,
+    "missing_values": "raise",
+    "variables": None,
+}
```
solegalli, line 125: the 2 dictionaries are identical... this is probably a bug? we want to test that different parameter values are correctly assigned.

tests/test_selection/test_drop_high_psi_features.py
```diff
@@ -0,0 +1,543 @@
...
+    assert cut_offs == expected_cut_offs
+
+    # Test when no columns is defined
+    test = DropHighPSIFeatures(split_frac=0.5)
+    test.fit_transform(df_mixed_types)
+    assert test.psi_values_ == pytest.approx({"A": 8.283089355027482, "B": 0.0}, 12)
+
+
+def test_calculation_no_split_columns():
+    """Test the split of the dataframe using different type of variables."""
```
solegalli, line 296: I am still not clear what the aim of this test is. Is it to test `variables =None`? `split_col=None`? also, why do you set split_distinct to True for the test?

Could we add  the object(s) of the test in assert statements? This way I find it clearer to understand what the objective of the test is

tests/test_selection/test_drop_high_psi_features.py
```diff
@@ -0,0 +1,543 @@
...
+    """Test the cut off for different data types.
+
+    For columns B, C and time we have 6 distinct values, 5 appearing 20 times and
+    1 appearing 100 times. A 50% split based on the number of values will results
+    in 2 groups of 3. One has 60 appearances (in total) and the other has 140.
+    """
+    data = pd.DataFrame(
+        {
+            "A": [it for it in range(0, 200)],
+            "B": [1, 2, 3, 4, 5, 6, 6, 6, 6, 6] * 20,
```
solegalli, line 379: If B was instead [1, 1, 1, 1, 1, 2, 3, 4, 5, 6], does this still work as expected? 

could we add a test for this example? 

because I can't see it happening as expected from the logic in the class.

tests/test_selection/test_drop_high_psi_features.py
```diff
@@ -0,0 +1,543 @@
...
+        split_col="A", split_frac=split_frac, split_distinct=False
+    )
+    test.fit_transform(df)
+    assert test.cut_off_ == expected
+
+
+def test_calculation_distinct_value():
+    """Test the calculation of the quantiles using distinct values."""
+    df = pd.DataFrame(
+        {"A": [it for it in range(0, 30)], "C": ["A", "B", "C", "D", "D", "D"] * 5}
```
solegalli, line 257: In: `"C": ["A", "B", "C", "D", "D", "D"] * 5` there are:

- 5 x A observations
- 5 x B observations
- 5 x C observations
- 15 x D observations

total = 30 observations

So if split_frac = 0.5 I expect the same result for both split_distinct=True or False ==> 50% is 15 labels, so that would be A, B and C in the basis set.

I think there is a problem with the logic in the class, or maybe I do not understand the logic.

A simpler test would be directly on  `"C": ["A", "B", "D", "D", "D", "D"] ` :

- if split_distinct=False I expect  ["A", "B", "D",] and ["D", "D", "D"]
- if split_distinct=True I expect  ["A", "B",,] and [ "D", "D", "D", "D"]

Also, can we  make sure split_distinct also works well when the split_col is numerical? That is:

If `"C": [1, 2, 3, 3, 3, 3] ` :

- if split_distinct=False I expect  [1,2,3,] and [3,3,3]
- if split_distinct=True I expect  1,2] and [ 3,3,3,3,]

solegalli:
Hi @gverbock 

I made a PR to your repo: https://github.com/gverbock/feature_engine/pull/3

and added a few questions in the test file mostly.

Could you have a look?

Thank you!

gverbock:
@solegalli Thanks. I will have a look and I will try to clearly explain the use of split distinct (which is a difficult task for me)

tests/test_selection/test_drop_high_psi_features.py
gverbock: The behaviour is what I expect but also probably something I did not explained correctly. Split distinct will look at the vector of unique values and split that vector according to the split_frac. 
image

This can be used to split a dataframe according to the customer ID (half customer in basis and half in test) regardless of the number of entries each customer has without having to compute the exact split_frac or the cut_off value.

gverbock: So if we take the following case 
image

gverbock: split_distinct will results in two dataframes of equal sizes wen applying ._split_dataframe
image

gverbock: While split_distinct = True results in two set of customer ID with equal size
image

gverbock: @solegalli I guess we first need to agree on that behaviour and then I will add the test you suggested.

tests/test_selection/test_drop_high_psi_features.py
gverbock: I see this test is redundant as it is included in the test just above it when doing 
image
The idea was to test string and date would not lead to an error when no split_column is defined.

I will remove the test.

tests/test_selection/test_drop_high_psi_features.py
gverbock: I see also the comment on the docstring is not clear about that. Once we agreed on the behaviour, I'll adjust it.

solegalli:
https://github.com/feature-engine/feature_engine/pull/311#discussion_r745104682

Hi @gverbock 

Thank you for the detailed explanation. That helps. I had understood something different from the docstrings.

The way I thought about this logic is:

I have different customer IDs, I want to separate the data into basis and test, sending 50% of the rows to each data set, but, I do not want the same customer to be in both data sets, so split_distinct would ensure that.

The above logic ensures that we have a split more or less similar to what is specified in split_frac, but without having the same customer in both datasets.

The way the current logic works may return dfs that are far of from the split_frac. For example, in the extreme case that 50% of my rows are for customer 1, and 50% of the remaining rows are for different other customers, then, I would end up with a very big basis df and a very small test df. Am I correct?

I was thinking of some logic similar to this: https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.GroupKFold.html#sklearn.model_selection.GroupKFold

And I wonder if we can borrow some code from here: https://github.com/scikit-learn/scikit-learn/blob/0d378913b/sklearn/model_selection/_split.py#L508

What do you think?

solegalli:
Ok, thinking it further... the logic that I propose, in the same extreme example where 50% of the data belongs to one customer, would return a basis df where we have data only for that customer, and a test set where we have data for all the rest, so it is not exactly representative :/

I can see that in this case the logic as you describe it makes more sense.

What is your view on this? I guess you have more experience in the sector?

solegalli:
As I go along my day, I continue thinking about this, and in fact, I think the GroupKFold is doing something similar to the splitting you propose, and not the one I thought of.

So I think we should go with your implementation, change the wording in the docstrings to make it more clear, and if poss expand the tests to make sure that it works as expected also when the variable is numeric.

This is probably your preferred way forward as well? otherwise let me know

gverbock:
Hi @solegalli This is indeed my preferred way forward. My next steps are:

- I will adjust the doc string to make it clear what split_distinct is doing.
- I will extend the tests.  
- For the name, I need to think about the best option. To me _group_ implies I will need to pass a series with labels but so far I do not have an alternative.

feature_engine/selection/drop_psi_features.py
```diff
@@ -123,12 +123,17 @@ class DropHighPSIFeatures(BaseSelector):
...
+        instead of being applied to the whole vector of values. For example, if the
+        values in `split_col` are [1, 1, 1, 1, 2, 2, 3, 4] and `split_frac` is
+        0.5, we have the following:
+        - `split_distinct=False` splits the vector in two equally sized parts:
+        [1, 1, 1, 1] and [2, 2, 3, 4]. This involves that 2 dataframes with 4
+        observations each are used for the PSI calculations.
+        - `split_distinct=True` computes the vector of unique values in `split_col`
+        ([1, 2, 3, 4]) and splits that vector in two equal parts: [1, 2] and [3, 4].
+        The number of observations in the two dataframes used for the PSI calculations
+        is respectively 6 ([1, 1, 1, 1, 2, 2]) and 2 ([3, 4]).
```
solegalli, line 136: Excellent! Crystal clear! Thank you!

solegalli:
Great work @gverbock thank you so much!

There is a lot going on in this class, so the development took a bit longer than usual. But I think it is now very versatile and can accommodate a lot of scenarios. Very well thought through!

The code base I would say is ready. We need to add some documentation. We need to let users know of all these scenarios that you considered.

The thing is, I am updating the documentation and changing the format in this PR: #314 

So I see 2 potential ways forward:

1. I merge this PR to main, rebase main onto PR #314, draft a skeleton for the docs and hopefully you can help me fill the examples?
2. wait until I finish #314 (probably end of this or next week), rebase the changes onto this branch, then I can draft a skeleton for the docs, and hopefully you help me fill the examples?

Do you have any preference?

gverbock:
Thanks for the support and the comments @solegalli . I am also glad with the result. I will surely help you with the documentation. Regarding the process, I leave it up to you. I have no preference.

solegalli:
@gverbock I am having problems rebasing main onto this PR, so I will merge and close, and start a new PR for the docs.

I hope that's OK.

PR #326, opened 2021-11-17
adds psi selection docs

solegalli:
Hi @gverbock 

I made a first draft of the documentation for the PSI selector. In short, I linked it to all the pages where it should be linked.

The major changes required are:

- [x] Add explanations and examples in  docs/user_guide/selection/DropHighPSIFeatures.rst
- [x] Simplify the docstrings in  feature_engine/selection/drop_psi_features.py

In fact, I think we should leave in the `.py` file just the bare minimum for the user to understand what the class does and what the parameters do. And pass all other text to the user guide `rst` above.

If you add examples using a dataset, please make sure it is publicly available, for example from the UCI Machine Learning repository. Alternatively, you can use toy data. But I think with a bigger dataset, we could perhaps show some plots on how the distribution changes (or not).

Minor edits:

- [x] see if you would like to edit  docs/api_doc/selection/index.rst
- [x] please edit  docs/user_guide/selection/index.rst to introduce the class in a more relevant manner

For help with documentation, check [this link](https://feature-engine.readthedocs.io/en/pydata_template/contribute/contribute_docs.html)

Don't forget to get the lastest changes from the repo main branch (there was a huge update) and sync your fork. For help syncing forks check [here](https://feature-engine.readthedocs.io/en/pydata_template/contribute/contribute_code.html#keep-your-fork-up-to-date)

After this PR we'll release a new version of Feature-engine with this and other latest additions :)

related to #311 and #307


gverbock:
Thanks for the explanation. I will have a look at it, probably at the beginning of next week.

gverbock:
@solegalli . While writing the documentation I made some tests and I realised there was a bug in the current code. When the `split_col` is not provided, the class uses the index, however this was not working correctly when the index was shuffled. Do you mind if I use the current PR to solve this issues? I will also write additional tests for the case the index is shuffled.

In addition I would like to propose a new name for the `switch` argument. I believe it would be more informative to call it `swap_binning_reference`. What do you think? I you agree I can also change this is the current PR.

solegalli:
Go for it. It is super common to find some small bugs when creating docs and examples. :)

gverbock:
@solegalli I wanted to make a first push of the documentation I had so far but I see that I created a new PR (sorry about that). Not sure what is the best way to solve this. I had issue with setting the upstream branch in my repo and I guess I messed up when trying to solve it.

solegalli:
We can continue working in the new PR if you prefer. But note that here 7 files were changed and in the new PR only 4. Pls, make sure you migrate all these changes to the new PR. Cheers

solegalli:
> 

I didn't realise that you PRed to this branch instead of main in #328 hence my comments (which now I see dont make a lot of sense, lol). Apologies.

That was a great idea, thanks.

PR #332, opened 2021-12-01
Correct `quantile` -> `quantiles`

david-cortes:
The docs for `Winsorizer` mention "quantile" as capping method, but it only accepts "quantileS".


solegalli:
@david-cortes thank you!

PR #333, opened 2021-12-01
adds Datetime transformer, closes #67

### LINKED ISSUE #67: include transformer for datetime variables
The first version of this module should include the following:

- New module: datetime (folder)

Three transformers:
- ExtractDateFeatures
- ExtractTimeFeatures
- ExtractDateTimeFeatures

A base class:
- DateTimeBaseTransformer

The base class should:

- check that the variables entered by the user are datetime, or if string / object, transform them to datetime (I wonder if we should make this a function part of [variable_manipulation](https://github.com/solegalli/feature_engine/blob/master/feature_engine/variable_manipulation.py)? and call it in this transfomrer? In case we develop the timeseries module, we may need it later for other modules as well.

- have a method to return time features
- have a method to return date features
- methods to check the input dataframe, and options on what to do if outliers are present

The ExtractDateFeatures should derive the following features:

- [ ] month, quarter, semester and year (all numeric outputs)
- [ ] week of the year
- [ ] is week of the month supported by pandas? if yes, when we should return it
- [ ] day (numeric 1-31), day of the week (numeric 0-6), and is_weekend (binary)
- [ ] anything else supported by pandas?
- [ ] anything else that would be useful

The ExtractTimeFeatures should extract the following features

- [ ] hr, minute, second
- [ ] timezone: with parameter `return_timezone=False` by default, we allow the user to return a time zone categorical feature
- [ ] unify_timezone=False, to handle different timezones. If True, the transformer is timezone aware, unifies to greenwich and then derived features (should we give option to user to unify to something else? probably yes)
- [ ]  **to discuss**: is_morning, is_afternoon, is_evening
- [ ] **to discuss**: working hrs (I am thinking of passing a string parameter like '9-17' and use those to determine this feature
- [ ] anything else that would be useful

The ExtractDateTimeFeatures is a sum of the previous transformers, so it should return all possible date and time features.

The reason I suggest to break this into 3 classes is because some timestamps contain only dates, some only times and some both. And I think it would be easier if the user, who knows the timestamp selects the appropriate transformer, instead of adding code to understand which type of timestamp it is, and then derive the features

To consider: should we in version 1 of this transformer return all possible features? or should we give the user an option of which features to return? example, the user may want year and month but not quarter and semester.

Example code in recipes 2 to 5 of the [this link.](https://github.com/solegalli/packt_featureengineering_cookbook/tree/master/ch07-datetime).

Things to think about for the transformers design:

- [ ] Transformer returns all new variables by default, or only those indicated by the user. This behavior could regulated by a parameter in the init. As per previous question, should we leave this for version 2 of this transformer? or add it straightaway?
- [ ] Transformer should be able to derive features from more than 1 datetime variable at the time, like all feature engine transformers.
- [ ] Option in Transformer to drops original datetime variables (as those are not useful for machine learning), drop_datetime_variables=True or similar. 

Files needed in addition to code files:
- [ ] add transformer to readme list
- [ ] add transformer to docs/index.rst
- [ ] add docs folder exclusive for this transformer with rst files with examples
- [ ] add jupyter notebooks showing how to use this transformers

This issue can be done in 1 big PR, or multiple PRs, maybe 1 per transformer.

dodoarg:
Not meant to be definitive but I thought it was worth a review at this point
based on the #67 comment todo list:

main features implemented:
- datetime module
- base class DateTimeBaseTransformer in base_transformers.py
- ExtractDateFeatures transformer 

main features not implemented:
- ExtractTimeFeatures, ExtractDateTimeFeatures
- docs

baseclass features implemented:
- checks the variables entered by user (or all if None), converts str/object labelled variables that actually contain datetime values into datetime. The core level functions are part of variable_manipulation.py and are called by the transformer. <br>
On this note, since pd.to_datetime(X) converts all numerical/object/category variables in the dataframe X to datetime - including numerical variables that are *not* actually datetime - the conversion called by the transformer acts at the pd.Series level and treats numerical variables as *non*-convertible to datetime. This doesn't have any drawbacks when it comes to extracting features via the transformer, as one cannot extract any additional datetime info from a column that only has info on, say, the year, or the month, while it eliminates the risk of getting rid of valuable numeric data that would be parsed onto a generic datetime format by a pd.to_datetime call at the dataframe level.
<br> Also, for the sake of consistency, the _find_or_check_categorical_variables method from variable_manipulation now uses the datetime "smart" converters when checking for object/categorical variables, so that datetime variables (even when disguised as obj/cat) are not misinterpreted as actually categorical variables

baseclass features *not* implemented:
- methods to return time/date features. I am not too sure about what this one should do. Currently, a *variables_* attribute is available upon calling the fit method of the transformer, that returns the datetime variables from which the date features are going to be extracted by the transform method. How far is it from the original request?
-  methods to deal with the outliers. Haven't really put any work just yet on this one.

ExtractDateFeatures features implemented:
- month, quarter, semester, year, woty, wotm (naive, not actually taking the calendar into account), dotm, dotw, is_weekend feature extraction
- option to drop the datetime features from which date features were extracted, defaulting to True
- option to pick a subset of variables for the transformer to process, defaulting to None, i.e. the transformer checks the whole dataframe
- option to choose what features to extract among those implemented, defaulting to "year" (for lack of a better idea); also, setting features_to_extract to "all" extracts all the supported date features.

ExtractDateFeatures features *not* implemented:
- option to return a user-requested subset of variables. How is this different from the drop_datetime option? Do we want to make it so that the transformer can drop any feature, even non-datetime, when instructed to do so? 

miscellaneous/technical stuff:
- in order to test the new code, I needed to add a new fixture to the conftest file. Now, this fixture "df_vartypes2" creates a dataframe that is basically an extension of the previously existing "df_vartypes" with some datetime-like data. The reason I did this is because simply changing the former fixture would break pretty much all the tests in the package that are hard coded to work on the size and content of that specific dataframe. Is this good/ok practice, or should I do it differently?
- the _convert_variables_to_datetime and _convert_variable_to_datetime from variable_manipulation are functions that take and return converted copies of pd.df and pd.series respectively. How does this comply with the naming/scoping standards?
- On a side note, lint tests always cause a line swap in the outliers/init file that result in a circular import that breaks everything. I noticed that a [skip] kind of command is present in the tox.ini file, but that does not seem to be working. Any hints on that one?

PS: should I PR on feature_engine_examples with my notebook demo draft?

closes #234 
closes #111


feature_engine/variable_manipulation.py
```diff
@@ -127,6 +135,116 @@ def _find_or_check_categorical_variables(
...
+    variables : variable or list of variables. Defaults to None.
+
+    Raises
+    ------
+    ValueError
+        If there are no datetime variables in df or df is empty
+    TypeError
+        If any of the user provided variables are not datetime or
+        [failed to convert str/obj to datetime]?
+
```
solegalli, line 218: I would structure this function more or less like this:

For variables indicated by the user in a list:

if they are not type timestamp or object, return an error
for those that are object, attempt to convert to datetime, if fails, return an error

if user does not indicate which variables, then we should find them:
step 1: find those variables in the df that are of timestamp dtype, 
step 2: find those that are object that can be convertible to dt

feature_engine/parameter_checks.py
```diff
@@ -34,3 +34,52 @@ def _define_numerical_dict(dict_: Optional[dict]) -> Optional[dict]:
         raise TypeError("The parameter can only take a dictionary or None")
 
     return dict_
+
+
+def _parse_features_to_extract(feats_requested, feats_supported: List[str]) -> Any:
+    """
```
solegalli, line 40: in this script, we include functions that check the parameters entered in the __init__ method of the classes. Not sure if this is the intention of this function.

solegalli (commented):
Hi @dodoarg 

First of all, thank you so much for the enormous amount of work you put into developing this transformers. I really appreciate it. And users will too once it is ready and released :)

There is a lot going on here and a lot of code we need for these transformers, so I suggest, we take smaller steps into crafting this functions and classes.

I suggest starting for the following quick fixes:

1. remove all parts in docstrings with the heading "Raises". I explained why in one of my comments
2. remove the files that do not belong to this PR, I indicated which ones in comments

Next,  suggest we focus first on the variable to detect datetime variables. 

datetime variables can be in format timestamp I think or can be object. So if they are in timestamp format, then we accept. If they are object, then we try to convert to datetime, if success ok, if not error. I added some comments on the variable manipulation script with more details

Do you think we could simplify this function to do just that? 

After this, we adjust the tests, and I could have a look again, and then we move to a following bit of code.

Does this sound like a plan?

feature_engine/base_transformers.py
```diff
@@ -144,3 +148,131 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
     # for the check_estimator tests
     def _more_tags(self):
         return _return_tags()
+
+
+class DateTimeBaseTransformer(BaseEstimator, TransformerMixin):
```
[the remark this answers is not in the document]
dodoarg: Got it, sure

feature_engine/base_transformers.py
```diff
@@ -144,3 +148,131 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+    """shared set-up procedures across datetime transformers"""
+
+    def _select_variables_from_dict(
+        self, X: pd.DataFrame, user_dict_: Dict
+    ) -> pd.DataFrame:
+        """
+        Checks that input is a dataframe, checks that variables in the dictionary
+        entered by the user are of type datetime.
```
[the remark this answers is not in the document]
dodoarg: That method was a leftover from an initial copy paste I had done from the BaseNumericalTransformer() that contains it.<br> I can remove it from the BaseDateTimeTransformer as it isn't currently called anywhere by the child class, in fact ExtractDateFeatures() takes indeed a list of variables as argument

feature_engine/base_transformers.py
```diff
@@ -144,3 +148,131 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+
+        Parameters
+        ----------
+        X : Pandas DataFrame
+
+        user_dict_ : Dictionary. Default = None
+            Any dictionary allowed by the transformer and entered by user.
+
+        Raises
```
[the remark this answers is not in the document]
dodoarg: Got it, sure

feature_engine/outliers/__init__.py
```diff
@@ -6,4 +6,5 @@
 from .winsorizer import Winsorizer
 from .trimmer import OutlierTrimmer
 
+
```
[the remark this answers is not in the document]
dodoarg: Yea I was quite confused by the amount of files that were included in the PR (I'm a git noobie) <br> I think all those changes are a result of some lint/typecheck/stylecheck call, I never edited all those files myself. Then they somehow got staged together with the rest. I'll try and fix that

feature_engine/parameter_checks.py
dodoarg: That function does indeed check the argument 'features_to_extract' of the ExtractDateFeatures.__init__ method. <br> All those checks were originally within the class init method itself but I thought it better moved somewhere else, also because a future time transformer or something would probably be able to call that function too <br>
Is it fine or am I still missing something?

feature_engine/outliers/__init__.py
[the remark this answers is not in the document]
solegalli: Yes, I thought so. It doesn't matter. We can remove them. I tend to do `git add file1.py file2.py` instead of `git add .` to avoid committing unwanted files

feature_engine/variable_manipulation.py
```diff
@@ -110,15 +110,23 @@ def _find_or_check_categorical_variables(
         variables = [variables]
 
     elif not variables:
-        variables = list(X.select_dtypes(include=["O", "category"]).columns)
+        variables = list(
```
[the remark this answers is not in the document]
dodoarg: The reason is: that function detected datetime variables that were disguised as object as categorical variables. E.g. the column "data" in pd.DataFrame({"data" : ['10/03/21', '05/02/09']}) <br>
Now I believe that was fine up until we didn't have any datetime transformers, but now that datetime transformers are looking to convert such variables into datetime to perform downstream tasks such as feature extraction, I thought categorical encoders had better *not* work on those features, for the sake of consistency.

Otherwise, if say a categorical encoder was __init__'d with variables=None and a datetime transformer was also __init__'d with variables=None, both would work on those variables whose dtype is object but actually contain datetime information.

Hope I made myself clearer now, obviously it's a design choice and it's completely up to you to decide what to do with it

feature_engine/variable_manipulation.py
dodoarg: This part is a little more iffy, it is definitely the one that took me the longest to figure out. Unfortunately I think it still needs a bit more discussing

The way it currently works is more or less as you outlined above. However:
- Timestamp is the type of a __single__ datetime observation, not a column/series. After calling pd.to_datetime(), the series.dtype is _usually_ np.datetime64
- The reason why I say _usually_ is because when the data is timezone aware, the dtype is something slightly different and pandas has a _duck type_ for it called pd.DatetimeTZDtype. Also, there might be cases where datetimes object get cast to something else that I haven't come across.
- Therefore, instead of checking the _type_ of the datetime variables, I decided to check whether they have the __dt__ attribute, which happens to be the attribute used by the ExtractDateTransformer to extract the features afterwards. I believe all datetime specific dtypes have in fact the attribute dt; in addition, if some obscure datetime format ever shows up that doesn't have the dt attribute, it would be intended behaviour to reject it as the transformer wouldn't be able to work on such variables anyway.

To sum up on _find_or_check_datetime_variables:

if the user indicates the variables:

1. pass those variables to _convert_variables_to_datetime, which will attempt to convert non-numerical variables to datetime. The variables for which a conversion was not possible will be returned unchanged by passing errors='ignore' to pd.to_datetime() calls

2. if after the conversion some variables still do not have the dt attribute (which, I reiterate, is our argument to determine that they are in fact datetime variables that the downstream transformers can work with), throw a TypeError

if the user does not indicate the variables:

1. same as above, but pass the whole dataframe
2. if no variables have the dt attribute after the conversion, throw an error

In both cases, if no errors are thrown, return the variables that have the dt attribute

dodoarg: Sorry about the textwall but I couldn't really be more concise than this I don't think. I found this specific piece of the code quite cumbersome to go through. What do you think?

feature_engine/variable_manipulation.py
```diff
@@ -127,6 +135,116 @@ def _find_or_check_categorical_variables(
     return variables
 
 
+def _convert_variable_to_datetime(X: pd.Series, **kwargs) -> pd.Series:
+    """
```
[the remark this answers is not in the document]
dodoarg: Agreed on moving the functions. Any suggestions on where? <br>
More about the role of these functions in the comment section below

feature_engine/variable_manipulation.py
solegalli: did you consider pandas [s_datetime64_any_dtype](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.api.types.is_datetime64_any_dtype.html)? I found it in this [stackoverflow thread](https://stackoverflow.com/questions/43214204/how-do-i-tell-if-a-column-in-a-pandas-dataframe-is-of-type-datetime-how-do-i-te).

Not sure it would work but I would try something like:

`from pandas.api.types import is_datetime64_any_dtype as is_datetime`

If user does not indicate list:
```
vars_df = [column for column in df.columns if is_datetime(df[column])]]
vars_ob = [column for column in df.columns if is_datetimepd.to_dattime((df[column]))]
variables = vars_df + vars_ob
```

If user indicates list:
```
var_ = c for c in df.olumns if not is_datetime(df[column] or not is_datetimepd.to_dattime((df[column])
if len(var_)>0 ==> return error
```

feature_engine/variable_manipulation.py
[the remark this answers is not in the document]
solegalli: Not sure we need them. See my comment below

feature_engine/variable_manipulation.py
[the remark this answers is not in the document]
solegalli: Thanks for explaining. I think let's not mix functions and let's keep checking datetime to a different function.

feature_engine/base_transformers.py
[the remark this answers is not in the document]
solegalli: Let's wait with this change, because I think it might be better instead of having a base class and then 3 additional classes, to have just 1 class that does everything, and then regulate their behaviour with params.

feature_engine/parameter_checks.py
solegalli: I prefer having the checks in the init itself. It is more readable, than if we have to navigate several scripts. I would take this out of the init and make a function only if this function was to be used by more than 1 class.

See for example all the checks that we have in init params [here](https://github.com/feature-engine/feature_engine/blob/main/feature_engine/selection/drop_psi_features.py#L234)

dodoarg: Okay, fine, if anything we could always refactor those checks out of the __init__ in case a new transformer needs them too

feature_engine/variable_manipulation.py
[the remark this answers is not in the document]
dodoarg: Okay, although I believe this will result in requiring the user to be more diligent when initializing the transformer (i.e., passing reasonable variables), otherwise stuff like this could happen:

1. user calls a categorical encoder on a "datetime" variable (e.g. by passing variables=None, and _find_or_check_categorical_variables will consider it a categorical variable if its dtype is object), and the variable gets encoded and its dtype changed to some numerical type in the process.
2. user then calls a datetime transformer on the same variable (either by passing variables=None or specifying it), but since the previous call to the categorical transformer encoded that variable into a numerical feature, the datetime transformer will not be able to process it, thus the actual datetime information on that variable is lost

This sounds potentially dangerous to me, but if you're okay with it then I will revert any changes made on that function

feature_engine/variable_manipulation.py
solegalli: Thanks for raising this.

To start of, I would just consider object. I think this would capture the great majority of cases. And make our lives easier :)

feature_engine/variable_manipulation.py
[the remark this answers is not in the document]
solegalli: I see your point. Let me sleep on it

dodoarg:
Think I've done pretty much all the cleaning. Let me know if something is still out of place. This has been a great learning endeavor for me so far!

feature_engine/datetime/date.py
```diff
@@ -0,0 +1,169 @@
...
+    fit_transform:
+        Fit to the data, then transform it.
+    """
+
+    def __init__(
+        self,
+        variables: Union[None, int, str, List[Union[str, int]]] = None,
+        features_to_extract: Union[str, List[str]] = "year",
+        drop_datetime: bool = True,
+        **kwargs
```
solegalli, line 75: what would be the kwargs for?

In general, unless there is a very good reason, I would not allow the use of kwargs. As a user I almost never really know what a kwarg could be when i see them in the docs, lol

feature_engine/datetime/date.py
```diff
@@ -0,0 +1,169 @@
...
+    transform:
+        Add the new date features specified in features_to_extract argument
+    fit_transform:
+        Fit to the data, then transform it.
+    """
+
+    def __init__(
+        self,
+        variables: Union[None, int, str, List[Union[str, int]]] = None,
+        features_to_extract: Union[str, List[str]] = "year",
```
solegalli, line 73: I would default this to None, in which case we would extract all features. 

Alternatively, the user can only pass a list with the features to extract. Even if it is only 1 feature, we should allow only lists. Makes our lives easier, and for the user is not a massive headache to put the string in a list :p

feature_engine/datetime/date.py
```diff
@@ -0,0 +1,169 @@
...
+            "day_of_the_week",
+            "day_of_the_month",
+            "is_weekend",
+            "week_of_the_month",
+        ]
+
+        if not isinstance(features_to_extract, (str, list)):
+            raise TypeError("features_to_extract must be either a str or a list of str")
+
+        if features_to_extract == "all":
```
solegalli, line 94: as per scikit-learn conventions, we should leave the parameters to whatever values are entered by the user (otherwise, the check_estimator tests from sklearn will fail). This bit of code should go in the fit section, and it would look like this:

```
if self.features_to_extract is None:
   self.features_to_extract_ = supported
```

Note the trailing underscore, all parameters or attributes assigned during fit should have a trailing underscore as per sklearn conventions as well. More on this later

feature_engine/datetime/date.py
```diff
@@ -0,0 +1,169 @@
...
+            features_to_extract = self.supported
+
+        elif isinstance(features_to_extract, str):
```
solegalli, line 97: we would enforce lists, so this is not needed

feature_engine/datetime/date.py
```diff
@@ -0,0 +1,169 @@
...
+
+        if any(feature not in self.supported for feature in features_to_extract):
+            raise ValueError(
+                "At least one of the requested feature is not supported. "
+                "Supported features are {}.".format(", ".join(self.supported))
+            )
+
+        self.variables = _check_input_parameter_variables(variables)
+        self.drop_datetime = drop_datetime
+        self.kwargs = kwargs
```
solegalli, line 108: dangerous

feature_engine/datetime/date.py
```diff
@@ -0,0 +1,169 @@
...
+        self.features_to_extract = features_to_extract
+
+    def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
+        X = super().fit(X, y)
+        self.n_features_in_ = X.shape[1]
+        return self
+
+    def transform(self, X: pd.DataFrame) -> pd.DataFrame:
+
+        X = _is_dataframe(X)
```
solegalli, line 118: this is already in the base class

feature_engine/datetime/date.py
```diff
@@ -0,0 +1,169 @@
...
+        X = super().transform(X)
+
+        # maybe iterate the following with func handles or smth
+        # maybe rearrange final columns so that features extracted from
+        # the same variable are grouped together
+
```
solegalli, line 124: we should put this in a loop, so that features are derived for all datetime variables indicated by the user

`for var in self.variables`

feature_engine/datetime/date.py
```diff
@@ -0,0 +1,169 @@
...
+        if "month" in self.features_to_extract:
+            for var in self.variables_:
```
solegalli, line 126: I see you put the loop in, it generates boiler plate, if you take this line out of the loop, then you don't need to re-write it so many times. And it will also generate that the features from the same variable are one next to the other in the final df. Easier to understand for the user i think

solegalli (commented):
Hi @dodoarg 

Thank you so much for the quick turn-around with the changes.

Well done on the test to check the find_or_check_datetime_variables. Very thorough.

Thank you so much for the code in this PR, there is a lot already. This is looking really good.

I made a few suggestions throughout. Could you have a look at those? let me know if they make sense.

One question that I have not tested myself.

If we have a column object with values like this: '2016/09/09', and we want to derive time features, will this create the features hr:0, min:0 and s:0? or will it raise an error?

If we have a column object with values like this: '10:15:30', and we want to derive date features will pandas return values for year? month? day, etc? or will it return an error?

If it does not return an error, we may actually not need to separate the date and time into different classes and we could actually have them altogether in one class. It is much simpler.

Would you be able to test that and let me know what you find?

thank you!!

feature_engine/datetime/base_transformer.py
```diff
@@ -0,0 +1,95 @@
...
+        # check that input is a dataframe
+        X = _is_dataframe(X)
+
+        # Check if input data contains same number of columns as dataframe used to fit.
+        _check_input_matches_training_df(X, self.n_features_in_)
+
+        # check if dataset contains na
+        _check_contains_na(X, self.variables_)
+
+        X = pd.concat(
```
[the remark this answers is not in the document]
dodoarg: 1. the (super) fit method calls _find_or_check_datetime_variables, that returns the list of datetime variables for the transformer to work with
2. the (super) transform method actually does the __conversion__ of such variables to datetime, leaving the others (i.e., those not in self.variables_) unchanged.
3. the (child) transform method extract features via the .dt attribute, now available on the self.variables_ that have been converted

As for the actual way that is coded, since the conversion needs to be done at the _series_ level, I thought the easiest way to do it was to use the concat method on a series list comprehension

feature_engine/datetime/date.py
dodoarg: You want it to default to all features? When I was coding I had this design choice to make and I thought 7+ columns to add to the dataset __per variable__ would be a lot, that's why I thought of the two options "all" and None where None is default and returns one or a few (most relevant) features.
But I see your point, I guess it makes the code easier

feature_engine/datetime/date.py
dodoarg: The kwargs were meant to be propagated up to the pd.to_datetime call. The thing is, when parsing datetime data there's a couple things to consider, most importantly, the format - e.g. d/m/y or y/m/d - and pd.to_datetime has options to deal with that. <br> Currently, the transformer will always return extracted features from datetime that was converted with the default options, and the user won't have a say in that

feature_engine/datetime/date.py
dodoarg: Yea no this was terrible, I don't know what I was thinking

feature_engine/variable_manipulation.py
```diff
@@ -107,6 +109,7 @@ def _find_or_check_categorical_variables(
     """
 
     if isinstance(variables, (str, int)):
+        print(variables)
```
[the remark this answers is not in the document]
dodoarg: oops

feature_engine/variable_manipulation.py
```diff
@@ -127,6 +130,64 @@ def _find_or_check_categorical_variables(
...
+
+        if len(variables) == 0:
+            raise ValueError(
+                "No datetime variables in this dataframe. "
+                "Note: purely numeric variables representing dates or times "
+                "will not be treated as datetime by this transformer."
+            )
+        return variables
+
+    if isinstance(variables, (str, int)):
```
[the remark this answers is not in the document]
dodoarg: Both _find_or_check_numerical_variables and _find_or_check_categorical_variables actually [do](https://github.com/feature-engine/feature_engine/blob/main/feature_engine/variable_manipulation.py#L62) that check.
In fact, I think this along with the logic of the functions - if, elif, else structure - is a problem. When either of those _find_or_check* functions is passed a string or an int, it will
1. make a list that contains it
2. return that list
__without__ actually entering the checks! This has probably gone unnoticed because there are no tests where a string/int corresponding to a non-numerical or non-categorical variable is passed to its _find_or_check counterpart

tests/conftest.py
```diff
@@ -18,6 +18,23 @@ def df_vartypes():
     return df
 
 
+@pytest.fixture(scope="module")
```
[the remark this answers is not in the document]
dodoarg: Sounds good!

tests/test_variable_manipulation.py
```diff
@@ -78,8 +80,69 @@ def test_find_or_check_categorical_variables(df_vartypes, df_numeric_columns):
...
+            df_vartypes2.join(tz_date),
+            variables=None,
+        )
+        == vars_convertible_to_dt + ["dofZ"]
+    )
+
+
 def test_find_all_variables(df_vartypes):
-    all_vars = ["Name", "City", "Age", "Marks", "dob"]
+    all_vars = list(df_vartypes.columns)
```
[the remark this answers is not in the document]
dodoarg: Yea no this was very much needed. I haven't actually changed what does tests do, that's just a renaming, but it was unnecessary

feature_engine/variable_manipulation.py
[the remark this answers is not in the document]
solegalli: Actually you are right. We do need that there. 

I forgot I changed the logic of `_check_input_parameter_variables` when I added the checks for the check_estimators from sklearn.

Apologies :_(

tests/conftest.py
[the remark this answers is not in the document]
solegalli: You are right again, then we leave the df there. Thank you for being in top of things :)

feature_engine/datetime/date.py
solegalli: can we instead specifically pass the pandas dt parameters in the init, those that we would need? like coerce and the string format? or are they too many parameters? sorry I ask, I don't remember of the top of my head.

[here](https://github.com/feature-engine/feature_engine/blob/main/feature_engine/selection/drop_correlated_features.py#L89) for example the param method is the param taken by pandas corr.

feature_engine/datetime/date.py
solegalli: The thing is that the most relevant will change with the use case. If it is something recorded every minute, then year does not make much sense. And if it is something recorded monthly, then daily would not be appropriately. I know, the feature space may explode. But we hope the user is on top of what they need and can narrow down the selection? not sure.

dodoarg: Yea that's why I wasn't sure about what to have it default it to. Maybe we make it a positional argument of the __init__ so that the user __has to__ specify what features to extract?

Especially if we add all those binary attributes you mentioned in another comment like is_month_start etc the feature_space is really gonna explode with something like 15+ new columns per variable

feature_engine/datetime/date.py
dodoarg: That could work. To keep things simple, I think we can get away with dayfirst and yearfirst only ([see here](https://pandas.pydata.org/docs/reference/api/pandas.to_datetime.html?highlight=pd%20to_datetime#:~:text=return%20the%20input.-,dayfirstbool%2C%20default%20False,first%20(this%20is%20a%20known%20bug%2C%20based%20on%20dateutil%20behavior).,-utcbool%2C%20default%20None)) which are obviously clutch to extract the features correctly

feature_engine/variable_manipulation.py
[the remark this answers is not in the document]
dodoarg: No problem, I'm still not too sure what to do though, I would be inclined to change the logic inside _find_or_check_numerical_variables and _find_or_check_categorical_variables aswell. Please have a look at the demo below (on the PowerTransformer, which I haven't touched in my fork) <br>
Here, _variables_ is passed as a list and the correct TypeError is thrown at the _find_or_check_numerical_variables level:
```
>>> pt = PowerTransformer(variables=["City"])
>>> pt.fit(df)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "D:\feature_engine\feature_engine\transformation\power.py", line 82, in fit
    X = super().fit(X)
  File "D:\feature_engine\feature_engine\base_transformers.py", line 98, in fit
  File "D:\feature_engine\feature_engine\variable_manipulation.py", line 82, in _find_or_check_numerical_variables
    raise TypeError(
TypeError: Some of the variables are not numerical. Please cast them as numerical before using this transformer
```
Here instead variables is passed as a str (which _check_input_parameter_variables __allows__) but _find_or_check_numerical_variables doesn't actually run any checks, so the logic goes on up until the line where the column is checked for the presence of inf, which fails because City is an object variable, and should not have made it that far in the code
```
>>> pt = PowerTransformer(variables="City")   
>>> pt.fit(df)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "D:\feature_engine\feature_engine\transformation\power.py", line 82, in fit
    X = super().fit(X)
  File "D:\feature_engine\feature_engine\base_transformers.py", line 102, in fit
  File "D:\feature_engine\feature_engine\dataframe_checks.py", line 127, in _check_contains_inf
    if np.isinf(X[variables]).values.any():
  File "D:\Vari\Anaconda3\envs\ML\lib\site-packages\pandas\core\generic.py", line 1936, in __array_ufunc__
    return arraylike.array_ufunc(self, ufunc, method, *inputs, **kwargs)
  File "D:\Vari\Anaconda3\envs\ML\lib\site-packages\pandas\core\arraylike.py", line 366, in array_ufunc
    result = mgr.apply(getattr(ufunc, method))
  File "D:\Vari\Anaconda3\envs\ML\lib\site-packages\pandas\core\internals\managers.py", line 425, in apply
    applied = b.apply(f, **kwargs)
  File "D:\Vari\Anaconda3\envs\ML\lib\site-packages\pandas\core\internals\blocks.py", line 378, in apply
    result = func(self.values, **kwargs)
TypeError: ufunc 'isinf' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
>>>
```
Sorry If I insist of this one but I really need to know what to do to proceed because this issue clearly applies to _find_or_check_datetime_variables aswell

tests/test_datetime/test_extract_date_features.py
```diff
@@ -43,13 +43,19 @@ def test_extract_date_features(df_vartypes2):
             }
         )
     )
+    dates_nan = pd.DataFrame({"dates_na": ["Feb-2010", np.nan, "Jun-1922", np.nan]})
 
     # check exceptions upon class instantiation
```
dodoarg, line 55: In case you're wondering what's the big fuss about it well tests were thorough and that but I came across a huge error i.e. I had more than one error assertion per pytest.raises with block so basically only the first one per block was being checked.

feature_engine/datetime/date.py
```diff
@@ -84,21 +84,16 @@ def __init__(
...
-
-        elif isinstance(features_to_extract, str):
-            features_to_extract = [features_to_extract]
-
-        if any(feature not in self.supported for feature in features_to_extract):
-            raise ValueError(
-                "At least one of the requested feature is not supported. "
-                "Supported features are {}.".format(", ".join(self.supported))
-            )
+
```
dodoarg: couldn't really think of a non-nested way to do it

feature_engine/datetime/date.py
```diff
@@ -111,57 +106,40 @@ def __init__(
     def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
         X = super().fit(X, y)
         self.n_features_in_ = X.shape[1]
```
dodoarg, line 126: hopefully this complies with the standards you mentioned?

dodoarg:
As for your question, yes you can extract dates from time-only variables and times from date-only variables.

```
>>> df
       date      time
0  10/11/12  10:15:30
1  09/09/09  05:27:54
2  05/25/95  22:22:32
3  03/17/04  14:09:45
>>> df.apply(pd.to_datetime)
        date                time
0 2012-10-11 2021-12-03 10:15:30
1 2009-09-09 2021-12-03 05:27:54
2 1995-05-25 2021-12-03 22:22:32
3 2004-03-17 2021-12-03 14:09:45
>>> df.apply(pd.to_datetime).date.dt.hour
0    0
1    0
2    0
3    0
Name: date, dtype: int64
>>> df.apply(pd.to_datetime).time.dt.isocalendar()
   year  week  day
0  2021    48    5
1  2021    48    5
2  2021    48    5
3  2021    48    5
>>>
```

time attributes are defaulted to zeros, date attributes are defaulted to today's date

feature_engine/variable_manipulation.py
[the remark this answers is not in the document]
solegalli: I see your point here, and I think it makes sense to ignore datetime variables which type is object. But I think we should do that in a different PR. 

I created a new issue #336

feature_engine/variable_manipulation.py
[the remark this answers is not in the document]
solegalli: Well spotted. It is indeed a bug in both the `_find_or_check_numerical_variables` and the _find_or_check_caterical_variables`. 

If the user passes a string instead of a list, we are not currently checking that the variable is of the corresponding type. 

I create a separate issue to address this bug:

feature_engine/datetime/date.py
solegalli: Scikit-learn encourages us to have all parameters with default values, so that users can implement it out of the box. So if we can, let's stick to that.

The list of variables to get from time is quite extensive, I agree with you on the explosion of the feature space, so maybe, instead of outputting just one variable like "year", or all variables, the default could be to output the group that we consider the most common.

Maybe we can return by default year, month, day (1-31) and day of the week (0-6)?

So the parameter would take 3 types of input:

- `None`: returns the above 4 features
-` "all"`: returns all features
- `list:` with short names to the features the user wants returned.

What do you think?

solegalli:
> attributes are defaulted to zeros, date attribu

Thank you for checking! This is great :)

In this case, let's create 1 class only called DateTimeFeatures() instead of the 3 we had originally planned.

Would you agree?

solegalli:
Hi @dodoarg 

I went through the code changes you've done since the last time. It is looking really good. 

I also tried to answer some questions. 

Let me know if you are waiting to hear from me on something that I missed, or when you would like me to have a look at the code again.

Thanks a lot!

dodoarg:
Hi @solegalli, merging your previous comments onto a single topic

> In this case, let's create 1 class only called DateTimeFeatures() instead of the 3 we had originally planned.
> So the parameter would take 3 types of input:

> we can return by default year, month, day (1-31) and day of the week (0-6)?
> ...
> None: returns the above 4 features
> "all": returns all features
> list: with short names to the features the user wants returned.

A few comments/ideas, let me know what you think

- I think if we want to make a date __and__ time feature extractor we want the default option _None_ to return something related to time aswell, maybe hour, minute, second.
- Although I had initially coded the _all_ option myself, I really can't see a situation where every single one of those features are useful, there is simply too many of them and most of them, especially the binary ones that I recently added, are extremely situational I think. So to thin out the code I would suggest to get rid of the _all_ option in the first place. <br>
- Since time-only features are parsed as _today's date_ and date-only features are parsed as _00:00:00_: <br> let's say features_to_extract=None defaults to returning month, day, year, hour, minute, second. Or maybe you disagree with getting rid of the _all_ option so the _all_ option returns all the supported date __and__ time features. In this case, if one or more datetime features are time-only or date-only, there will be potentially several features that are effectively garbage and do not add the slightest information to the dataset. Therefore, I was thinking maybe we could call DropConstantFeatures before returning the transformed df.

solegalli:
Ok to everything :)

Good ideas @dodoarg

dodoarg:
A couple comments on these last commits @solegalli :
for starters, sorry about the duplicate commits, I slightly messed up the branch history as I thought a merge with a temporary local branch wouldn't be kept track of remotely. Looks like I was wrong

Speaking of commits, up to commit 211f796 it should be alright as I just did what we had discussed. The last two might be controversial in the way I decided to rework the code, so I will briefly explain:

- as I was adding more and more features to the (now datetime) transformer, I felt like the class was starting to be quite overcrowded, mostly due to string parsing and repeated if logic blocks
- to implement the option to drop constant features, since I didn't want to drop features that were __not__ extracted by the transformer, I needed to access the names of said extracted features to pass them to DropConstantFeatures. That would have added even more strings/boilerplate code

Thus, I decided to thin out the class (as well as the test file) by importing the constant stuff related to the features from a separate file that I put in the module.  I think this also has the positive side effect that in order to change anything in what features are supported, what are the default ones, how to compute them etc. one just needs to edit the correspondent constant variable rather than touch the code inside the class.
Let me know what you think, if it's unwanted I can just revert to the previous state

solegalli:
Hi @dodoarg 

Thank you so much for such an amazing contribution. I particularly enjoyed the code of the datetime feature extraction class. Very tidy, very readable. Loved it.

I made a [PR to your repo](https://github.com/dodoarg/feature_engine/pull/1) with some code changes. Would you mind having a look and merging? That will update automatically the code in this PR.

This is what I've done:

**Big changes**

- removed drop constant functionality: the reason is that is non-deterministic. In other words, it will not output dataframes with the same number of features. Extreme example: if I transform a dataframe with only 2 rows that have the same date, all features will be constant and dropped. We don't want that, we want the output of the transform method to be a dataframe with the same number of features no matter what X is passed to the transform.
- I reformatted the check datetime variable, it should now accept categorical (I know we said not to at the beginning, but I can't remember why). I thought we would just move forward with this.

**Minor changes**
- I added the functionality to return all features when passing "all" as param to features to extract. It is just a few extra lines of code, so i think it does not hurt.
- I renamed the class, the script, the tests and one of the parameters for consistancy on how we call the same functionality in other classes

Main things to do:

- [ ] break down the tests. At the moment, we have only 1 test that tests everything. We should break this down into smaller tests, with a name that allows us to identify what is being tested. See for example [here](https://github.com/feature-engine/feature_engine/blob/main/tests/test_imputation/test_categorical_imputer.py). This makes it much easier to troubleshoot, identify what is failing and helps us maintain looking forward
- [ ] add test that class works as expected when df has variables of type categorical
- [ ] add test that class works as expected when var in different timezones

solegalli:
on the drop constant, the user can easily do that by implementing this:

```
extract_dt = Pipeline([
  ('dt_features', DateTimeFeatures)(),
  ('drop_constatn', DropConstantFeatures()),
  ])

extract_df.fit(X_train)

Xtrain_t = extract_dt.transform(X_train)
Xtest_t = extract_dt.transform(X_test)
```

I think we should add this in the documentation as an example to guide the users, instead of complicating our lives with trying to introduce this logic here.

Thoughts?

dodoarg:
Thank you for your support.
I merged and (I believe) fixed your recent changes. Now I can crack on with the test tweaks you've suggested.

dodoarg:
About dropping constant variables, I've thought it over and I agree with you to a certain extent.

let's say we have a column 
date_var : ['Dec-2010', 'Feb-2010', 'Mar-2010']
and the user calls DatetimeFeatures without specifying features_to_extract (which currently defaults to month, year, day, hour, minute, second)
There's two fundamentally different types of constant variables that get generated by the DatetimeFeatures transformer:

- constant features that are extracted from inherently constant data - date_var_year : [2010, 2010, 2010]
- constant features that are extracted from variables that may or may not be constant that are generated as a byproduct of the pd.to_datetime parsing - date_var_hour: [0, 0, 0]

While I agree with you that we might not want the transformer to get rid of the former kind of constant features, the latter kind of constant features are effectively filling the dataset with made up information, on top of being clearly useless, and I can't see a user ever intending to extract a column (or multiple columns) arbitrarily filled with today's date or midnight o'clock time.

I don't necessarily think that having a call to DropConstantFeatures - perhaps limited to the "fake" features - would be particularly problematic on the technical side, but then again the easiest way would be to ask the user to be more conscious about what they're doing and it's ultimately up to you

solegalli:
> About dropping constant variables, I've thought it over and I agree with you to a certain extent.
> 
> let's say we have a column date_var : ['Dec-2010', 'Feb-2010', 'Mar-2010'] and the user calls DatetimeFeatures without specifying features_to_extract (which currently defaults to month, year, day, hour, minute, second) There's two fundamentally different types of constant variables that get generated by the DatetimeFeatures transformer:
> 
>     * constant features that are extracted from inherently constant data - date_var_year : [2010, 2010, 2010]
> 
>     * constant features that are extracted from variables that may or may not be constant that are generated as a byproduct of the pd.to_datetime parsing - date_var_hour: [0, 0, 0]
> 
> 
> While I agree with you that we might not want the transformer to get rid of the former kind of constant features, the latter are effectively filling the dataset with made up data and I can't see a user ever intending to have a column (or multiple columns) suddenly filled with today's date or midnight o'clock time.
> 
> I don't necessarily think that having a call to DropConstantFeatures - perhaps limited to the "fake" features - would be particularly problematic on the technical side, but then again the easiest way would be to ask the user to be more conscious about what they're doing and it's ultimately up to you

I see the problem. Don't get me wrong. But there are a couple of constrains:

- the output of transform should have always the same number of columns.
- everything that needs to be learned, needs to be learned during fit.

So the only way we could implement the drop constant is to:

1. extract the features during fit.
2. find the constant features in the train set (during fit)
3. in transform always remove those features that were found constant in the train set

That will ensure that the output of transform contains always the same number of columns.

It is doable. It involves adding a more logic to our class. And it is so easily accomplished by tagging the dropconstantvariables in a pipeline at the back of this transformer, that it feels a bit like we are making our lives difficult when the functionality is already there.

I would be inclined not to add this functionality to the class. Explain how it could be done in the documentation, and see if users are happy with this, or if we should add the functionality to the class, which we could do later.

What do you think?

dodoarg:
Yea I see. For consistency, it would be best not to have anything learned outside of the _fit_ scope, while my point precisely addresses something that comes up at the _transform_ stage, with the extracted features. We might aswell leave it like that then and rely on the docs and the awareness of users

solegalli:
Is this ready for review?

dodoarg:
I revamped the datetime testing with several functions and a datetime scoped conftest but I still wanted to see if I couldn't throw in some pyest.parametrize decorators

dodoarg:
Small update on last commit: I found out that pd.to_datetime won't convert a series to datetime if different observations are in different timezones (it will rather leave it as object, preventing us from accessing dt attributes), so I decided to pass __utc=True__ everytime pd.to_datetime is called. This standardizes all times to utc and makes sure everything works properly. Let me know if you agree

About the parametrized tests, I couldn't really see a way to use it extensively in this case :/ I could use it for the attribute checks or some verbose error tests but I don't think it would look much better

solegalli:
> 

What about making utc a parameter in the init instead and allowing the user to decide? It would default to None, just like in pandas datetime, and if user knows that they have time aware zones, they can switch it to true.

solegalli:
After some research, it seems that the only way to have different tz in the same variable is by the string with the plus or minus amount of hrs from utc.

```
data = {
    "Name": ["tom", "nick", "krish", "jack"],
    "Age": [20, 21, 19, 18],
    "datetime_range": pd.date_range("2020-02-24", periods=4, freq="D"),
    "date_obj1": ["01-Jan-2010", "24-Feb-1945", "14-Jun-2100", "17-May-1999"],
    "date_obj2": ["10/11/12", "12/31/09", "06/30/95", "03/17/04"],
    "time_obj": ["21:45:23", "09:15:33", "12:34:59", "03:27:02"],
    "tz_obj": ['12:00:00+0100', '12:00:00+0200',
               '12:00:00-0100', '12:00:00+0000'],
}

df = pd.DataFrame(data)
```

```
a = pd.concat([
    pd.Series(
        pd.date_range(
            start='2014-08-01 09:00', freq='H', periods=3,
            tz='Europe/Berlin')),
    pd.Series(
        pd.date_range(
            start='2014-08-01 09:00', freq='H', periods=3, tz='US/Central'))
    ], axis=0).reset_index(drop=True)

df['tz'] = a

df
```

```
   Name  Age datetime_range    date_obj1 date_obj2  time_obj         tz_obj  \
0    tom   20     2020-02-24  01-Jan-2010  10/11/12  21:45:23  12:00:00+0100   
1   nick   21     2020-02-25  24-Feb-1945  12/31/09  09:15:33  12:00:00+0200   
2  krish   19     2020-02-26  14-Jun-2100  06/30/95  12:34:59  12:00:00-0100   
3   jack   18     2020-02-27  17-May-1999  03/17/04  03:27:02  12:00:00+0000   

                          tz  
0  2014-08-01 09:00:00+02:00  
1  2014-08-01 10:00:00+02:00  
2  2014-08-01 11:00:00+02:00  
3  2014-08-01 09:00:00-05:00  
```

We could add another parameter in the init, tz_aware=False. And when user sets to true, we could extract the tz in another feature like this:
```

df['tz'].astype(str).str[-6:]

0    +02:00
1    +02:00
2    +02:00
3    -05:00
```

Not sure how much value it adds. It might be an overkill.

Thoughts?

dodoarg:
>After some research, it seems that the only way to have different tz in the same variable is by the string with the plus or minus amount of hrs from utc.

Indeed, if you look at my demo above and the tests I added with the last commit, that's how I've created tz-aware variables

> What about making utc a parameter in the init instead and allowing the user to decide? 

The issue with your approach is that those variables are not datetime and __cannot__ be converted into datetime:
```
>>> a = pd.concat([
    pd.Series(
        pd.date_range(
            start='2014-08-01 09:00', freq='H', periods=3,
            tz='Europe/Berlin')),
    pd.Series(
        pd.date_range(
            start='2014-08-01 09:00', freq='H', periods=3, tz='US/Central'))
    ], axis=0).reset_index(drop=True)

>>> a.dtype
dtype('O')

>>> pd.to_datetime(a)
Traceback (most recent call last):
  File "D:\Vari\Anaconda3\envs\ML\lib\site-packages\pandas\core\arrays\datetimes.py", line 2085, in objects_to_datetime64ns
    values, tz_parsed = conversion.datetime_to_datetime64(data)
  File "pandas\_libs\tslibs\conversion.pyx", line 335, in pandas._libs.tslibs.conversion.datetime_to_datetime64
ValueError: Array must be all same time zone

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "D:\Vari\Anaconda3\envs\ML\lib\site-packages\pandas\core\tools\datetimes.py", line 805, in to_datetime
    values = convert_listlike(arg._values, format)
  File "D:\Vari\Anaconda3\envs\ML\lib\site-packages\pandas\core\tools\datetimes.py", line 465, in _convert_listlike_datetimes
    result, tz_parsed = objects_to_datetime64ns(
  File "D:\Vari\Anaconda3\envs\ML\lib\site-packages\pandas\core\arrays\datetimes.py", line 2090, in objects_to_datetime64ns
    raise e
  File "D:\Vari\Anaconda3\envs\ML\lib\site-packages\pandas\core\arrays\datetimes.py", line 2075, in objects_to_datetime64ns
    result, tz_parsed = tslib.array_to_datetime(
  File "pandas\_libs\tslib.pyx", line 364, in pandas._libs.tslib.array_to_datetime
  File "pandas\_libs\tslib.pyx", line 447, in pandas._libs.tslib.array_to_datetime
ValueError: Tz-aware datetime.datetime cannot be converted to datetime64 unless utc=True
```
Note: the same error is thrown even when errors="ignore" in pd.to_datetime (because it's not a parsing error, it's a value error)

dodoarg:
To add to that, I'm not sure I got how the datetime would be parsed in case the user says tz_aware=False
Like, what would be the non-tz-aware _hour_ here?
0  2014-08-01 09:00:00+02:00  
1  2014-08-01 10:00:00+02:00  
2  2014-08-01 11:00:00+02:00  
3  2014-08-01 09:00:00-05:00 

[09, 10, 11, 09] just ignoring the offset from utc time? 
I believe we'd need a workaround like a parser that gets rid of whatever comes after the + or - before calling pd.to_datetime

solegalli:
If time_aware=true, then utc=True. We would first capture the +06:00 in a variable. That should do it. I think.

If time_aware=False and the variable has +06:00, then pandas will raise an error. We would go with that. The user should know if their variables are time aware or not.

But still, I am not convinced we need this. Maybe we leave it for now?

I would just make utc a param in the init.

dodoarg:
> The user should know if their variables are time aware or not.

Oh well, if you put it that way...I guess I keep underestimating feature engine's users :p

However, I would keep utc fixed to true in the pd.to_datetime calls that are inside _find_or_check_datetime_variables so that even when time_aware=false and the variable has +6:00 it's still recognized as datetime and the error is thrown only when it reaches the transform method. Does that sound right?

dodoarg:
I've adjusted the code @solegalli the only thing I'm not too sure I like is how the errors are handled as far as the time zone goes. This is not a huge problem but I would still like to hear your opinion on this as I'm trying to work things out as thoroughly as possible

Say you have the object-like variable

```
>>> tz
0    21:24:00+3
1    17:18:00-5
dtype: object 
```
If the user (erroneously) calls DatetimeFeatures(__time_aware=False__) or DatetimeFeatures(__time_aware=None__) the variable is correctly identified as datetime but when it gets to the transform you get this
```
>>> pd.to_datetime(tz, utc=None)
0    2021-12-10 21:24:00+03:00
1    2021-12-10 17:18:00-05:00
dtype: object
```
which does not raise an error but instead returns a variable that __still has dtype=object__.
This means that an __AttributeError__ is raised when the transformer tries to access the _dt_ method afterwards
`AttributeError: Can only use .dt accessor with datetimelike values`

Why does pd.to_datetime not return an error in the first place? Only calling the function __again__ on the series results in a value error:
```
>>> pd.to_datetime(pd.to_datetime(tz, utc=None))
...long traceback...
ValueError: Tz-aware datetime.datetime cannot be converted to datetime64 unless utc=True
```


This is is because the first call is actually converting __the single items__ from strings to datetime objects while leaving the series dtype unchanged
```
>>> pd.to_datetime(tz, utc=None)[0]
datetime.datetime(2021, 12, 10, 21, 24, tzinfo=tzoffset(None, 10800))
``` 

while the second call then tries to convert the series dtype and it throws a ValueError because the items are in different timezones (it cannot assign a unique _tzoffset_ value to the whole series)


What is the big deal then? 
Well, as things stand there is no second call to pd.to_datetime before extracting the features via _dt_, so the user __gets an attribute error rather than a more informative value error__ and might not understand what is going on

tests/test_datetime/conftest.py
```diff
@@ -0,0 +1,95 @@
...
+        "datetime_range_month": [2, 2, 2, 2],
+        "datetime_range_quarter": [1, 1, 1, 1],
+        "datetime_range_semester": [1, 1, 1, 1],
+        "datetime_range_year": [2020, 2020, 2020, 2020],
+        "datetime_range_wotm": [4, 4, 4, 4],
+        "datetime_range_woty": [9, 9, 9, 9],
+        "datetime_range_dotw": [0, 1, 2, 3],
+        "datetime_range_dotm": [24, 25, 26, 27],
+        "datetime_range_doty": [55, 56, 57, 58],
+        "datetime_range_weekend": [False, False, False, False],
```
solegalli, line 18: I am noticing that the default return for some of these datetime features is True and False. 

However, the idea of feature-engine's transformers, whenever possible, is to return numerical features that are ready ti use with sklearn estimators.

Could we add a workaround to transform these booleans into [0,1] binary vectors?

solegalli (commented):
Hi @dodoarg 

First of all, apologies for the delay. I had to finish off some work till today :/

Second, it's been really inspiring working with you. You are very thorough and you have amazing attention to detail. Thank you very much.

Third, well done on rebasing the main branch :)
This is something that many often overlook.

Fourth, I made a PR to your repo with some refactoring of the tests and renaming of the params and typehints. Unfort, mypy is failing and I did not have time to figure out why.

I added a suggestions somewhere down in a comment to work on the issue you raised with time aware zones. **We could add an error to alert users to set utc to true.**

And I made another comment regarding the default true, false values of pandas when extracting features. Could you have a look at those? **I would prefer if it returned [0, 1] values.**

Also, I added the files where we need to include documentation for the new class, which is what will be visible to users. **Could you have a look and see if you could add description / examples?**

I need to do something else now, but will be around later today and tomorrow. Thank you!

feature_engine/datetime/datetime.py
```diff
@@ -159,7 +159,7 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
 
         if self.features_to_extract is None:
             self.features_to_extract_ = FEATURES_DEFAULT
-        elif self.features_to_extract == "all":
+        elif isinstance(self.features_to_extract, str):
```
dodoarg, line 162: Based on [this](https://stackoverflow.com/questions/51656300/mypy-falsely-reports-error-on-union-typed-variable-when-branching-on-type)
It should be harmless since we reject any string that isn't equal to _all_ in a previous check

feature_engine/datetime/datetime.py
```diff
@@ -0,0 +1,203 @@
...
+            [
+                pd.to_datetime(
+                    X[variable], dayfirst=self.dayfirst,
+                    yearfirst=self.yearfirst, utc=self.time_aware
+                )
+                for variable in self.variables_
+            ],
+            axis=1,
+        )
+
```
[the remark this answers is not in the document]
dodoarg: Great idea actually, also very easily implemented

dodoarg:
Hi @solegalli, thank you for your kind words and well done on the tests refactoring. The only thing I might consider worth double-checking is the time_zone related tests as I think there might be some redundancy considering the test you added jointly with the one I had written already (the one with the different timezones). Maybe we want to merge them into a single comprehensive test?

I think I've fixed the three code issues you mentioned.

I've also added some explanations and a basic example to the user guide. Since I'm not very experienced with building documentations etc would you mind giving me some heads up on what needs more attention?
On that note, I've also had an example notebook ready for quite some time, should I PR on feature-engine-examples?

solegalli (commented):
Hi @dodoarg 

Thanks for the quick turnaround with the documentation! And for fixing my mypy issue :)

I made a [PR](https://github.com/dodoarg/feature_engine/pull/4) reorganising bits and pieces of the docs.

The main part remaining is the user_guide. I added guidelines within the file itself.

**In answer to your questions:**
Yes, please make a PR with the jupyter notebook, and once it is merged, we link it to the user guide. Please make sure that you are using a dataset that has a license that explicitly says that we can use the dataset, and add instructions showing users how they can download it.

Note: if it is from Kaggle, most likely, it does not have a license. You can try the UCI machine learning repository instead.

**Redundant tests on timezones:** I don't think they are redundant, your test is for object like variables (right?), the one I added is for datetime  localized variables. We should keep both.

Also, can you extend the error to specify which variables could not be parsed to datetime? and we should probably add a test to catch that error as well. Sorry :/

[Here](https://github.com/feature-engine/feature_engine/blob/encoders_nan/tests/test_encoding/test_count_frequency_encoder.py#L120) an example on how to test an error message.

feature_engine/datetime/datetime.py
```diff
@@ -0,0 +1,215 @@
...
+                    yearfirst=self.yearfirst, utc=self.utc
+                )
+                for variable in self.variables_
+            ],
+            axis=1,
+        )
+
+        if any(not is_datetime(datetime_df[var]) for var in datetime_df.columns):
+            raise ValueError(
+                "ValueError: variables x, y and z could not be converted to datetime. "
```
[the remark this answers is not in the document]
dodoarg: lol that was not meant to be committed. I literally copy pasted the msg from your comment above and then forgot to change it. Serves me right for not testing the msg content!

dodoarg:
@solegalli Here's some examples, let me know if we're going in the right direction now 🤔

solegalli:
> @solegalli Here's some examples, let me know if we're going in the right direction now thinking

Masterpiece. I added a few more examples and made a PR to your repo.

dodoarg:
> Finally, we need to add the tests from check_estimator from sklearn. 

We still need this right?

solegalli:
> y ideas on what's going

The check_estimator from sklearn creates a couple of numpy arrays to test the basic functionality of the transformers. To fullfill sklearn conventions, the transformers need to do a bunch of things, like be able to fit anew everytime you call fit, it tests that you do not overwrite the init params, that all init params have default values, that the transformer is compatible with the pipeline and the grid searches, that it has get params and set params, and a few other things.

All numerical transformers of sklearn pass most of the tests. For the categorical transformers, I needed to include that flag to ignore format, which on one hand allows the transformers to work with numerical variables as if they were categorical, which was requested by some users, and on the other hand, the transformers can now be checked by sklearn checker.

I also tried the check_estimator in my last PR and realised quickly that we will need a massive workaround to be able to test it. Because even if we add a flag to ignore format, whenever it hits the transform method it will raise an error when it is not able to get date features from the variables.

So I am inclined to leave it for now. We can create an issue to think about it and come back to it in the future.

So from my perspective, the only thing remaining would be to remove that week feature from the functionality and also the docs demo (there is one demo where I set the features_to_extract to all, we need to delete a couple of columns from the output.

dodoarg:
Here's the thing, pd.datetime actually does work on numerical variables

```
>>> pd.to_datetime(pd.Series([1,2,3,4]))
0   1970-01-01 00:00:00.000000001
1   1970-01-01 00:00:00.000000002
2   1970-01-01 00:00:00.000000003
3   1970-01-01 00:00:00.000000004
dtype: datetime64[ns]
```

which is exactly we we need to have datetime checkers in the actual functionality otherwise whenever variables=None one would end up with a bunch of meaningless features. But in order to make those tests pass, maybe...this would actually be helpful?

Anyway I removed that feature @solegalli 
On the failing type tests: I believe it's got something to do with [this](https://github.com/python/mypy/commit/22ab84acc3f6174905a05ddadab94b6c2d83f9a1) recent update on mypy. The affected file is not part of this PR and it also breaks the test on the main branch, is it worth a quick separate issue before merging this feature branch?

solegalli:
Hi @dodoarg 

You are a source of knowledge!!!!

I asked a friend of mine who is an experienced python developer, if there is a way to add a hidden parameter in the init, say to ignore variable format, that is visible to us developers but invisible to users. This way, we could use this parameter to test the transformer. As soon as I hear from him, I'll ping an answer here. 

If you have any ideas on how we could implement the above, let me know as well. I wonder if we could create a base class with that init param, that could be inherited by this transformer.

On the second point, yes, you are right, it fails with the lastest version of mypy. There are 3 ways forward, as I see it:

-  The simplest, add `# type: ignore` to feature_engine\selection\base_selector.py:28: line 28 in that file, as per [this thread](https://stackoverflow.com/questions/27567846/how-can-i-check-out-a-github-pull-request-with-git).
- The second simplest, restrict the version of mypy in the test requirements file.
- The hardest, find out why we have that error and fix it (we should probably do this in a separate PR, unless it is a modest workaround)

Unless you have another suggestion?

solegalli:
Hi @ChristopherGS 

We want to introduce the test you find in [this file](https://github.com/feature-engine/feature_engine/blob/main/tests/test_preprocessing/test_check_estimator_preprocessing.py), for the class we are developing in this PR, which you find [here](https://github.com/feature-engine/feature_engine/pull/333/files#diff-5b7c0e1c3591791959727acd2b5d8f597cd1dfde501ec0116b06448689b545abR27).

The thing is, our class only works with variables of type datetime or object. But for the tests, sklearn creates random numerical numpy arrays. So our class will fail all tests, unless we had a flag that allowed the transformer to work with numerical variables as well, but just for the tests. This should not be (obviously) available to users.

Any idea how this could be implemented?

feature_engine/selection/base_selector.py
```diff
@@ -25,7 +25,9 @@ def get_feature_importances(estimator):
         else:
             importances = np.linalg.norm(coef_, axis=0, ord=len(estimator.coef_))
 
-    return list(importances)
+        importances = list(importances)
```
solegalli, line 28: well that was simple! Thank you!

feature_engine/selection/base_selector.py
```diff
@@ -25,7 +25,9 @@ def get_feature_importances(estimator):
         else:
             importances = np.linalg.norm(coef_, axis=0, ord=len(estimator.coef_))
 
-    return list(importances)
```
dodoarg, line 28: the issue - untracked by mypy up to this point - was caused by _importances_ potentially being None (and list(None) isn't allowed)

solegalli:
@ChristopherGS and @dodoarg 

I implemented a plausible solution in this [PR](https://github.com/dodoarg/feature_engine/pull/6)

Let me know what you think please.

solegalli:
@dodoarg 

The only thing remaining in this PR was whether to add the check_estimator or not, correct?

So if we were to not add those tests for now, I think we are good to merge?

I'll capture that in a separate issue. We would like to have other transformers what work with datetime, so it will come back at some point again.

dodoarg:
I think so yes but you have the last word on it. If you spot something missing or wrong I'll try and fix that. I can't wait to see this horrible mega commit history squashed into one

Issue #295, opened 2021-07-18
add functionality to detect introduction of NAN in transform method of discretizers


solegalli:
For the arbitrary, equal width and equal frequency, we should include in the transform method a check to see if NaN are being introduced accidentally.

This can happen if a value is outside the boundaries entered by the user in the arbitrary, or of the variable is too skewed for the other transformers.

The catch should be as a warning to begin with, not to break backwards compatibility perhaps, and should inform the user exactly in which variables the NaN were introduced.

We should also expand the tests of all the transformers to make sure this functionality works. As it is the same functionality for all transformers, and also for the categorical encoders as well as per issue #294, I wonder if we should make a master function from this? I think it depends in the amount of code, if it is a one liner probably not.

solegalli:
At the back of #169

solegalli:
Hi @Morgan-Sell

Would you like to pick up this issue? 

It is very similar to the one we just worked on. 

We need to expand the error message to clarify which variables are the ones where nan were introduced.

Morgan-Sell:
Por supuesto!

Confirming the high-level scope:
- Create a new branch, PR, etc.
- Assess whether there is a NaN value in the dataframe that is inputted in the `transform()` methods for both equal_frequency.py and equal_width.py.
- Create a warning that is in the same vein as the one we developed for the "base_enconder NaN detection" issue.
- Implement the appropriate tests using `pytest.raises` in both test_equal_frequency_discretiser.py and test_equal_width_frequency.py.

Morgan-Sell:
Ah, yes, class inheritance! Thanks for the framework!

Morgan-Sell:
@solegalli, I thought I created a PR for this task. I went to the branch called "discretizer-nan-detection", which I created for this task, and selected "Open pull request".

<img width="927" alt="Screen Shot 2021-12-16 at 4 38 21 PM" src="https://user-images.githubusercontent.com/5661386/146469595-c6d536c9-b7c6-468e-b2c6-f67ae27c0c10.png">

However, I do not see a PR. I've already committed two edits to the "discretizer-nan-detection" branch.

What am I missing? Thanks!

solegalli:
We have a PR here #341 

If you commit changes to the branch called "discretizer-nan-detection" they should appear in the PR straightaway.

Issue #294, opened 2021-07-18
expand categorical detection of NAN after encoding


solegalli:
At the moment, the categorical encoders will raise a warning if NAN are introduced after the encoding. This is controlled from the [BaseCategoricalEncoderClass](https://github.com/solegalli/feature_engine/blob/master/feature_engine/encoding/base_encoder.py#L155).

We should expand this warning to inform the user in which variables exactly the NaN were introduced. 

This functionality is useful in particular to help troubleshoot the SelectByTargetMean transformer.

We also need to expand the test for this class, an example [here](https://github.com/solegalli/feature_engine/blob/master/tests/test_encoding/test_count_frequency_encoder.py#L120), and make sure that all relevant categorical encoders have this check:

solegalli:
at the back of #238

Morgan-Sell:
I'm happy to work on the warning telling the user where the NaN exists. Please assign the task to me.

solegalli:
Super, thank you! Go for it :)

Morgan-Sell:
Just to confirm, the objective is to have the warning, "NaN values were introduced in the returned dataframe by the encoder..." include the names of the columns that are comprised of NaN values and to implement the corresponding unit tests. Is there anything else?

solegalli:
The idea is to change this sentence: 
"NaN values were introduced in the returned dataframe by the encoder." 

by something like:
"NaN values were introduced in variables [list with the variable names] during the encoding."

where [list with the variable names] is the variables in which NaN were introduced. 

We need to change the error message and the code in the class so that it captures the names of the variables in which NaN were introduced.

Then, we need to expand the tests to check that the test is raised correctly and it indicates in which variable NaN were introduced.

Morgan-Sell:
Ok, great, that's what I thought. I implemented the code. I now need to develop the unit test. I don't see a **test_base_encoder.py** file. Should I create one or add to one of the other "test_encoding" files?

solegalli:
We would have to add 1 test in each one of the encoders :)

Also, would you mind making a PR with the code changes you already made and linking this issue?

Morgan-Sell:
I sent the PR. Of course, I immediately noticed an error after sending the PR. The existing code does not pass the tests.

How will I know when I can send a new PR w/ the revised code?

I'll work on the tests later this week.

solegalli:
I left a comment in the PR #338 

We follow the conversation there now that we have a PR :)

Well done with the code changes

PR #334, opened 2021-12-02
WIP: improves SelectByTargetMeanPerformance functinality

SunnyxBd:


Morgan-Sell:
I fetched the branch. I'll get to work ;) 

To help w/ documentation, I pasted the comments/notes from the initial task below.

### INITIAL COMMENT

Idea:

Create predictor class TargetMeanPrediction or similar name with methods fit and predict.

- Fit - learns transformation
- predict - returns the mean target value per observation

Output:

- This transformer will automatically output the mean value of the target per category if variable is categorical (we have an encoder for this).
- If a variable is numerical, it will first discretize it (we have discretizers for this, equal width and frequency, user selects) and then replace by the target mean.

The reason to create a predictor class is that then, we can use it with cross_validate and cross_val_score, in the main selector function.

Things to consider:

- if categorical variables are highly cardinal, the encoders will introduce NaN. In issue expand categorical detection of NAN after encoding #294 we expand the functionality to inform in which variables nan are being introduced to help user troubleshoot
- if a numerical variable is highly skewed, nan can be introduced. In issue, add functionality to detect introduction of NAN in transform method of discretizers #295 we expand the functionality to inform in which variables nan are being introduced to help user troubleshoot

Then we need to re-code the class SelectByTargetMeanPerformance to call our predictor, and use it with cross-validate to return the important features. The advantage of using corss_validate is not just the cross_validation, which offers a less biased score, but it also allows the use of other metrics, not just roc and r2 as what we have at the moment.


### ADDT'L GUIDANCE

It requires investigating the sklearn base classes that are suitable for this estimator. We need (if it exists) the equivalent of the `TransformerMixin` but for an estimator instead of a transformer.

You can get some info from the sklearn docs. Search for "building your own estimator" and "sklearn".

Morgan-Sell:
Hi @solegalli,

It seems like much of the required code for the TargetMeanPredictor class exists within the existing codebase for the SelectByTargetMeanPerformance class. What's the advantage of having the encoder, discretisers, and other functionality within the TargetMeanPredictor class?

solegalli:
The target mean predictor should have methods fit and predict.

It should take a variable, if it is categorical, output the mean value of the target per category. If it is numerical, sort the variable in bins and then replace the bins by the mean value of the target. This would be done in the predict method. The fit method learns the bins limits and the mappings.

If we pack the above code in a predictor class, with fit predict methods, then we can use that class as part of a cross-validate function.

So in short, I would like to take the functionality out of the SelectByTargetMeanPerformance into a predictor class, and then call this new predictor from the SelectByTargetMeanPerformance using cross-validation, instead of the current data split that we do at the moment.

Morgan-Sell:
hola @solegalli,

Thanks for the explanation! 

Should the `predict()` method generate a new dataframe or replace the inputted dataframe?

solegalli:
Predict should return a pandas series with the predictions only, which are the mean value of the target per category or bin. And this is actually a good point, because then this predictor class would only take 1 variable as input, not multiple.

PR #335, opened 2021-12-05
adds functionality to drop input columns in CombineWithReferenceFeature and MatchCombination

kylegilde:
adding the drop_original argument to MathematicalCombination and CombineWithReferenceFeature

### DIFF
--- a/feature_engine/creation/combine_with_reference_feature.py
+++ b/feature_engine/creation/combine_with_reference_feature.py
@@ -98,6 +98,7 @@ class CombineWithReferenceFeature(BaseEstimator, TransformerMixin):
         operations: List[str] = ["sub"],
         new_variables_names: Optional[List[str]] = None,
         missing_values: str = "ignore",
+        drop_original: Optional[bool] = False,
     ) -> None:
 
         # check input types
@@ -158,11 +159,15 @@ class CombineWithReferenceFeature(BaseEstimator, TransformerMixin):
         if missing_values not in ["raise", "ignore"]:
             raise ValueError("missing_values takes only values 'raise' or 'ignore'")
 
+        if not isinstance(drop_original, bool):
+            raise TypeError("drop_original takes only boolean values True and False.")
+
         self.reference_variables = reference_variables
         self.variables_to_combine = variables_to_combine
         self.new_variables_names = new_variables_names
         self.operations = operations
         self.missing_values = missing_values
+        self.drop_original = drop_original
 
     def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
         """
@@ -285,6 +290,9 @@ class CombineWithReferenceFeature(BaseEstimator, TransformerMixin):
         # replace created variable names with user ones.
         if self.new_variables_names:
             X.columns = original_col_names + self.new_variables_names
+
+        if self.drop_original:
+            X.drop(columns=self.variables_, inplace=True)
 
         return X
 
--- a/feature_engine/creation/combine_with_reference_feature.py
+++ b/feature_engine/creation/combine_with_reference_feature.py
@@ -65,6 +65,8 @@ class CombineWithReferenceFeature(BaseEstimator, TransformerMixin):
         transformer will return an error if the training or the datasets to transform
         contain missing values.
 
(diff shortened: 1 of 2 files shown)


feature_engine/creation/combine_with_reference_feature.py
```diff
@@ -65,6 +65,8 @@ class CombineWithReferenceFeature(BaseEstimator, TransformerMixin):
         transformer will return an error if the training or the datasets to transform
         contain missing values.
 
+    drop_original: bool, default=False
+        If True, the original variables to transform will be dropped from the dataframe.
```
solegalli, line 69: I would change the wording to:

If True, the original variables will be dropped from the dataframe after their combination.

feature_engine/creation/mathematical_combination.py
```diff
@@ -61,6 +61,9 @@ class MathematicalCombination(BaseEstimator, TransformerMixin):
         contain missing values. If 'ignore', missing data will be ignored when
         performing the calculations.
 
+    drop_original: bool, default=False
+        If True, the original variables to transform will be dropped from the dataframe.
```
solegalli, line 65: I would change the wording to:

If True, the original variables will be dropped from the dataframe after their combination.

feature_engine/creation/combine_with_reference_feature.py
```diff
@@ -286,6 +293,9 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
         if self.new_variables_names:
             X.columns = original_col_names + self.new_variables_names
 
+        if self.drop_original:
+            X.drop(columns=self.variables_, inplace=True)
```
solegalli, line 297: here we need to drop `self.variables_co_combine + self.reference_variables`.

I don't think this transformer has the attribute `variables_`, does it?

feature_engine/creation/mathematical_combination.py
```diff
@@ -251,6 +259,9 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
         for new_variable_name, operation in self.combination_dict_.items():
             X[new_variable_name] = X[self.variables_to_combine].agg(operation, axis=1)
 
+        if self.drop_original:
+            X.drop(columns=self.variables_, inplace=True)
```
solegalli, line 263: here we need to drop `self.variables_to_combine`.

solegalli (commented):
Hi @kylegilde 

thanks a lot for the addition to the functionality. 

I made a few comments regarding wording of the docs, and the variables that need to be dropped from the dataframe. If I understood correctly, we will be dropping the variables that we combine mathematically, correct?

Also, we would need to add a couple of tests for each of the transformers to confirm that when drop_original=True, the original variables are dropped.

Would you mind adding those tests please?


PS: maybe you were going to add those already and I jumped ahead of my game, if so, apologies.

feature_engine/creation/combine_with_reference_feature.py
kylegilde: Oh, I just copied and pasted the wording from CyclicalTransformer w/o even reading it. Should we update that wording too?

feature_engine/creation/mathematical_combination.py
kylegilde: > 

Can you provide an example of how to write a test? Thank you

feature_engine/creation/combine_with_reference_feature.py
solegalli: Just here please :)

feature_engine/creation/mathematical_combination.py
solegalli: In this [code block](https://github.com/feature-engine/feature_engine/blob/main/tests/test_creation/test_cyclical_transformer.py#L75), we extend the first test of the cyclical transformer to test the drop_original attribute.

In short, I would suggest you pick one of the tests that already exist for each of the transformers, and replicate it, changing the expected output when drop_original=True

solegalli:
@dodoarg if you have time, maybe you would like to pick this one up? :p 

So we can squeeze it in the next release.

PR #339, opened 2021-12-07
Introduces check for var type when user passes var name outside a list closes #337

### LINKED ISSUE #337: no numerical or categorical variable checks when user passes string instead of list
`_find_or_check_numerical_variable`s and `_find_or_check_categorical_variables` don't check variable type when user passes string instead of list

Copying @dodoarg code that spots this issue:

Please have a look at the demo below (on the PowerTransformer, which I haven't touched in my fork) <br>
Here, _variables_ is passed as a list and the correct TypeError is thrown at the _find_or_check_numerical_variables level:
```
>>> pt = PowerTransformer(variables=["City"])
>>> pt.fit(df)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "D:\feature_engine\feature_engine\transformation\power.py", line 82, in fit
    X = super().fit(X)
  File "D:\feature_engine\feature_engine\base_transformers.py", line 98, in fit
  File "D:\feature_engine\feature_engine\variable_manipulation.py", line 82, in _find_or_check_numerical_variables
    raise TypeError(
TypeError: Some of the variables are not numerical. Please cast them as numerical before using this transformer
```
Here instead variables is passed as a str (which _check_input_parameter_variables __allows__) but _find_or_check_numerical_variables doesn't actually run any checks, so the logic goes on up until the line where the column is checked for the presence of inf, which fails because City is an object variable, and should not have made it that far in the code
```
>>> pt = PowerTransformer(variables="City")   
>>> pt.fit(df)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "D:\feature_engine\feature_engine\transformation\power.py", line 82, in fit
    X = super().fit(X)
  File "D:\feature_engine\feature_engine\base_transformers.py", line 102, in fit
  File "D:\feature_engine\feature_engine\dataframe_checks.py", line 127, in _check_contains_inf
    if np.isinf(X[variables]).values.any():
  File "D:\Vari\Anaconda3\envs\ML\lib\site-packages\pandas\core\generic.py", line 1936, in __array_ufunc__
    return arraylike.array_ufunc(self, ufunc, method, *inputs, **kwargs)
  File "D:\Vari\Anaconda3\envs\ML\lib\site-packages\pandas\core\arraylike.py", line 366, in array_ufunc
    result = mgr.apply(getattr(ufunc, method))
  File "D:\Vari\Anaconda3\envs\ML\lib\site-packages\pandas\core\internals\managers.py", line 425, in apply
    applied = b.apply(f, **kwargs)
  File "D:\Vari\Anaconda3\envs\ML\lib\site-packages\pandas\core\internals\blocks.py", line 378, in apply
    result = func(self.values, **kwargs)
TypeError: ufunc 'isinf' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
>>>
```

dodoarg:
Also fixes _int_ variables not being checked
Also fixes the edge case variables = 0 being detected as empty set instead of column named 0 (_any_ replaced by check for non-empty list)

### DIFF
--- a/feature_engine/variable_manipulation.py
+++ b/feature_engine/variable_manipulation.py
@@ -4,6 +4,10 @@ from typing import Any, List, Union
 
 import pandas as pd
 
+from pandas.api.types import is_numeric_dtype as is_numeric
+from pandas.api.types import is_categorical_dtype as is_categorical
+from pandas.api.types import is_object_dtype as is_object
+
 Variables = Union[None, int, str, List[Union[str, int]]]
 
 
@@ -59,10 +63,7 @@ def _find_or_check_numerical_variables(
     variables: List of numerical variables
     """
 
-    if isinstance(variables, (str, int)):
-        variables = [variables]
-
-    if variables is None or len(variables) == 0:
+    if variables is None:
         # find numerical variables in dataset
         variables = list(X.select_dtypes(include="number").columns)
         if len(variables) == 0:
@@ -71,9 +72,18 @@ def _find_or_check_numerical_variables(
                 "format with pandas dtypes"
             )
 
+    elif isinstance(variables, (str, int)):
+        if is_numeric(X[variables]):
+            variables = [variables]
+        else:
+            raise TypeError("The variable entered is not numeric.")
+
     else:
+        if len(variables) == 0:
+            raise ValueError("The indicated list of variables is empty.")
+
         # check that user entered variables are of type numerical
-        if len(X[variables].select_dtypes(exclude="number").columns) > 0:
+        elif len(X[variables].select_dtypes(exclude="number").columns) > 0:
             raise TypeError(
                 "Some of the variables are not numerical. Please cast them as "
                 "numerical before using this transformer"
@@ -106,22 +116,30 @@ def _find_or_check_categorical_variables(
     variables : List of categorical variables
     """
 
-    if isinstance(variables, (str, int)):
-        variables = [variables]
-
-    if variables is None or len(variables) == 0:
+    if variables is None:
+        # find numerical variables in dataset
         variables = list(X.select_dtypes(include=["O", "category"]).columns)
         if len(variables) == 0:
             raise ValueError(
-                "No categorical variables in this dataframe. Please check the "
-                "variables format with pandas dtypes"
+                "No categorical variables in this dataframe. Please check variable"
+                "format with pandas dtypes"
             )
 
+    elif isinstance(variables, (str, int)):
+        if is_categorical(X[variables]) or is_object(X[variables]):
+            variables = [variables]
+        else:
+            raise TypeError("The variable entered is not categorical.")
+
     else:
-        if len(X[variables].select_dtypes(exclude=["O", "category"]).columns) > 0:
+        if len(variables) == 0:
+            raise ValueError("The indicated list of variables is empty.")
+
+        # check that user entered variables are of type numerical
+        elif len(X[variables].select_dtypes(exclude=["O", "category"]).columns) > 0:
             raise TypeError(
-                "Some of the variables are not categorical. Please cast them as object "
-                "or category before calling this transformer"
+                "Some of the variables are not categorical. Please cast them as "
+                "categorical before using this transformer"
             )
 
     return variables
--- a/feature_engine/variable_manipulation.py
+++ b/feature_engine/variable_manipulation.py
@@ -62,7 +62,7 @@ def _find_or_check_numerical_variables(
     if isinstance(variables, (str, int)):
         variables = [variables]
 
-    elif not variables:
+    if variables is None or len(variables) == 0:
         # find numerical variables in dataset
         variables = list(X.select_dtypes(include="number").columns)
         if len(variables) == 0:
@@ -73,7 +73,7 @@ def _find_or_check_numerical_variables(
 
     else:
         # check that user entered variables are of type numerical
-        if any(X[variables].select_dtypes(exclude="number").columns):
+        if len(X[variables].select_dtypes(exclude="number").columns) > 0:
             raise TypeError(
                 "Some of the variables are not numerical. Please cast them as "
                 "numerical before using this transformer"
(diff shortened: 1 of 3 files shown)


feature_engine/variable_manipulation.py
```diff
@@ -62,7 +62,7 @@ def _find_or_check_numerical_variables(
     if isinstance(variables, (str, int)):
         variables = [variables]
 
-    elif not variables:
+    if variables is None or len(variables) == 0:
```
solegalli, line 65: If the user passes an empty list, the transformer should raise an error and not select the numerical variables as we would do with this implementation. This would also break backwards compatibility.

I would keep the elif, so that it is checked only if the first condition is not met.

Could we rework?

feature_engine/variable_manipulation.py
```diff
@@ -62,7 +62,7 @@ def _find_or_check_numerical_variables(
     if isinstance(variables, (str, int)):
         variables = [variables]
```
solegalli, line 63: I would check that this variable is numerical here.

feature_engine/variable_manipulation.py
```diff
@@ -73,7 +73,7 @@ def _find_or_check_numerical_variables(
 
     else:
         # check that user entered variables are of type numerical
-        if any(X[variables].select_dtypes(exclude="number").columns):
+        if len(X[variables].select_dtypes(exclude="number").columns) > 0:
```
solegalli, line 76: why change this line? it is performing the same logic and the previous one was more pythonic.

solegalli (commented):
HI @dodoarg 

thank you for working on this changes!. This is almost there. 

I have a few suggestions. I think the minimum amount of changes to accomplish a task the better.

For this particular issue, adding a check for numerical or categorical just after capturing the string in a list in the first if of the logic is enough. We wouldn't need to change anything else.

If you want, we could introduce a check that the variable list entered by the user is not empty, in this case, we should do that in the final bit of the logic for each function and then add a test to check that it catches this error.

Would you mind updating the code with this changes?

feature_engine/variable_manipulation.py
dodoarg: I'm afraid it is a little more complicated than this, and it took me a few hours of debugging. I'll try to explain.

Before changing the logic, neither strings or int were checked for variable type. Also, _empty lists_ were getting processed exactly the same way as None since __not__ means several things like 0, empty string, empty list, None, etc. 

When I tried to change the logic to fix the string/int issue, one of the check_estimator tests broke (specifically, SelectByTargetMeanPerformance) and after some digging I found out that __empty lists were actually getting passed__ at some point during that check. But since an empty list is not None, the second condition was not met anymore and it would get now processed by the else block in the _find_or_check function and throw an error at some point because no numerical (categorical) variables were found in an empty dataframe.

So I decided to fix said bug by adding the empty list condition (len==0). I reiterate that this was the state of things before changing the code, since "not variables" captures both the None case and the [] case. Again, it went unnoticed because there is no test that directly passes an empty list to the _find_or_check functions.

Finally, about the elif:
keeping the elif means that not only the elif block but also the following __else__ block is only executed if the first condition is not met. But that would get us back to str/int not getting checked, because if a str/int is passed then it matches the first condition and it will be converted into a list, but then it won't enter the else block and it won't be checked for typing which is exactly what was happening before,

feature_engine/variable_manipulation.py
dodoarg: > Also fixes the edge case variables = 0 being detected as empty set instead of column named 0 (any replaced by check for non-empty list)

Because of this

let's say you have a dataframe where one of the columns is named __0__
if you call _find_or_check_numerical_variables(X, variables = [0,1]) you want it to check the variables named 0 and 1 and throw an error because 0 is categorical.
However, the any logic will return False:
```
>>> X = pd.DataFrame({0: ['ohf39', '130hfin'], 1: [234, 19889]})
>>> not_numeric = X[[0,1]].select_dtypes(exclude="number").columns
>>> not_numeric
Int64Index([0], dtype='int64')
>>> any(not_numeric)
False
```
and no error will be thrown. While
```
>>> len(not_numeric) > 0
True
```
thus the error will be thrown. Again, this went unnoticed because no test was passing a column named 0 to the "wrong" _find_or_check function

dodoarg:
Sorry about the verbose responses @solegalli, I tried to keep it simple in the PR description but unfortunately this opened a small can of worms :/

feature_engine/variable_manipulation.py
solegalli: wow, crazy. I did not know this. Thank you!

feature_engine/variable_manipulation.py
solegalli: how about this:

1. if isinstance(variables, (str, int)) ==> if it is of type numeric then [variables] else error
2. elif varables is None (instead of not variables): logic as we have now
4. else: first check that it is not empty, then the code we have now with your len change suggestion.

Would this work?

solegalli:
> Sorry about the verbose responses @solegalli, I tried to keep it simple in the PR description but unfortunately this opened a small can of worms :/

No problem at all, I am always happy to learn new stuff :) 

thanks for the detail and dedication

feature_engine/variable_manipulation.py
dodoarg: I tried to implement the same logic you proposed in issue #67 for the datetime type checker, it works just fine as far as the desired logic goes, but it still breaks the test I mentioned above due to __empty lists being passed__ during that test. Does this qualify as another new issue?

solegalli: could you commit all code you have so far so I can have a look?

dodoarg: @solegalli: changing https://github.com/feature-engine/feature_engine/blob/main/feature_engine/selection/target_mean_selection.py#L204 to include="number" incidentally solves this issue without breaking any other test!

solegalli:
Hi @dodoarg 

Suggested a few changes here:https://github.com/dodoarg/feature_engine/pull/2

This is ready to go. Thank you!

dodoarg:
I believe this is ready to go now?

solegalli:
@SunnyxBd small change here in one of the classes you are working on.

PR #340, opened 2021-12-15
[MRG] Add indicator variables as outputs from winsorizer

### LINKED ISSUE #329: Request: add indicator variables as output from `Winsorizer`
When clipping extreme values in data, oftentimes having an indicator of whether a feature ended up being clipped or not turns out to be a helpful feature. Currently, `Winsorizer` will clip the features and output them with the same name, but does not produce indicator variables.

Would be nice if `Winsorizer` had an option `add_indicator_variables` that would additionally output 0/1 variables indicating whether each of the input columns was clipped towards each side in the output transformation or not.

david-cortes:
Fixes #329 

This PR adds an option in the `Winsorizer` class to output indicator columns telling whether each variable was capped to the left or right sides.


feature_engine/outliers/winsorizer.py
```diff
@@ -95,6 +200,11 @@ class Winsorizer(BaseOutlier):
         both sides. Thus, when `capping_method='quantiles'`, then `'fold'` takes values
         between 0 and 0.20.
 
+    add_indicators: bool, default=False
+        Whether to add indicator variables indicating in each variable was capped.
+        If passing 'True', will output these with the same names as the input
+        variables plus suffizes '_low' and '_high'.
+
```
solegalli, line 207: Could we re-phrase these lines to say something like this:

Whether to add indicator variables to flag the capped outliers.
If 'True', binary variables will be added to flag outliers on the left and right 
tail of the distribution. One binary variable per tail, per variable.

feature_engine/outliers/winsorizer.py
```diff
@@ -136,108 +246,28 @@ def __init__(
...
-                ).to_dict()
-
-            elif self.capping_method == "quantiles":
-                self.left_tail_caps_ = X[self.variables_].quantile(self.fold).to_dict()
-
-        self.n_features_in_ = X.shape[1]
-
-        return self
+        super().__init__(capping_method, tail, fold, variables, missing_values)
+        self.add_indicators = add_indicators
```
solegalli, line 254: we need to add a check here to ensure that the user passes a boolean and not say a list or an integer to this parameter. Could you add that?

And then, we would need to add a test, so ensure that this check is working. Could you add that as well?

feature_engine/outliers/winsorizer.py
```diff
@@ -136,108 +246,28 @@ def __init__(
...
+                X_low = X_out > X
+                X_low.columns = [str(cl) + "_low" for cl in X.columns.values]
+            if self.tail in ["right", "both"]:
+                X_high = X_out < X
+                X_high.columns = [str(cl) + "_high" for cl in X.columns.values]
+            if self.tail in ["left", "both"]:
+                X_out = pd.concat([X_out, X_low], axis=1)
+            if self.tail in ["right", "both"]:
+                X_out = pd.concat([X_out, X_high], axis=1)
+        return X_out
```
solegalli, line 271: A few things to consider here:

1. X.columns.values will add indicators to all variables in the data to transform. But the transformer only caps numerical variables or those entered by the user, so here we should probably use `self.variables_`
2. can we change the names low and high to left and right to be consistent with the names we are already using in the docs and in the class?
3. Instead of X_out > X, which will compare the entire dataset and not just the variables of interest, could we add code that is more specific, and make use of the values that were detected per variable during fit and are stored in the dictionaries?

feature_engine/outliers/winsorizer.py
```diff
@@ -136,108 +246,28 @@ def __init__(
...
 
     transform.__doc__ = BaseOutlier.transform.__doc__
```
solegalli, line 273: Also, instead of this, could we copy and paste the docstrings from the base outlier here so that we can add the bit about:

           **Note:** for 'Winsorizer()' with 'add_indicators=True', the
            output shape will be '[n_samples, 2*n_features]' or
            '[n_samples, 3*n_features]' depending on the tail mode.

Here and not in the base class? we want the docstrings to be specific and not communal :p

tests/test_outliers/test_winsorizer.py
```diff
@@ -164,6 +164,26 @@ def test_quantile_capping_both_tails_with_fold_15_percent(df_normal_dist):
     assert np.round(df_normal_dist["var"].max(), 3) > np.round(0.11823196128033647, 3)
 
 
+def test_indicators_are_added(df_normal_dist):
```
solegalli, line 167: could we add an additional test to test this class when the user selects only 1 or 2 variables from the dataframe to be capped?

This will flag issue about adding indicators only to those variables and not to the entire dataset.

feature_engine/outliers/base_outlier.py
```diff
@@ -65,6 +65,10 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
         -------
         X_new: pandas dataframe of shape = [n_samples, n_features]
             The dataframe with the capped variables.
+
+            **Note:** for 'Winsorizer()' with 'add_indicators=True', the
+            output shape will be '[n_samples, 2*n_features]' or
+            '[n_samples, 3*n_features]' depending on the tail mode.
```
solegalli, line 71: Please remove this lines in the docstrings, we should add it instead to the Winsorizer class.

solegalli (commented):
Hi @david-cortes 

Thank you so much for the changes. This is looking good. Well done. And thank you for contributing to the project.

I have a few suggestions, which I made in the relevant files in this PR. Could you please have a look?

david-cortes:
Added a PR following those comments.

Only, for the part about changing the mechanism from `df1 > df2`, I don't think the alternative would be better, unless implemented directly in the base outlier class. Would you want this to be moved to the base class instead?

david-cortes:
I think the failing test is not related to this PR, as it's also failing in the `main` branch:
```
typechecks create: /home/circleci/project/.tox/unit_tests
typechecks installdeps: mypy
typechecks installed: mypy==0.920,mypy-extensions==0.4.3,tomli==2.0.0,typing-extensions==4.0.1
typechecks run-test-pre: PYTHONHASHSEED='160612845'
typechecks run-test: commands[0] | mypy feature_engine
feature_engine/selection/base_selector.py: note: In function "get_feature_importances":
feature_engine/selection/base_selector.py:28: error: Argument 1 to "list" has incompatible type "Optional[Any]"; expected "Iterable[Any]"
Found 1 error in 1 file (checked 62 source files)
ERROR: InvocationError for command /home/circleci/project/.tox/unit_tests/bin/mypy feature_engine (exited with code 1)
```

solegalli:
> I think the failing test is not related to this PR, as it's also failing in the `main` branch:
> 
> ```
> typechecks create: /home/circleci/project/.tox/unit_tests
> typechecks installdeps: mypy
> typechecks installed: mypy==0.920,mypy-extensions==0.4.3,tomli==2.0.0,typing-extensions==4.0.1
> typechecks run-test-pre: PYTHONHASHSEED='160612845'
> typechecks run-test: commands[0] | mypy feature_engine
> feature_engine/selection/base_selector.py: note: In function "get_feature_importances":
> feature_engine/selection/base_selector.py:28: error: Argument 1 to "list" has incompatible type "Optional[Any]"; expected "Iterable[Any]"
> Found 1 error in 1 file (checked 62 source files)
> ERROR: InvocationError for command /home/circleci/project/.tox/unit_tests/bin/mypy feature_engine (exited with code 1)
> ```

Yes, this error comes with the new release of mypy. We fixed this in a different PR #333

tests/test_outliers/test_winsorizer.py
```diff
@@ -164,6 +164,46 @@ def test_quantile_capping_both_tails_with_fold_15_percent(df_normal_dist):
...
+    transformer = Winsorizer(
+        tail="both", capping_method="quantiles", fold=0.1, add_indicators=True)
+    X = transformer.fit_transform(df_normal_dist)
+    assert X.shape[1] == 3 * df_normal_dist.shape[1]
+    assert np.all(X.iloc[:, df_normal_dist.shape[1]:].sum(axis=0) > 0)
```
solegalli, line 172: I am not sure what this is testing. But for maximum security, could we have a dataframe hardcoded with the expected results, and check that the transformer returns a dataframe exactly like the one we expect?

I think we have some examples where we use pandas assert dataframe equals in this file that you could borrow. 

This way, we are sure that the left indicators are used for values on the left tail and the right indicators for values on the right tail.

tests/test_outliers/test_winsorizer.py
```diff
@@ -164,6 +164,46 @@ def test_quantile_capping_both_tails_with_fold_15_percent(df_normal_dist):
...
+
+def test_indicators_filter_variables(df_vartypes):
+    transformer = Winsorizer(
+        variables=["Age", "Marks"],
+        tail="both",
+        capping_method="quantiles",
+        fold=0.1,
+        add_indicators=True
+    )
+    X = transformer.fit_transform(df_vartypes)
```
solegalli, line 195: same in this test, could we have a df with the expected results? also to test variable name outputs of the added indicators

solegalli (commented):
Hi @david-cortes 

thank you for the quick turnaround with the changes. This is looking really good.

I have a couple of nitpicks for the main transformer, a question about the add indicator logic and some suggestions to make the tests more thorough.

Would you mind having a look at those?

thank you!

solegalli:
> Added a PR following those comments.
> 
> Only, for the part about changing the mechanism from `df1 > df2`, I don't think the alternative would be better, unless implemented directly in the base outlier class. Would you want this to be moved to the base class instead?

No, I think you are right. Like this is better. Thank you!

feature_engine/outliers/winsorizer.py
```diff
@@ -136,108 +133,46 @@ def __init__(
...
-
-        return X
-
-    transform.__doc__ = BaseOutlier.transform.__doc__
+        X_out = super().transform(X)
+        if self.add_indicators:
+            X = X[self.variables_]
+            X_out_filtered = X_out[self.variables_]
+            if self.tail in ["left", "both"]:
+                X_left = X_out_filtered > X
```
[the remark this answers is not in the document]
david-cortes: If an outlier is on the left side, it gets replaced with a higher value. Therefore, the output from the winsorizer will have higher values than the original when left-capping.

david-cortes:
Added a PR with the requested changes in the comments.

feature_engine/outliers/winsorizer.py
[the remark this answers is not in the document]
solegalli: Of course!!!! Thank you!

tests/test_outliers/test_winsorizer.py
```diff
@@ -164,6 +164,83 @@ def test_quantile_capping_both_tails_with_fold_15_percent(df_normal_dist):
...
+    df_out = transformer.fit_transform(df)
+    expected_ind = np.r_[np.repeat(False, 90), np.repeat(True, 10)]
+    pd.testing.assert_frame_equal(
+        df_out.drop("col", axis=1),
+        df.assign(col_right=expected_ind).drop("col", axis=1)
+    )
+
+    transformer.set_params(tail="both")
+    df_out = transformer.fit_transform(df)
+    expected_ind_left = np.r_[np.repeat(True, 10), np.repeat(False, 90)]
```
solegalli, line 233: Am I right to assume that the output of the indicators is a true / false vector? 

can we change to 1/0? we would like the data to be numeric as much as possible, so it can be used straightaway with sklearn models.

solegalli (commented):
Hi @david-cortes 

Excellent changes, thank you so much.

And thank you for answering my questions :)


One tiny extra request, the indicators at the moment are True/False vectors, correct? is there a reason for this?

I would prefer if we could output numerical values, 1/0. Could you please make this change? 

Then we are good to merge :)

Thank you so much for the contribution!

david-cortes:
Changed the indicator variables to be float64 instead.

solegalli:
Thank you @david-cortes 

This is good to go.

PR #338, opened 2021-12-07
base_encoder NaN Detection

Morgan-Sell:
Identifies which columns in the dataframe used in the base_encoder have null values and informs the user of those columns.

closes #351

PR #341, opened 2021-12-16
Add NaN detection to in discretizers transform method by creating a BaseDiscretizer class 

Morgan-Sell:
Started implementation of the `BaseDiscretizer` class which is a child class of the `BaseNumericalTransformer` class.


Morgan-Sell:
Hi @solegalli, I incorporated some of the lessons learned from #338, e.g., add init params - `return_object` and `return_boundaries` - to the new parent class BaseDiscretiser.

Should we add the `rare_labels` parameter w/ the corresponding code raise an error if `self.rare_labels == "raise"`?

Also, I fixed the language to British English, i.e. discretizer --> discretiser ;)

feature_engine/discretisation/__init__.py
```diff
@@ -4,6 +4,7 @@
 """
 
 from .arbitrary import ArbitraryDiscretiser
+from .base_discretiser import BaseDiscretiser
```
solegalli, line 7: We do not add base classes to the init. In the init we only add the transformers that are meant to be used by external users.

The base classes are meant for us, developers :p

Can you please remove it?

feature_engine/discretisation/arbitrary.py
```diff
@@ -7,9 +7,10 @@
 
 from feature_engine.base_transformers import BaseNumericalTransformer
 from feature_engine.validation import _return_tags
+from feature_engine.discretisation import BaseDiscretiser
```
solegalli, line 10: to import it when it is not in the init, you need to do:

`from feature_engine.discretisation.base_discretiser import BaseDiscretiser`

feature_engine/discretisation/base_discretiser.py
```diff
@@ -0,0 +1,120 @@
...
+from feature_engine.dataframe_checks import (
+    _check_contains_na,
+    _check_input_matches_training_df,
+    _is_dataframe,
+)
+
+from feature_engine.base_transformers import BaseNumericalTransformer
+from feature_engine.dataframe_checks import _check_contains_na
+
+import warnings
```
solegalli, line 19: this one should be at the top. 

If you install isort (`pip install isort`), then you can run `isort myfile` and that will sort the imports according to pep8

Here be extra careful, if you run `isort feature_engine`, it might reorganise the imports of many files, and we do not want files not relevant to this PR added here. 

Final note, we normally run isort after we added all relevant code. Like with black and flake8

feature_engine/discretisation/base_discretiser.py
```diff
@@ -0,0 +1,120 @@
...
+    """
+
+    def __init__(
+            self,
+            return_object: bool = False,
+            return_boundaries: bool = False,
+    ) -> None:
+
+        if not isinstance(return_object, bool):
+            raise ValueError("return_object must be True or False")
```
solegalli, line 47: Since we are here, maybe we can expand the error messages to inform the user what value they actually entered. Can you modify the error messages so they have the sentence: `f"Got {entered_parameter} instead."`

So for this example, it should look like this:

```
raise ValueError("return_object must be True or False. "
                   f"Got {return_object} instead.)
```

Thank you!

Some nitpicks: be mindful of the . and the space after it. Otherwise, the second sentence will be printed out just next to the dot in the users console.

feature_engine/discretisation/base_discretiser.py
```diff
@@ -0,0 +1,120 @@
...
+
+        if not isinstance(return_boundaries, bool):
+            raise ValueError("return_boundaries must be True or False")
+
+        self.return_object = return_object
+        self.return_boundaries = return_boundaries
+
+    def _check_transform_input_and_state(self, X: pd.DataFrame) -> pd.DataFrame:
```
solegalli, line 55: We don't need this here. We are importing this functionality from the fit method from the BaseNumericalTransformer.

I know that it looks different from the BaseCategorical, but this was a design error if you want, that is too big to fix in this PR. 

Can you pls remove this function?

feature_engine/discretisation/base_discretiser.py
```diff
@@ -0,0 +1,120 @@
...
+        X: pandas dataframe of shape = [n_samples, n_features].
+            The dataset to transform.
+
+        Returns
+        -------
+        X_new: pandas dataframe of shape = [n_samples, n_features].
+            The dataframe containing the categories replaced by numbers.
+        """
+
+        X = self._check_transform_input_and_state(X)
```
solegalli, line 107: we are inheriting the transform method of the BaseNumericalTransformer, so instead of this, we can use:

`X = super().transform(X)`

That will do the checks you want.

feature_engine/discretisation/base_discretiser.py
```diff
@@ -0,0 +1,120 @@
...
+
+        # check if NaN values were introduced by the encoding
```
solegalli, line 109: replace encoding with discretisation procedure or similar

feature_engine/discretisation/base_discretiser.py
```diff
@@ -0,0 +1,120 @@
+# Authors: Morgan Sell <morganpsell@gmail.com>
```
solegalli, line 1: good call creating this base class.

To summarize, the base class should have a common init shared by most discretisers and a common transform method, also shared by most discretisers.

feature_engine/discretisation/base_discretiser.py
```diff
@@ -0,0 +1,120 @@
...
+        # Check input data contains same number of columns as df used to fit
+        _check_input_matches_training_df(X, self.n_features_in_)
+
+        # check if dataset contains na
+        _check_contains_na(X, self.variables_)
+
+        return X
+
+
+    def transform(self, X: pd.DataFrame) -> pd.DataFrame:
```
solegalli, line 93: if you look at the transform methods from the equal_width, equalfrequency and arbitrary discretizer, they are identical (please double check that what I am saying is true :p). 

So I think we can bring the entire functionality here. More details below.

feature_engine/discretisation/base_discretiser.py
```diff
@@ -0,0 +1,120 @@
...
+            The dataset to transform.
+
+        Returns
+        -------
+        X_new: pandas dataframe of shape = [n_samples, n_features].
+            The dataframe containing the categories replaced by numbers.
+        """
+
+        X = self._check_transform_input_and_state(X)
+
```
solegalli, line 108: We need to bring here the entire transform functionality. First we check that the df is indeed a df and does not contain nan. That comes from super().

Then we discretise:

```
        # check input dataframe and if class was fitted
        X = super().transform(X)

        # transform variables
        if self.return_boundaries:
            for feature in self.variables_:
                X[feature] = pd.cut(X[feature], self.binner_dict_[feature])

        else:
            for feature in self.variables_:
                X[feature] = pd.cut(
                    X[feature], self.binner_dict_[feature], labels=False
                )

            # return object
            if self.return_object:
                X[self.variables_] = X[self.variables_].astype("O")
```

And only now, we can check if the discretisation procedure introduced Nan, which is the bit you added below.

feature_engine/discretisation/equal_frequency.py
```diff
@@ -81,23 +82,16 @@ def __init__(
...
-
+        super().__init__(return_object, return_boundaries)
+        
         self.q = q
         self.variables = _check_input_parameter_variables(variables)
-        self.return_object = return_object
-        self.return_boundaries = return_boundaries
+
 
     def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
```
solegalli, line 96: if all works as planned, we should be able to delete the transform method from this class, because it would inherit that on the base class.

But be mindful of the dosctrings. In short, we need to copy and paste the docstrings from the transform method in this class to the base class.

solegalli (commented):
Hi @Morgan-Sell 

Good call on creating the base class. Since you went there, maybe we can expand the base class with the entire functionality of the transform method.

One thing to notice, we are not encoding here, we are discretising or sorting values into discrete intervals, so we need to change the wording a bit in comments.

regarding the rare_labels, yes, I would like to introduce functionality to give the user the option to raise a warning or an error when NaN are introduced by the discretization procedure. but we can't call it rare_labels, because these are numerical values.

Maybe we just call it errors, like in pandas to datetime?

And if yes, we should probable rename rare_labels to errors as well in the encoders. So we have parameters with similar functionality and similar name across transformers.

feature_engine/discretisation/__init__.py
Morgan-Sell: Ok, good to know. I was unsure. PyCharm suggested it. I assumed the IDE knew better than me. I'm glad to know PyCharm cannot replace me yet ;)

Morgan-Sell:
Hola @solegalli,

I'm still updating the `transform()` method in the BaseDiscretiser class. I need to add the error message. I also need to implement the tests for the 3 discretisers.

Morgan-Sell:
Feliz fiestas, @solegalli!

I created/revised two tests - `test_error_if_input_df_contains_na_in_transform()` and `test_error_if_not_permitted_value_is_errors()` - for ArbirtraryDiscretiser, EqualDistanceDiscretiser, and EqualWidthDiscretiser.

I'm a bit uncertain if I used the correct dataframes - `df_vartypes` and `df_na[["Name", "City", "Age", "Marks", "dob"]]` - for `test_error_if_input_df_contains_na_in_transform()`.

tests/test_discretisation/test_arbitrary_discretiser.py
```diff
@@ -34,3 +36,35 @@ def test_arbitrary_discretiser():
     )
     X = transformer.fit_transform(data)
     pd.testing.assert_frame_equal(X, data_t1)
+
+
+def test_error_if_input_df_contains_na_in_transform(df_vartypes, df_na):
```
solegalli, line 41: I would rename this test to
`test_error_when_nan_introduced_during_transform()`

Because like with the encoders, the transformer should raise an error 1) if the data to transform has NA or 2) when NA are introduced by the transformation itself (the latter is the one you are working on)

tests/test_discretisation/test_equal_frequency_discretiser.py
```diff
@@ -58,13 +58,37 @@ def test_error_if_input_df_contains_na_in_fit(df_na):
 
 def test_error_if_input_df_contains_na_in_transform(df_vartypes, df_na):
     # test case 4: when dataset contains na, transform method
-    with pytest.raises(ValueError):
-        transformer = EqualFrequencyDiscretiser()
+    msg = "During the discretisation, NaN values were introduced " \
```
solegalli, line 61: we need a new test to test your new functionality. But we also need the old test, because that is testing something else.

Could you please restore the old test and pack your new code into a new test?

tests/test_discretisation/test_equal_width_discretiser.py
```diff
@@ -58,13 +58,36 @@ def test_error_if_input_df_contains_na_in_fit(df_na):
 
 def test_error_if_input_df_contains_na_in_transform(df_vartypes, df_na):
     # test case 4: when dataset contains na, transform method
-    with pytest.raises(ValueError):
-        transformer = EqualWidthDiscretiser()
+    msg = "During the discretisation, NaN values were introduced " \
```
solegalli, line 61: same here, we need the old test, and your code into a new function.

feature_engine/discretisation/arbitrary.py
```diff
@@ -106,6 +103,7 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
 
         return self
 
+
     def transform(self, X: pd.DataFrame) -> pd.DataFrame:
```
solegalli, line 107: we need to delete the transform method entirely. It is directly inherited from the parent class.

feature_engine/discretisation/base_discretiser.py
```diff
@@ -0,0 +1,131 @@
...
+                             f"Got {errors} instead.")
+
+        self.return_object = return_object
+        self.return_boundaries = return_boundaries
+        self.errors = errors
+
+
+    def transform(self, X: pd.DataFrame) -> pd.DataFrame:
+        """
+        Replace categories with the learned parameters.
```
solegalli, line 79: the docstrings are wrong. Could you please paste here the docstrings from the transform method from either the equal-frequency or equal-width discretizer?

feature_engine/discretisation/equal_frequency.py
```diff
@@ -144,22 +138,8 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
             The transformed data with the discrete variables.
         """
 
-        # check input dataframe and if class was fitted
+        # transform variables
```
solegalli, line 141: we need to remove the transform method entirely

feature_engine/discretisation/equal_width.py
```diff
@@ -158,22 +151,7 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
-        else:
-            for feature in self.variables_:
-                X[feature] = pd.cut(
-                    X[feature], self.binner_dict_[feature], labels=False
-                )
-
-            # return object
-            if self.return_object:
-                X[self.variables_] = X[self.variables_].astype("O")
+        X = super().transform(X)
```
solegalli, line 155: we need to remove the transform method entirely.

solegalli (commented):
Hi @Morgan-Sell 

Thank you so much for the quick turn-around with the changes. This is looking really good. 

I made a few suggestions for changes here and there. Could you please have a look at those? hopefully that resolves the failing tests.

Otherwise, we need to dig to see what is going on.

Thank you!

Morgan-Sell:
Hi @solegalli,

I've resolved some of the errors. I have a couple of observations/questions:

1.  It seems that ethical questions have been raised about using sklearn `load_boston()`. The sklearn maintainers suggest that we use [fetch_california_housing()](https://scikit-learn.org/stable/modules/generated/sklearn.datasets.fetch_california_housing.html). Are you aware of this issue? Should we replace `load_boston` with `fetch_california_housing`?

2. An error is occurring when `_check_contain_na` is being executed during the discretisers' tests. The specific test is `test_error_when_nan_introduced_during_transform()`. The error seems to be caused by the dataframe that is being introduced to the `transform`.  I see that `_check_contain_na` is executed in the `transform()` method of the BaseNumericalTransformer class. 

- I'm struggling to find the dataframes that are used in the tests, e.g., `df_vartyples` and `df_na`. What's the best approach to audit this error?

When I run `pytest` in PyCharm, I see a subset of the dataframe.

```
X =     Name        City   Age  Marks                 dob
0    tom      London  20.0    0.9 2020-02-24 00:00:00
1   nick  ...24 00:05:00
6   fred     Bristol  41.0    0.8 2020-02-24 00:06:00
7    sam  Manchester  37.0    0.6 2020-02-24 00:07:00
variables = ['Age', 'Marks']
```

solegalli:
Hi @Morgan-Sell 

Thank you so much for working on this issue.

I made a PR to your repo with some changes. As it turns out, the equal width and equal frequency discretizers would not introduce errors, because the end of the first and last bucket are set to infinite.

So the only one that could return an error is the arbitrary discretizer. Thus, I reorganised the code a bit and added the missing test.

Please have a look and let me know your thoughts.

Thank you!

Re:  your questions:

1) yes, I was aware, thank you for changing it.
2) the error was due to the df_na dataframe having NA itself (before the transformation), which is not allowed by the discretizers
3) they are in the conftest.py at the root of tests.

Always check for the dataframes used in the tests in a conftest.py at the root of tests or at the root of the folder, eg, tests_discretizer

Cheers

Morgan-Sell:
Thanks for the help, @solegalli!

Just so I understand, since the min and max of EqualFrequencyDiscretizer and EqualWidthDiscretizer are negative infinity and infinity, there is no possible value that could be out of range when introduced into the `transform()` method of these two classes. 

Consequently, we move the code associated w/ "check if NaN values were introduced by the discretisation procedure" to the ArbitraryDiscretiser because the arbitrary range has finite limits.

Am I correctly understanding the change?

Morgan-Sell (commented):
Negligible error: zero is negative on line 83 of test_arbitrary_discretiser.py. Is that intentional?

`limits_dict = {'var_a':[-5, -2, 0, 2, 5], 'var_b':[-0, 2, 5]}`

feature_engine/discretisation/arbitrary.py
```diff
@@ -39,6 +40,12 @@ class ArbitraryDiscretiser(BaseNumericalTransformer):
         Whether the output, that is the bins, should be the interval boundaries. If
         True, it returns the interval boundaries. If False, it returns integers.
 
+    errors: string, default='ignore'
+        Indicates what to do if no value is assigned to one or more intervals in a
```
solegalli, line 44: We need to re-word this docstring. This is not correct. @Morgan-Sell can you change it to:

Indicates what to do when a value is outside the limits indicated in the `binning_dict`. If 'raise', the transformation will raise an error. If 'ignore', values outside the limits are returned as NaN and a warning will be raised instead.

feature_engine/discretisation/arbitrary.py
```diff
@@ -107,36 +120,32 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
...
-        X_new: pandas dataframe of shape = [n_samples, n_features]
-            The transformed data with the discrete variables.
-        """
+        # check if NaN values were introduced by the discretisation procedure.
+        if X[self.variables_].isnull().sum().sum() > 0:
 
-        # check input dataframe and if class was fitted
-        X = super().transform(X)
+            # obtain the name(s) of the columns with null values
+            tmp = X[self.variables_].copy()
```
solegalli, line 130: @Morgan-Sell 

Can we remove creating a copy and just pass x[self.variables_] to the nan_columns below?

If the df is big, creating a copy may take up a lot of memory in the users computer.

solegalli:
> Negligible error: zero is negative on line 83 of test_arbitrary_discretiser.py. Is that intentional?
> 
> `limits_dict = {'var_a':[-5, -2, 0, 2, 5], 'var_b':[-0, 2, 5]}`

Nope, it was unintentional. Can you fix? and also make the 2 other small changes?

then I think we are good to merge :)

feature_engine/discretisation/arbitrary.py
Morgan-Sell: @solegalli 

If we delete `X = super().transform(X)`, three tests fail. See PyTest summary below:

1.  FAILED test_arbitrary_discretiser.py::test_arbitrary_discretiser - AssertionError: Attributes of DataFrame.iloc[:, 1] (column name="HouseAge") are different
2. FAILED test_arbitrary_discretiser.py::test_error_if_input_df_contains_na_in_transform - Failed: DID NOT RAISE <class 'ValueError'>
3. FAILED test_arbitrary_discretiser.py::test_error_when_nan_introduced_during_transform - Failed: DID NOT WARN. No warnings of type (<class 'UserWarning'>,) was emitted. The list of emitted warnings is: [].

Don't we need to use `X = super().transform()` to inherit the BaseDiscretiser's `transform()` properties?

Morgan-Sell:
@solegalli,

I added back `X = super().transform()` to the ArbitraryDiscretiser `transform()` method for now.

tests/test_discretisation/test_arbitrary_discretiser.py
```diff
@@ -17,12 +17,14 @@ def test_arbitrary_discretiser():
...
     data_t1 = data.copy()
     data_t2 = data.copy()
+
     # HouseAge is the median house age in the block group.
     data_t1["HouseAge"] = pd.cut(data["HouseAge"], bins=[0, 20, 40, 60, np.Inf])
     data_t2["HouseAge"] = pd.cut(
         data["HouseAge"], bins=[0, 20, 40, 60, np.Inf], labels=False
     )
-
+    print(data_t1.head())
```
solegalli, line 26: pls remove these print statements

feature_engine/discretisation/arbitrary.py
solegalli: What I meant here is to delete this line: `tmp = X[self.variables_].copy()` and then to modify line 130 so that it is:

`nan_columns = X[self.variables_].columns[X[self.variables_].isnull().any()].tolist()`

solegalli:
Hi @Morgan-Sell 

Happy new year. Two more changes to go :)

Thank you!

solegalli:
Hi @Morgan-Sell 

Would you be able to fix the style error?

`feature_engine/discretisation/arbitrary.py:130:17: E128 continuation line under-indented for visual indent`

Morgan-Sell:
@solegalli, ooops! I missed that! All fixed!

feature_engine/discretisation/arbitrary.py
```diff
@@ -107,36 +120,32 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
         return self
 
     def transform(self, X: pd.DataFrame) -> pd.DataFrame:
-        """Sort the variable values into the intervals.
```
solegalli, line 110: question: the docstrings were removed. Do they still appear in the docs? otherwise we need to re-introduce

feature_engine/discretisation/arbitrary.py
```diff
@@ -107,36 +120,32 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
...
-
-        else:
-            for feature in self.variables_:
-                X[feature] = pd.cut(
-                    X[feature], self.binner_dict_[feature], labels=False
+        # check if NaN values were introduced by the discretisation procedure.
+        if X[self.variables_].isnull().sum().sum() > 0:
+
+            # obtain the name(s) of the columns with null values
+            nan_columns = X[self.variables_] \
```
solegalli, line 129: I don't think we need the \ to break lines. 

We can write a 1 liner and let black tidy it up.

solegalli (commented):
Hi @Morgan-Sell 

Thank you so much for the quick turn-around.

Apologies that I always seem to come with something else. 

I am just not sure if the docstrings from the transform method will show up correctly for the arbitrary discretizer, since we removed them from the class. Did you build the docs to check?

If not, you can find instructions to build the docs in the Readme file in the repo. `sphink build` and something else is the command.

And finally, a very pedantic comment, I read somewhere that using \ to break lines was not good practice. I am not sure now if this is true or not, but pandas can break automatically across multiple code lines, so we don't need the \. 

Minor, but since we need to check the docs, maybe we can try and remove them?

Thank you!

feature_engine/discretisation/arbitrary.py
Morgan-Sell: What's black? 

The `\` was how I resolved the `feature_engine/discretisation/arbitrary.py:130:17: E128 continuation line under-indented for visual indent` style error.

Creating a new line w/o the `\` still led to the error. The solution was suggested by PyCharm, which is I realize is not necessarily the style guidebook.

Morgan-Sell:
@solegalli,

Wow, I never heard of Sphinx! What a cool package! I <3 the PyCommunity.

The AribtraryDiscretiser `transform()` was listed on "read the docs". I still ran `sphinx-build -b html docs build` to ensure all docstrings are properly captured.

feature_engine/discretisation/arbitrary.py
```diff
@@ -109,34 +122,42 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
...
+           The dataframe to be transformed.
 
-        Returns
-        -------
-        X_new: pandas dataframe of shape = [n_samples, n_features]
-            The transformed data with the discrete variables.
-        """
+       Returns
+       -------
+       X: pandas dataframe of shape = [n_samples, n_features]
```
solegalli, line 132: should be **X_new**

feature_engine/discretisation/base_discretiser.py
```diff
@@ -0,0 +1,82 @@
...
+        """Sort the variable values into the intervals.
+
+        Parameters
+        ----------
+        X: pandas dataframe of shape = [n_samples, n_features]
+            The data to transform.
+
+        Returns
+        -------
+        X: pandas dataframe of shape = [n_samples, n_features]
```
solegalli, line 60: should be **X_new**

feature_engine/discretisation/arbitrary.py
solegalli: Should read like this:
```

            nan_columns = (
                X[self.variables_]
                .columns[X[self.variables].isnull().any()]
                .tolist()
            )
```

with  no \ at the end of the line

solegalli (commented):
Hi @Morgan-Sell 

Almost there. Three more tiny tweaks.

Thanks for checking the docs.

Morgan-Sell:
Hi @solegalli,

It seems like you've incorporated the tweaks and merged the changes to `feature-engine:main`. I'm going to delete the branch.

solegalli:
It is just how sklearn names its inputs and outputs. So I expect most users would be familiar with that. We just copy :p

PR #344, opened 2021-12-18
adds functionality to avoid selecting datetime variables as if they were categorical closes #336

### LINKED ISSUE #336: categorical encoders should not pick datetime variables automatically
If a datetime variable is encoded as object or categorical, and the parameter `variables` is set to None, the transformer will detect those datetime variables as categorical and then encode them.

It might be better practice to exclude datetime variables when the transformer searches for variables automatically

dodoarg:
Hi @solegalli 
this time the commit history shouldn't be too messy.

In case you're wondering why the condition for a variable being categorical is so verbose:
Remember about pd.to_datetime working on numeric columns? 
While that might be convenient for the check estimators, it turns out to be pretty annoying when we need to discriminate between variables that contain datetime/categorical data. I'll explain the two problematic cases that justify the logic I implemented, guiding you through the commits.

- Problematic case 1: categorical variables with integer category types - e.g. `var = [1, 4, 7, 0].astype("category")`
What happens here is that pd.to_datetime(var) returns `ValueError: Unable to fill values because Int64Index cannot contain NA` even when `errors="ignore"`. To prevent that, we need to run the datetime check only when var.dtype.categories is not an integer (or a number, for all matters). 
This is addressed in the first commit of this PR [here](https://github.com/feature-engine/feature_engine/blob/9bb5fc1ed0c79c1cd5941a53da8402c89d4cab31/feature_engine/variable_manipulation.py#L128-L132).

- Problematic case 2: numerical variables cast as object - e.g. `var = ["0", "9", "25", "11"]`
Assumption: we want it to be considered a categorical var (I noticed this is the case in several tests from other modules).
What happens here is that pd.to_datetime(var) converts _var_ back into a numeric variable and then converts it (successfully) to datetime. But since it's now a datetime, as things stand, it will not be returned by the categorical checker function. Therefore, we need to run the datetime check only if we can't convert _var_ into a numeric type. This is what 4123570e7aa303ca2a4b59c47812139fb117038b addresses, leveraging pd.to_numeric.

Now that hopefully things are clearer, there's a couple more things worth mentioning:
- since the typecheck logic was verbose and repeated across the different cases of the _variables_ argument, I decided to refactor it in a new function. I named it _contains\_categorical_ because as things stand _is\_categorical_ is already taken in the namespace of that file. You tell me how to rename that more properly
- the above cases are most likely not accounted for in _find\_or\_check\_datetime\_variables_ and consequently in the _DatetimeFeatures_ transformer. The tests might have been thorough, but turns out the tests are never thorough enough. I guess I'll try and fix that in a separate issue? Shouldn't take too long.
- I finally decided to rearrange the tests for the categorical checker, along the lines of the datetime checker tests (we start with type errors and value errors, then variables=None, etc.). Let me know if it's ok or you'd rather revert to previous state

### DIFF
--- a/feature_engine/variable_manipulation.py
+++ b/feature_engine/variable_manipulation.py
@@ -93,18 +93,23 @@ def _find_or_check_numerical_variables(
     return variables
 
 
-def contains_categorical(column: pd.Series) -> bool:
-    return is_object(column) and (
-            is_numeric(pd.to_numeric(column, errors="ignore"))
-            or not is_datetime(pd.to_datetime(column, errors="ignore", utc=True))
-        ) or (
-            is_categorical(column) and (
-                is_numeric(column.dtype.categories)
-                or not is_datetime(
-                    pd.to_datetime(column, errors="ignore", utc=True)
-                )
-            )
-        )
+def _is_categorical_and_is_not_datetime(column: pd.Series) -> bool:
+
+    # check for datetime only if object cannot be cast as numeric because
+    # if it could pd.to_datetime would convert it to datetime regardless
+    if is_object(column):
+        is_categorical_and_is_not_datetime = is_numeric(
+            pd.to_numeric(column, errors="ignore")
+        ) or not is_datetime(pd.to_datetime(column, errors="ignore", utc=True))
+
+    # check for datetime only if the type of the categories is not numeric
+    # because pd.to_datetime throws an error when it is an integer
+    elif is_categorical(column):
+        is_categorical_and_is_not_datetime = is_numeric(
+            column.dtype.categories
+        ) or not is_datetime(pd.to_datetime(column, errors="ignore", utc=True))
+
+    return is_categorical_and_is_not_datetime
 
 
 def _find_or_check_categorical_variables(
@@ -134,8 +139,9 @@ def _find_or_check_categorical_variables(
     if variables is None:
         # find categorical variables in dataset
         variables = [
-            column for column in X.select_dtypes(exclude="number").columns
-            if contains_categorical(X[column])
+            column
+            for column in X.select_dtypes(include=["O", "category"]).columns
+            if _is_categorical_and_is_not_datetime(X[column])
         ]
         if len(variables) == 0:
             raise ValueError(
--- a/feature_engine/variable_manipulation.py
+++ b/feature_engine/variable_manipulation.py
@@ -123,7 +123,10 @@ def _find_or_check_categorical_variables(
             column
             for column in X.select_dtypes(exclude="number")
             if is_object(X[column])
-            and not is_datetime(pd.to_datetime(X[column], errors="ignore", utc=True))
+            and (
+                is_numeric(pd.to_numeric(X[column], errors="ignore"))
+                or not is_datetime(pd.to_datetime(X[column], errors="ignore", utc=True))
+            )
             or (
                 is_categorical(X[column])
                 and (
@@ -143,7 +146,10 @@ def _find_or_check_categorical_variables(
     elif isinstance(variables, (str, int)):
         if (
             is_object(X[variables])
-            and not is_datetime(pd.to_datetime(X[variables], errors="ignore"))
+            and (
+                is_numeric(pd.to_numeric(X[variables], errors="ignore"))
+                or not is_datetime(pd.to_datetime(X[variables], errors="ignore"))
+            )
             or (
                 is_categorical(X[variables])
                 and (
@@ -169,8 +175,11 @@ def _find_or_check_categorical_variables(
                 for column in variables
                 if not (
                     is_object(X[column])
-                    and not is_datetime(
-                        pd.to_datetime(X[column], errors="ignore", utc=True)
+                    and (
+                        is_numeric(pd.to_numeric(X[column], errors="ignore"))
+                        or not is_datetime(
+                            pd.to_datetime(X[column], errors="ignore", utc=True)
+                        )
                     )
                 )
                 and not (
--- a/feature_engine/variable_manipulation.py
+++ b/feature_engine/variable_manipulation.py
@@ -119,7 +119,21 @@ def _find_or_check_categorical_variables(
 
     if variables is None:
         # find categorical variables in dataset
-        variables = list(X.select_dtypes(include=["O", "category"]).columns)
+        variables = [
+            column
+            for column in X.select_dtypes(exclude="number")
+            if is_object(X[column])
+            and not is_datetime(pd.to_datetime(X[column], errors="ignore", utc=True))
+            or (
+                is_categorical(X[column])
+                and (
+                    is_numeric(X[column].dtype.categories)
+                    or not is_datetime(
+                        pd.to_datetime(X[column], errors="ignore", utc=True)
+                    )
+                )
+            )
+        ]
         if len(variables) == 0:
             raise ValueError(
                 "No categorical variables found in this dataframe. Please check "
@@ -127,7 +141,19 @@ def _find_or_check_categorical_variables(
             )
 
     elif isinstance(variables, (str, int)):
-        if is_categorical(X[variables]) or is_object(X[variables]):
+        if (
+            is_object(X[variables])
+            and not is_datetime(pd.to_datetime(X[variables], errors="ignore"))
+            or (
+                is_categorical(X[variables])
+                and (
+                    is_numeric(X[variables].dtype.categories)
+                    or not is_datetime(
+                        pd.to_datetime(X[variables], errors="ignore", utc=True)
+                    )
+                )
+            )
+        ):
             variables = [variables]
         else:
             raise TypeError("The variable entered is not categorical.")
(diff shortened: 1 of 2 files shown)


feature_engine/variable_manipulation.py
```diff
@@ -93,6 +93,20 @@ def _find_or_check_numerical_variables(
     return variables
 
 
+def contains_categorical(column: pd.Series) -> bool:
+    return is_object(column) and (
```
solegalli, line 97: instead of having this massive return statement, could we include the logic inside the function to output a value? and then return that value?

This logic is designed to tackle the 2 situations you very clearly explained in the PR. The thing is, soon after we merge this PR, we will forget (or at least I will). So could we capture some of the reasoning / problematic cases in comments before the logic?

There are 2 situations, one if it is categorical, maybe capture that individually in some logic (with an if loop?). And the second one if it is object, maybe capture that in a second part of a loop or something. Both blocks of code could output a value. And then, the function returns that value.

Could you see if something on those lines would be possible? it will help us maintain.

feature_engine/variable_manipulation.py
```diff
@@ -93,6 +93,20 @@ def _find_or_check_numerical_variables(
     return variables
 
 
+def contains_categorical(column: pd.Series) -> bool:
```
solegalli, line 96: maybe `is_categorical_variable()`, represents more what the function is doing. What do you think?

feature_engine/variable_manipulation.py
```diff
@@ -119,15 +133,18 @@ def _find_or_check_categorical_variables(
 
     if variables is None:
```
solegalli, line 134: We only want to check and exclude potential datetime variables when `variables is None`.

When the user passes a variable list, I am inclined to go with the user's judgement. If they pass a datetime variable to be treated as categorical, I would let them do that. They hopefully do that for a reason. Maybe it is just a datetime with few dates, like easter, christmas or something else, and it could be treated as categorical.

feature_engine/variable_manipulation.py
```diff
@@ -119,15 +133,18 @@ def _find_or_check_categorical_variables(
...
+        ]
         if len(variables) == 0:
             raise ValueError(
                 "No categorical variables found in this dataframe. Please check "
                 "variable format with pandas dtypes."
             )
 
     elif isinstance(variables, (str, int)):
-        if is_categorical(X[variables]) or is_object(X[variables]):
+        if contains_categorical(X[variables]):
```
solegalli, line 147: as per previous comment, I would not change the logic then variables is a str, or a list.

tests/test_variable_manipulation.py
```diff
@@ -68,42 +68,89 @@ def test_find_or_check_numerical_variables(df_vartypes, df_numeric_columns):
...
-    assert _find_or_check_categorical_variables(df_vartypes, vars_cat) == vars_cat
-    assert _find_or_check_categorical_variables(df_vartypes, None) == vars_cat
-
+    # errors when vars entered by user are not categorical
     with pytest.raises(TypeError):
         assert _find_or_check_categorical_variables(df_vartypes, "Marks")
-
+    with pytest.raises(TypeError):
+        assert _find_or_check_categorical_variables(df_datetime, "datetime_range")
+    with pytest.raises(TypeError):
```
solegalli, line 82: to be absolutely pedantic, could we add a test passing a single variable in a list?

for example add `assert _find_or_check_categorical_variables(df_datetime,[ "datetime_range"])`

this should fail as well, right?

tests/test_variable_manipulation.py
```diff
@@ -68,42 +68,89 @@ def test_find_or_check_numerical_variables(df_vartypes, df_numeric_columns):
...
+
+    # when variables=None
+    assert _find_or_check_categorical_variables(df_vartypes, None) == vars_cat
+    assert _find_or_check_categorical_variables(df_datetime, None) == ["Name"]
+
+    # when vars are specified
+    assert _find_or_check_categorical_variables(df_vartypes, "Name") == ["Name"]
+    assert _find_or_check_categorical_variables(df_vartypes, vars_cat) == vars_cat
 
+    # vars specified, index is numeric
```
solegalli, line 113: what do you mean index is numeric? could you expand the comment?

solegalli (commented):
Hi @dodoarg 

Very happy to see you contribute to another issue :))))))))

Thanks a lot! And very thorough as always.

The code implementation is really good. And the thought process behind it too.

I have 2 suggestions regarding 1) the final functionality we want, and 2) maintaining the code base looking forward.

Regarding the functionality, we only want to exclude datetime when `variables is None`. I would not like to interfere with the users judgement; whenever they think a datetime variable should be treated as categorical, I would let them go with that.

Regarding maintaining, I added a few comments to break the logic down a bit, and maybe expand some comments.

And regarding your question, since you are here, maybe just go ahead and add code to tackle the 2 problematic cases for the datetime variable checker function as well. If you can, please leave comments and break the logic into bite chunks that help us understand why we did these changes in a few months from now.

That's all! Thank you!

dodoarg:
I'm glad you appreciate my contribution, I feel like I have learned a lot considering where I started so thank you for your valuable feedback

Yes I agree on giving the user the last word on what variables should be considered categorical if they specify them. Not sure the way I've broken down the logic is exactly what you meant but it should look better now. What do you think?


> just go ahead and add code to tackle the 2 problematic cases for the datetime variable checker function as well

Just to make sure, you mean in this PR?

tests/test_variable_manipulation.py
```diff
@@ -68,42 +68,95 @@ def test_find_or_check_numerical_variables(df_vartypes, df_numeric_columns):
...
+    assert _find_or_check_categorical_variables(df_vartypes, "Age") == ["Age"]
     assert _find_or_check_categorical_variables(df_vartypes, None) == vars_cat + ["Age"]
     assert _find_or_check_categorical_variables(df_vartypes, ["Name", "Age"]) == [
         "Name",
         "Age",
     ]
 
+    # datetime vars cast as category
+    # object-like datetime
+    df_datetime["date_obj1"] = df_datetime["date_obj1"].astype("category")
```
solegalli, line 134: here we are testing the new functionality. If a variable of time category can be parsed as datetime, then the selector excludes it.

Could we add 1 test to confirm that if the user passes the datetime variable in the variables argument, it is indeed selected?

`assert _find_or_check_categorical_variables(df_datetime, ["Name", "date_obj1"]) == ["Name", "date_obj1"]`

solegalli (commented):
Hi @dodoarg 

Apologies again for the delay in coming back to you. 

This is good to merge. Please see my additional test request below. Could you add that one in please?

And then, let me know if you want to fix these exceptions in the check_datetime_variables in this PR or in a new one. If it is just a few lines of code, do it here. If you think you need to add a lot more code, then better in a new PR. Up to you.

Let me know either way so I know when to merge :)

Thanks a lot and apologies again for the delay. I am back on the roll now :)

dodoarg:
No problems @solegalli I only clicked on request review to make sure you hadn't missed the latest push as I had not tagged you explicitly

I did indeed forget about those tests, my bad, well spotted!

Not sure if this is the right way to go about it but I've worked on the datetime checker issue on a separate branch and then opened a PR to this branch. I think you should be able to see it here https://github.com/dodoarg/feature_engine/pull/7 ? You're not meant to review that code there just tell me if it's too large to be merged here then I'll do a separate PR

solegalli:
The thing is, you are making a PR within your own repo, then, I do not get notifications unless you tag me, and also, conversations will not be easy to find for other users. In other words, as much as possible, we should keep conversations in this repo.

So, I will go ahead and merge this PR. Then, you can make another PR to main from this new branch you created with the changes. And it should look exactly the same (hopefully :p)

dodoarg:
Oh that's why I said I didn't mean to review it there. Basically what I wanted to do is if the changes were small enough (as per your advice) we'd make them part of this PR. Which means I click merge on my own PR and that commit history gets added to this commit history and conversations and reviews continue as usual here where everyone can see them. Just like with your PRs to my datetime feature branch for the datetime issue, only it was me to open it. Basically the only reason behind that was to avoid adding and then reverting commits in the event you found the extra code too chunky.

Sorry I just need to know for learning purposes, am I still missing something or would that work?

solegalli:
I see the point...

The idea was, introduce the changes, check if we are good to continue depending on how big the changes were, if yes, then you merge, if not, you don't merge, and you still get to keep the changes, whereas if we revert the commit, it is somehow lost and more work. It was well thought.

What is the best way in a situation like this? 

I think, given that the issue in the PR was already addressed, and we were now to address another issue that was discovered at the back of this PR, but we were unsure as to how big the needed changes were, the best would have been to merge the original PR with the changes to resolve the original issue. And then open a new PR to address the new issue.

So basically what we did.

Why?

- with PRs to the main repo, I get the notifications and also they are available to all users
- changes may seem small at the beginning, but we may need more changes as we discover new things
- if we decide not to proceed, we can just close the new PR, and this will not affect the former one
- the previous PR will be already in main in case someone else needs it
- smaller PRs are easier to review.

This is what I would do. 

But, I am fairly new in open-source (just 2-3 years). It might be a good idea to contribute to some of the bigger projects, like Pandas or Scikit-learn. Those guys have been on this for much longer and I am sure they have more nailed contributing cycles.

dodoarg:
Thank you so much for the detailed answer! I think we can conclude that the above approach would be ok in principle, just it didn't fit this particular situation where in the end, the best solution was to keep it short and directly related to the issue the PR was originally meant to address

Well I only have roughly one month of experience in open source and more generically in software development done somewhat seriously and not locally or on my own. So I'd say it's much better for me to learn by getting feedback/advices by someone who's experienced in the field but still relatively new and open to discussing stuff etc so that I can actively give my contribution. Also I don't think I'm quite ready to contribute to such giants hehe

PR #347, opened 2021-12-23
updates datetime variable checker function to deal with cat/obj controversial cases raised in #336

dodoarg:
Hi @solegalli 
The last commit is a bit meh in that it actually needs almost as many lines of code as repeating the same tests for a different var type. The _request_ argument in the last commit is a trick I found to let mark.parametrize get their values from fixtures. Otherwise those tests could be put into a standard function (no decorators or anything) that gets called within _tests\_find\_or\_check\_datetime\_vars_. But honestly I might just drop that commit, you tell me

Also I left a comment with a quick question in the issue336 PR you've merged, would you mind having a look when you have time

### DIFF
--- a/feature_engine/variable_manipulation.py
+++ b/feature_engine/variable_manipulation.py
@@ -93,23 +93,31 @@ def _find_or_check_numerical_variables(
     return variables
 
 
+def _is_convertible_to_num(column: pd.Series) -> bool:
+    return is_numeric(pd.to_numeric(column, errors="ignore"))
+
+
+def _is_convertible_to_dt(column: pd.Series) -> bool:
+    return is_datetime(pd.to_datetime(column, errors="ignore", utc=True))
+
+
+def _is_categories_num(column: pd.Series) -> bool:
+    return is_numeric(column.dtype.categories)
+
+
 def _is_categorical_and_is_not_datetime(column: pd.Series) -> bool:
 
     # check for datetime only if object cannot be cast as numeric because
     # if it could pd.to_datetime would convert it to datetime regardless
     if is_object(column):
-        is_categorical_and_is_not_datetime = is_numeric(
-            pd.to_numeric(column, errors="ignore")
-        ) or not is_datetime(pd.to_datetime(column, errors="ignore", utc=True))
+        is_cat = _is_convertible_to_num(column) or not _is_convertible_to_dt(column)
 
     # check for datetime only if the type of the categories is not numeric
     # because pd.to_datetime throws an error when it is an integer
     elif is_categorical(column):
-        is_categorical_and_is_not_datetime = is_numeric(
-            column.dtype.categories
-        ) or not is_datetime(pd.to_datetime(column, errors="ignore", utc=True))
+        is_cat = _is_categories_num(column) or not _is_convertible_to_dt(column)
 
-    return is_categorical_and_is_not_datetime
+    return is_cat
 
 
 def _find_or_check_categorical_variables(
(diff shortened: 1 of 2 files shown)


feature_engine/variable_manipulation.py
```diff
@@ -216,14 +238,9 @@ def _find_or_check_datetime_variables(
...
+                for column in X[variables].select_dtypes(exclude="datetime")
                 if is_numeric(X[column])
-                or (
-                    not is_datetime(X[column])
-                    and not is_datetime(
-                        pd.to_datetime(X[column], errors="ignore", utc=True)
-                    )
-                )
+                or not _is_categorical_and_is_datetime(X[column])
             ]
```
solegalli, line 244: I love it. Very readable. Well done!

tests/test_variable_manipulation.py
```diff
@@ -229,19 +226,52 @@ def test_find_or_check_datetime_variables(df_datetime):
...
-        )
-        == vars_convertible_to_dt + ["date_obj1_cat"]
+        _find_or_check_datetime_variables(df_datetime, variables=vars_convertible_to_dt)
+        == vars_convertible_to_dt
     )
 
 
+@pytest.fixture()
+def cast_age_as_cat(df_datetime):
+    df_datetime["Age"] = df_datetime["Age"].astype("category")
```
solegalli, line 243: When re-defining values of a dataframe, I think a good safety measure is to copy the df first, not to override the original one. 

Not sure it is necessary here, but if we were to share these functions across scripts...

`df_datetime = df_datetime.copy()`

tests/test_variable_manipulation.py
```diff
@@ -229,19 +226,52 @@ def test_find_or_check_datetime_variables(df_datetime):
...
+@pytest.fixture()
+def cast_age_as_obj(df_datetime):
+    df_datetime["Age"] = df_datetime["Age"].astype("O")
+    return df_datetime
+
+
+@pytest.mark.parametrize(
+    "df_age_cast_as_categorical", ["cast_age_as_cat", "cast_age_as_obj"]
+)
+def test_dt_checker_when_numeric_is_cast_as_category_or_object(
```
solegalli, line 256: not sure it is clear which function this test is referencing, maybe change dt_checker by check_datetime_variables

Sorry it is a bit of a mouthful I know :/

solegalli (commented):
Hi @dodoarg 

thanks for the changes. Those in var manipulation are great.

Re: the parametrize, yes, I agree with you, it does not save lines of code.

If, those functions/fixtures were to be used by other tests, then I would probably say keep them. If they are only designed for these particular tests, given than re-casting is a one liner, then I would probably just add the line within the test. I find it more readable that way.

dodoarg:
> given than re-casting is a one liner, then I would probably just add the line within the test.

recasting is a one liner but my main concern was to have all the tests that follow the recasting operation repeated, like in 3f1444e. Which I guess could be obtained by simply grouping those tests in a function, recasting the var once, calling the function that runs them, recasting the var twice, calling the function again.

This last commits sticks with the parametrize decorators but 

1. uses one extra function rather than two
2. the fixture df is copied before recasting the var
3. no fancy pytest.request tricks involved
4. the parametrization is somewhat shared by the repeated tests across _find_or_check_categorical and _find_or_check_datetime
5. thanks to the parametrization, the "problematic cases" are tested both against _Age_ (int) and _Marks_ (float) in the categorical finder instead of just one of them, which isn't particularly relevant but I guess more tests without any extra lines of code are better than less

So I think if this still looks clunky I just revert to 663b3f6 and maybe do what's described in the first paragraph of this comment @solegalli

solegalli:
I think the tests look really good now, tidy and readable. Thank you @dodoarg !

Seems that we are good to merge. Anything else left from your side?

dodoarg:
No problems. The datetime transformer should be safe now until someone finds out it isn't! I think this is good to merge @solegalli

PR #350, opened 2021-12-23
Adds drop_original option to CombineWithReferenceFeatures, solves issue #331

### LINKED ISSUE #331: dropping input columns in CombineWithReferenceFeature
In CombineWithReferenceFeature, would you consider adding a boolean parameter indicating whether the variables_to_combine and reference_variables columns should be dropped? 

drop_inputs=True/False

dodoarg:
There's one more transformer to add this option to right @solegalli? If this is alright I'll just put it there too the same way

closes #335
closes #331

### DIFF
--- a/feature_engine/creation/combine_with_reference_feature.py
+++ b/feature_engine/creation/combine_with_reference_feature.py
@@ -66,7 +66,8 @@ class CombineWithReferenceFeature(BaseEstimator, TransformerMixin):
         contain missing values.
 
     drop_original: bool, default=False
-        If True, the original variables to transform will be dropped from the dataframe.
+        If True, the original variables will be dropped from the dataframe
+        after their combination.
 
     Attributes
     ----------
@@ -296,7 +297,7 @@ class CombineWithReferenceFeature(BaseEstimator, TransformerMixin):
         if self.drop_original:
             X.drop(
                 columns=pd.unique(self.variables_to_combine + self.reference_variables),
-                inplace=True
+                inplace=True,
             )
 
         return X
--- a/feature_engine/creation/combine_with_reference_feature.py
+++ b/feature_engine/creation/combine_with_reference_feature.py
@@ -299,7 +299,7 @@ class CombineWithReferenceFeature(BaseEstimator, TransformerMixin):
 
         if self.drop_original:
             X.drop(
-                columns=pd.unique(self.variables_to_combine + self.reference_variables),
+                columns=set(self.variables_to_combine + self.reference_variables),
                 inplace=True,
             )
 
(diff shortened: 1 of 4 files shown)


solegalli:
yes @dodoarg , the MathCombination.

Did you start from scratch? or pick up the changes from the other contributor in PR #335 ? if you picked up his changes, maybe make a PR to his fork and branch, so we keep it as team work.

dodoarg:
I think theres some overlap in the changes because there's very little code to be added here and we both naturally took inspiration from the existing transformer that has that option already. It's ok with me to PR on his branch but I felt comfortable working this out very quickly and I was worried about that PR being dormant for quite some time after your review so I just opened another one.
If you say it's best practice to PR on that branch do I just open it as is or am I supposed to tweak my commits as to sort of follow up their work?

feature_engine/creation/combine_with_reference_feature.py
```diff
@@ -286,6 +294,12 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
         if self.new_variables_names:
             X.columns = original_col_names + self.new_variables_names
 
+        if self.drop_original:
+            X.drop(
+                columns=pd.unique(self.variables_to_combine + self.reference_variables),
```
solegalli, line 299: I hear that set() instead of pd.unique() is more performant. As in faster. We should probably change.

solegalli (commented):
Nicely done.

feature_engine/creation/combine_with_reference_feature.py
```diff
@@ -158,11 +162,15 @@ def __init__(
         if missing_values not in ["raise", "ignore"]:
             raise ValueError("missing_values takes only values 'raise' or 'ignore'")
 
+        if not isinstance(drop_original, bool):
+            raise TypeError("drop_original takes only boolean values True and False.")
```
solegalli, line 166: can we add to the error message f"Got {drop_original} instead."

I am trying to roll out that way of displaying errors across transformers. It is very sklearn like :p

solegalli:
> 

You are right that these are small changes. So let's go ahead with this PR. It is almost ready. We just need to add the functionality to the MathCombination.

We try to be inclusive and acknowledge everyone's contribution, big or small, so in that sense, taking the other contributors changes and building on top of them, and then making a PR to their fork, is a way to say thank you. We give the opportunity to the contributor to review the code and merge. And if they don't manage to do it, which could happen, we can always make a new PR to the main repo. It might be a slower process. But it feels more welcoming :)

Next time :)

feature_engine/creation/combine_with_reference_feature.py
dodoarg: Sure

feature_engine/creation/combine_with_reference_feature.py
dodoarg: Oh is it? I thought of that for a second. Alright

dodoarg:
should  be good to merge @solegalli
I agree with your point though, if you really care for it I can just PR to that branch instead, it's no big deal for me

Happy holydays!

solegalli:
happy holidays @dodoarg!

And thank you for the great work.

Issue #189, opened 2020-12-05
automate detection of classifier and regressor for tree based transformers


solegalli:
At the moment, in the categorical tree encoder and the tree discretiser, we have an argument is_regression that the user needs to fill in in order to detect if the user is aiming to perform classification or regression.

Sklearn has an automated process with the is_classification (see Decision tree source code).

Can we bring this functionality to feature-engine?

I think we can :p

indymnv:
Hi @solegalli 

This problem looks interesting, is still available? I would like to get more details about it. :)

kind regards.

solegalli:
Hi @indymnv 

At the moment, the classes in feature-engine take a parameter for the user to indicate if if is a regression or classification algorithm. But sklearn has functionality to do this automatically.

This, my idea is to remove this parameter from the classes and instead use sklearn functions is_classifier and is_regressor, to detect this automatically:
https://github.com/scikit-learn/scikit-learn/blob/95119c13af77c76e150b753485c662b7c52a41a2/sklearn/base.py#L771

Should be fairly easy (she said :p)

I would be very happy for you to give this one a go

dataML007:
@solegalli  Looking to contribute. Can I take this if available?

solegalli:
Hey @dataML007 yes, this one is still for the taking. It would be great to have your contribution :)
thank you!

solegalli:
Hey @dataML007 , @indymnv and @NicoGalli 

Would any of you be able to pick up this issue?

It should be fairly straightforward :)

NicoGalli:
@solegalli please assign it to me.

solegalli:
assigned :p

Morgan-Sell:
If no one is tackling this issue, I'll take it on.

Btw, be careful tossing around the term "straightforward"! Very few things are as straightforward as they seem ;p

solegalli:
It would be great if you can pick this one up @Morgan-Sell :)

PR #353, opened 2021-12-29
create auto-detection of classifier and regressor for tree-based transformers

Morgan-Sell:


Morgan-Sell:
#189 - automate detection of classifier and regressor for tree-based transformers

Morgan-Sell:
hi @solegalli,

A couple of questions in how to approach this issue:

- Should I add an `estimator` variable to the DecisionTreeEncoder init() method?
- If so, then which sklearn model object should I use in the tests when instantiating DecisionTreeDiscretiser?

Gracias! Feliz ano!

solegalli:
My idea was to use a function from sklearn that evaluates the target and decides if it is binary or continuous. 

I found this function: [type_of_target](https://scikit-learn.org/stable/modules/generated/sklearn.utils.multiclass.type_of_target.html)

The thing is, that function works well for binary or continuous variables (floats). But it is unclear for targets with just integers.

If a target has integers only, say 0,1,2,3,4, then the `type_of_target` will return multiclass, and we will not know if we have to train a regression or a classification, because it could be either really, depending on the real meaning of the target.

So I am not sure this issue makes sense :/

I've seen a user inadvertently train a regression model on a binary target. Maybe instead of modifying the functionality too much, we could use the `type_of_target` as a sanity check. If binary and is_regression-True, then raise an error, if continuous and is_regression=False, then raise an error and if multiclass, we just go with the users judgement?

Unless you can find another function that we can rely more on?

solegalli:
If user sets `is_regression=False`, we could  check the target with [this function](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/utils/multiclass.py#L180)

If user sets `is_regression=True`, we use the type_of_target and if it is binary we raise an error.

Could you dig a bit in sklearn to see if there is something similar for continuous targets?

Morgan-Sell:
Hi @solegalli,

Am I correct in that the hypothetical `check_regression_target()` fcn would either return **discrete** or **continuous**?

If so, one approach is that we could create an init param called `discrete_threshold`. Let's assume we set the default value to 5%.  If variable "A" is comprised of 5% or less unique values then the function claims that the target series is discrete; otherwise, the target series is continuous.

What do you think?

For example, variable "A" has 1,000 observations and `discrete_threshold` is set to 5%. If variable "A" has 50 or less unique values, then variable "A" is comprised of discrete values; otherwise, the series is continuous.

I'll continue to search for a prebuilt function. However,  I have not found one yet.

Feliz ano!

solegalli:
Hi @Morgan-Sell 

I looked as well, I don't think sklearn check for continuous targets.

We should not add any more parameters to the init file.

If regression is true, we just check that the target is not binary with the type_of_target. If regression is false, we use the check_classification target function from sklearn to check the target. That's all.

If target is discrete, we leave it to the user's judgement.

Morgan-Sell:
Hi @solegalli,

I believe I implemented necessary changes - both in the init() and the test file. Lmk what you think!

feature_engine/encoding/decision_tree.py
```diff
@@ -81,6 +82,16 @@ class DecisionTreeEncoder(BaseCategoricalTransformer):
         type object or categorical. If True, the encoder will select all variables or
         accept all variables entered by the user, including those cast as numeric.
 
+    target_variables: list, default=None
```
solegalli, line 85: we should not add an extra init param. Please remove

feature_engine/encoding/decision_tree.py
```diff
@@ -81,6 +82,16 @@ class DecisionTreeEncoder(BaseCategoricalTransformer):
...
+        The response/target variables of the Sklearn estimator model. These values
+        are to be predicted by the estimator.
+
+    is_regression: bool, default=False
```
solegalli, line 89: we should not rename the param (regression).

feature_engine/encoding/decision_tree.py
```diff
@@ -133,20 +144,30 @@ def __init__(
         cv: int = 3,
         scoring: str = "neg_mean_squared_error",
         param_grid: Optional[dict] = None,
-        regression: bool = True,
```
solegalli, line 136: we should not replace this param by is_regression.

feature_engine/encoding/decision_tree.py
```diff
@@ -133,20 +144,30 @@ def __init__(
...
         random_state: Optional[int] = None,
         variables: Union[None, int, str, List[Union[str, int]]] = None,
         ignore_format: bool = False,
+        target_variables: Union[None, int, str, List[Union[str, int]]] = None,
+        is_regression: bool = False,
     ) -> None:
 
+        if is_regression and type_of_target(target_variables) == "binary":
```
solegalli, line 154: the target is only passed in fit, so this check should go in fit.

solegalli (commented):
Hi @Morgan-Sell 

Thank you so much for the code provided.

We should not rename parameters that already exist (regression, is_regression), because that creates backward incompatibility. And we avoid that as much as we can.

The target is only passed during fit, so the target checks should be moved to the fit method.

Also, try `if self.regression is True` instead of just if self.regression. It is more explicit. If regression was a list or a string, it would be true in the latter, but false in the former.

Thank you!

Morgan-Sell:
@solegalli,

I implemented your suggested edits. Hopefully, we're good to go!

feature_engine/encoding/decision_tree.py
```diff
@@ -81,6 +82,11 @@ class DecisionTreeEncoder(BaseCategoricalTransformer):
         type object or categorical. If True, the encoder will select all variables or
         accept all variables entered by the user, including those cast as numeric.
 
+    regression: bool, default=False
```
solegalli, line 85: Can we replace this with the original docstring? we can find it [here](https://github.com/feature-engine/feature_engine/blob/main/feature_engine/encoding/decision_tree.py#L60).

feature_engine/encoding/decision_tree.py
```diff
@@ -133,20 +139,20 @@ def __init__(
         cv: int = 3,
         scoring: str = "neg_mean_squared_error",
         param_grid: Optional[dict] = None,
-        regression: bool = True,
         random_state: Optional[int] = None,
         variables: Union[None, int, str, List[Union[str, int]]] = None,
         ignore_format: bool = False,
+        regression: bool = False,
```
solegalli, line 145: can we move this back to its original line?

feature_engine/encoding/decision_tree.py
```diff
@@ -133,20 +139,20 @@ def __init__(
...
 
         self.encoding_method = encoding_method
         self.cv = cv
         self.scoring = scoring
-        self.regression = regression
         self.param_grid = param_grid
         self.random_state = random_state
         self.variables = _check_input_parameter_variables(variables)
         self.ignore_format = ignore_format
+        self.regression = regression
```
solegalli, line 155: same here, it was in line 145 before,

feature_engine/encoding/decision_tree.py
```diff
@@ -163,6 +169,15 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
             ordered ordinal encoding.
         """
 
+        # confirm model type and target variables are compatible.
+        if self.regression is True and type_of_target(y) == "binary":
+            raise ValueError(f"'regression' is {self.regression} and target "
+                             f"variables are a binary. These two parameter settings "
+                             f"are not compatible.")
+
+        if self.regression is False:
```
solegalli, line 178: instead of the if condition I would use `else` here as part of the former if.

feature_engine/encoding/decision_tree.py
```diff
@@ -163,6 +169,15 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
             ordered ordinal encoding.
         """
 
+        # confirm model type and target variables are compatible.
+        if self.regression is True and type_of_target(y) == "binary":
+            raise ValueError(f"'regression' is {self.regression} and target "
+                             f"variables are a binary. These two parameter settings "
+                             f"are not compatible.")
```
solegalli, line 176: Can we re-phrase the error message to something like this:

"Trying to fit a regression to a binary target is not allowed by this transformer.
Check the target values or set regression to False."

tests/test_encoding/test_decision_tree_encoder.py
```diff
@@ -109,3 +116,17 @@ def test_variables_cast_as_category(df_enc_category_dtypes):
     transf_df["var_B"] = [0.2] * 10 + [0.4] * 10  # Tree: var_B <= 0.5 -> 0.2 else 0.4
     pd.testing.assert_frame_equal(X, transf_df[["var_A", "var_B"]], check_dtype=False)
     assert X["var_A"].dtypes == float
+
+
+def test_error_is_regression_true_and_target_variable_binary(df_enc):
```
solegalli, line 121: if you want to indulge me a bit more, can we change the test name to:

`test_error_when_regression_is_true_and_target_is_binary`

tests/test_encoding/test_decision_tree_encoder.py
```diff
@@ -109,3 +116,17 @@ def test_variables_cast_as_category(df_enc_category_dtypes):
...
+    with pytest.raises(ValueError):
+        encoder = DecisionTreeEncoder(regression=True)
+        encoder.fit(df_enc[["var_A", "var_B"]], df_enc["target"])
+
+
+def test_error_is_regression_false_and_target_variable_continuous(df_enc_numeric):
```
solegalli, line 127: same here, can we change test name to:

`test_error_when_regression_is_false_and_target_is_continuous`

solegalli (commented):
Hi Morgan,

Thank you very much for the changes and the quick turn-around with the fixes. This is looking really good.

On the files you submitted i have mostly nitpicks:

- we need the init method of the class to look exactly as before, for backwards compatibility. So I need you to change the dosctrings for the regression param back to the original one, and move the order of the regression param to where it was before. I highlighted this with some comments.
- we need some small changes in the if logic and the error message
- small renaming of the tests

And the most important things: 

- We also need thisfunctionality in the [decision tree discretiser](https://github.com/feature-engine/feature_engine/blob/main/feature_engine/discretisation/decision_tree.py)
- And there is some incompatibility to merge automatically. I wonder if this is just that it needs rebasing of main?

Could you have a look?

Thank you!

feature_engine/encoding/decision_tree.py
Morgan-Sell: Ok, I also made a slight revision to the prior `if` statement, so it's more intuitive to the user.

feature_engine/encoding/decision_tree.py
```diff
@@ -133,17 +134,18 @@ def __init__(
         cv: int = 3,
         scoring: str = "neg_mean_squared_error",
         param_grid: Optional[dict] = None,
-        regression: bool = True,
+        regression: bool = False,
```
solegalli, line 137: We need to keep the original defaults. I think it was True.

feature_engine/encoding/decision_tree.py
```diff
@@ -133,17 +134,18 @@ def __init__(
...
         ignore_format: bool = False,
+
     ) -> None:
 
         self.encoding_method = encoding_method
         self.cv = cv
         self.scoring = scoring
-        self.regression = regression
         self.param_grid = param_grid
+        self.regression = regression
```
solegalli, line 148: regression was under scoring. Pls check the original file for position and default values.

Morgan-Sell:
@solegalli, 

I checked the original file and you're correct about the default value and location. I changed the values.

When you say "backward compatibility" is it for functionality, consistency in appearance, or a combination of the two pending the change?

solegalli:
> @solegalli,
> 
> I checked the original file and you're correct about the default value and location. I changed the values.
> 
> When you say "backward compatibility" is it for functionality, consistency in appearance, or a combination of the two pending the change?

It is so that If I am using version 1.1, and I have a lot of code prepared already, and then I download version 1.2, I want to be able to run my code as is, without changing param names, order or having to now add the param because the default value has changed..

The things that you changed will affect those users who do not pass the parameter names to the class, so instead of 
`transformer(strategy='strategy', variables=[mylist], threshold=1]`, they call the transformer as `transformer('strategy', [mylist],1]`
or omit one of the parameters because they are using the default value, for example transformer(strategy='strategy']

solegalli:
I think we just need to add this to the tree discretizer, right?

Almost there :p

Morgan-Sell:
Are you "suggesting" that we refactor the init params names/order for the DecisionTreeDiscretiser class? ;)

I never changed the init params names/order for the DecisionTreeDiscretiser class. Example code:

```
    def __init__(
        self,
        cv: int = 3,
        regression: bool = True,
        ignore_format: bool = False,

    ) -> None:

        self.cv = cv
        self.scoring = scoring
        self.regression = regression
        self.param_grid = param_grid
        self.variables = _check_input_parameter_variables(variables)
        self.ignore_format = ignore_format
```

That's

solegalli:
We need to check that the target coincides with what the user passes in the param regression also in the tree discretizer, using the functions from sklearn

Morgan-Sell:
Will return to resolve the test failures.

Morgan-Sell:
Hi @solegalli,

I implemented the new code and the respective tests. However, I'm receiving the following error due to the ValueError that is raised in fit():

`AssertionError: The error message should contain one of the following patterns:
E               0 feature\(s\) \(shape=\(\d*, 0\)\) while a minimum of \d* is required.
`

I traced the error back to test_all_transformers() in test_check_estimator_discretisers(), which uses sklearn's check_estimator().

I'm a bit thrown off by `feature\`. As far as the regex goes, I understand that `\s` represents any whitespace; `\d` is a single digit; and `*` is a wild card.

The error message that corresponds to the ValueError is `"Trying to fit a regression to a binary target is not allowed by this transformer. Check the target values or set regression to False."`. The same message is used in the DecisionTreeEncoder().

I believe the ValueError message matches the stated criteria in the AssertionError.

Do you have any ideas?

tests/test_encoding/test_decision_tree_encoder.py
```diff
@@ -8,18 +8,22 @@
 
 def test_encoding_method_param(df_enc):
     # defaults
-    encoder = DecisionTreeEncoder()
+    encoder = DecisionTreeEncoder(regression=False)
```
solegalli, line 11: why do we need this change?

tests/test_encoding/test_decision_tree_encoder.py
```diff
@@ -8,18 +8,22 @@
...
     encoder.fit(df_enc, df_enc["target"])
     assert encoder.encoder_[0].encoding_method == "arbitrary"
 
     # ordered encoding
-    encoder = DecisionTreeEncoder(encoding_method="ordered")
+    encoder = DecisionTreeEncoder(
+        encoding_method="ordered", regression=False
```
solegalli, line 17: why do we need this change?

solegalli (commented):
Hi @Morgan-Sell 

Thank you for the changes!

Regarding the first error, my suggestion would be to add `regression=False` to the transformer in [this line](https://github.com/feature-engine/feature_engine/blob/main/tests/test_discretisation/test_check_estimator_discretisers.py#L16) and see if that resolves the issue and does not break it somewhere else. That would be the quickest (if it works)

Also, there is a style check failing:

`tests/test_discretisation/test_decision_tree_discretiser.py:121:62: W292 no newline at end of file`

See as well my comments below.

Thank you!

Morgan-Sell:
Hi @solegalli,

Of course, your suggestion worked! Why did adding `regression=False` to the test_check_estimator_discretisers.py file resolve the error?

I also removed the `regresssion=False` when instantiating the DecisionTreeEncoder() in the test file. That was a great catch! Earlier on I changed the default value of `regression` to `True`. I forgot to reset the value to `regression=False`. I fixed that error too.

Se acabo este PR?!

solegalli:
Hi @Morgan-Sell 

I made some cosmetic changes, and then broke the tests somewhere, so I had to fix and made a [PR](https://github.com/Morgan-Sell/feature_engine/pull/4) to your repo.

If you merge there, then we can merge and close here.

Thank you so much for the great work!!!

tests/test_discretisation/test_decision_tree_discretiser.py
```diff
@@ -106,6 +106,24 @@ def test_non_fitted_error(df_vartypes):
...
+        "var_A": s1,
+        "var_B": s2,
+        "target": [0, 1, 1, 0, 1, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 0, 1, 1, 1, 1]
+    }
+
+    df = pd.DataFrame(data)
+
+    return df
+
+
```
Morgan-Sell, line 126: Why do is `df_discterise()` in the test file and not the conftest file?

solegalli: in the general conftest we put dataframes that are used for testing across modules, by more than 1 transformer.

This df you created was exclusive to test this discretizer. So it is better to put it in the test file it is used.

PR #354, opened 2021-12-30
fixes bug in encoders warning

solegalli:
Atte @Morgan-Sell 

there was a tiny bug in the warning / error that we missed.

Fixing here.


Morgan-Sell:
Thanks, @solegalli! Great catch!

Issue #296, opened 2021-07-18
Improve SelectByTargetMeanPerformance functionality


solegalli:
Idea:

Create predictor class TargetMeanPrediction or similar name with methods fit and predict.

- Fit - learns transformation
- predict - returns the mean target value per observation

Output:

- This transformer will automatically output the mean value of the target per category if variable is categorical (we have an encoder for this). 
- If variable is numerical, it will first discretize it (we have discretizers for this, equal width and frequency, user selects) and then replace by the target mean.

The reason to create a predictor class is that then, we can use it with cross_validate and cross_val_score, in the main selector function.

Things to consider:

-  if categorical variables are highly cardinal, the encoders will introduce NaN. In issue #294 we expand the functionality to inform in which variables nan are being introduced to help user troubleshoot
- if numerical variable is highly skewed, nan can be introduced. In issue, #295 we expand the functionality to inform in which variables nan are being introduced to help user troubleshoot

Then we need to re-code the class SelectByTargetMeanPerformance to call our predictor, and use it with cross-validate to return the important features. The advantage of using corss_validate is not just the cross_validation, which offers a less biased score, but it also allows the use of other metrics, not just roc and r2 as what we have at the moment.

solegalli:
closes #238

SunnyxBd:
HI @solegalli , I'm thinking of picking up this issue.
what do you think?

solegalli:
I think it is a great idea :)
I look forward to it

SunnyxBd:
@solegalli in which folder does this predictor go? in 'creation' folder?

solegalli:
a new one: prediction

does not exist yet :)

SunnyxBd:
@solegalli 
I am trying to push into my branch but I am unable to do it.

remote: Permission to feature-engine/feature_engine.git denied to SunnyxBd.
fatal: unable to access 'https://github.com/feature-engine/feature_engine.git/': The requested URL returned error: 403

Morgan-Sell:
@SunnyxBd, @solegalli 

Has this task been scooped up? ;)

solegalli:
What do you mean? 

The basic idea is at the top of the issue.

It requires investigating the sklearn base classes that are suitable for this estimator. We need (if it exists) the equivalent of the `TransformerMixin` but for an estimator instead of a transformer.

You can get some info from the sklearn docs. Search for "building your own estimator" and "sklearn".

solegalli:
[...]

PR #358, opened 2022-01-05
Create TargetMeanPredictor class to enhance SelectByTargetMeanPerformance 

Morgan-Sell:
SelectByTargetMeanPerformance class will call TargetMeanPredictor allowing the use of cross-validation to return the important features.

Important links:
https://scikit-learn.org/stable/glossary.html#term-predictors
https://scikit-learn.org/stable/glossary.html#term-classifier

Important: in a classifier predict() should output the real label, not 1, 0:
https://scikit-learn.org/stable/glossary.html#term-predict

check example classifier here:
https://scikit-learn.org/stable/developers/develop.html#rolling-your-own-estimator


Morgan-Sell:
Need to link issues #296 and #334.

Morgan-Sell:
Hi @solegalli,

I've made some progress on the TargetMeanPredictor class. In PR #334, we discussed creating the codebase so the `predict()` method accepts a pandas series. This causes some problems given the use of the encoders and discretisers. I've documented the problems below and suggested one potential solution.

**Issues:**
- `X` for the `predict()` method must be a dataframe to be compatible w/ the BaseEncoder class which is the parent class to the MeanEncoder class.
- If we switch input `X` to a dataframe w/ one column, then `_check_input_matches_training_df` in the BaseEncoder class returns a ValueError b/c the shape of dataframe `X` does not coincide with the shape of the dataframe that is used to fit the MeanEncoder().

I suspect it is cumbersome to revise all the inheritance rules/checks from the parent classes, e.g., BaseEncoder, and the encoders/discretisers that are instantiated w/in the TargetMeanPredictor class.

One solution could be to require that the user inputs a dataframe w/ the same dimensions and variables names as the dataframe that is used to fit the TargetMeanPredictor. The `predict()` method could have another parameter - `variable_name` - that instructs the predictor which pandas series to return.

Do you have any thoughts? If so, please share!

Also, the codebase is still in WIP, so it's not in the best shape ;)

feature_engine/prediction/target_mean_prediction.py
```diff
@@ -0,0 +1,212 @@
+# Authors: Morgan Sell <morganpsell@gmail.com>
+# License: BSD 3 clause
+
+from typing import List, Union
+
+import pandas as pd
+from sklearn.base import BaseEstimator, ClassifierMixin, RegressorMixin
```
solegalli, line 7: which functionality do the ClassifierMixin and the RegressorMixin bring to the transformer?

Morgan-Sell: They both bring the `score()` method. ClassifierMixin `score()` returns the mean accuracy. RegressorMixin `score()` returns R-squared.

feature_engine/prediction/target_mean_prediction.py
```diff
@@ -0,0 +1,212 @@
...
+            The target variable.
+        """
+        # check if dataframe
+        _is_dataframe(X)
+
+        if not isinstance(y, pd.Series):
+            y = pd.Series(y)
+
+        # check variables
+        self.variables_ = _find_all_variables(X, self.variables)
```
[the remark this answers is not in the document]
Morgan-Sell: I implemented the functionality. 

One clarifying question, do we still pass a list of all the dataframe's columns to `_find_all_variables()` or should we directly assign `list(X.columns)` to `self.variables_`?

feature_engine/prediction/target_mean_prediction.py
```diff
@@ -0,0 +1,212 @@
...
+
+        # check if df contains na
+        _check_contains_na(X, self.variables_)
+
+        self.n_features_in_ = X.shape[1]
```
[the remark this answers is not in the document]
Morgan-Sell: I moved the code to the end. Two questions:

- Why do we not check for NaN values prior to passing `X` to the encoder and/or discretisers?
- Does `self.n_features_in_ = X.shape[1]` go at the end of `fit()` because the encoder and/or discretisers will transform `X` into `X_new` and the transformation may change the dataframe's shape?

feature_engine/prediction/target_mean_prediction.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+        # identify categorical and numerical variables
+        self.variables_categorical_ = list(X.select_dtypes(include="object").columns)
+        self.variables_numerical_ = list(X.select_dtypes(include="number").columns)
+
+        # transform categorical variables using the MeanEncoder
+        # Should I make this a distinct method?
+        self.encoder = MeanEncoder(variables=self.variables_categorical_)
```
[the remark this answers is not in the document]
Morgan-Sell: Implemented revision.

feature_engine/prediction/target_mean_prediction.py
solegalli: but then, inheriting both would over-write each other's method.

In this case, let's not inherit the Mixins, and let's create our own score() method. If regression one thing, if classification something else. Let's take a look at the score() methods of the mixins for inspiration.

feature_engine/prediction/target_mean_prediction.py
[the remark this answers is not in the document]
solegalli: `_find_all_variables` has functionality to handle different values for variables. Did you take a look at the function?

feature_engine/prediction/target_mean_prediction.py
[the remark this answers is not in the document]
solegalli: On the check of nan point, if I remember correctly, we did not include it, because the transformer you are using in the pipeline already do those checks. So it is redundant.

On the second point, the `check_is_fitted()` function from sklearn that we use in the transform method looks for trailing underscores to decide if the transformer was fit or not. If we do the `self.n_features_in_` at the beginning, this attribute will have a value, even if the functionality below fails, so the check_is_fitted() will think the transformer was indeed fit, when actually, it raised an error.

Putting that check at the end, means that everything else has worked, should the check_is_fitted() found trailing underscores.

feature_engine/prediction/target_mean_prediction.py
```diff
@@ -0,0 +1,244 @@
...
+
+        if self.strategy == "equal-width":
+            discretiser = EqualWidthDiscretiser(
+                bins=self.bins, variables=self.variables_numerical_, return_object=True
+            )
+        else:
+            discretiser = EqualFrequencyDiscretiser(
+                q=self.bins, variables=self.variables_numerical_, return_object=True
+            )
+
```
solegalli, line 204: I would capture lines 196 to 203 in a new method() `make_discretiser()`, that returns the discretizer. So that we can use this method instead of repeating all of this code in the _make_numerical_pipeline() and _make_combined_pipeline()

feature_engine/prediction/target_mean_prediction.py
```diff
@@ -0,0 +1,244 @@
...
+        encoder = MeanEncoder(variables=self.variables_numerical_)
```
solegalli, line 205: in the meanencoder, we need to set he param `errors='raise'`

This is why we introduced this change in the first PR in which we worked together

feature_engine/prediction/target_mean_prediction.py
```diff
@@ -0,0 +1,244 @@
...
+                ("discretisation": discretiser),
+                ("encoder": encoder),
+            ]
+        )
+
+        return _pipeline_numerical
+
+    def _make_categorical_pipeline(self):
+
+        return MeanEncoder(variables=self.variables_categorical_)
```
solegalli, line 218: errors=raise

feature_engine/prediction/target_mean_prediction.py
```diff
@@ -0,0 +1,244 @@
...
+            [
+                ("discretisation": discretiser),
+                ("encoder_num": encoder_num),
+                ("encoder_cat": encoder_cat),
+            ]
+        )
+
+        return _pipeline_combined
+
+
```
solegalli, line 244: need to add the score() method, pls have a look at the classifiermixin and regressionmixin to see what sklearn are doing :p

tests/conftest.py
```diff
@@ -199,3 +199,39 @@ def df_normal_dist():
     df.columns = ["var"]
 
     return df
+
+
+@pytest.fixture(scope="module")
+def df_pred():
```
solegalli, line 205: lets more this df into a conftest inside the test_prediction module.

solegalli (commented):
Hi @Morgan-Sell 

Thank you so much for the quick turnaround with the code.

I think I made some suggestions on how to re-write the transform() method that should answer the issues you raised in your last comment.

If you still encounter errors, I would say, write a couple of tests, it doesn't matter that they are failing, because tests normally helps us (or at least me :p) understand where the problem is coming from.

transform(X) should take in the same dataframe that was used in fit(). Fit can take a df with multiple columns or a df with 1 column. sklearn classes do not take in pandas series, so if we have a series, we need to add .to_frame() for it to work. Please add a test using a series.to_frame() to test the functionality as well.

If something does not work or you feel stuck, don't worry, just draft some tests and we take it from there.

Thank you!

feature_engine/prediction/target_mean_prediction.py
```diff
@@ -0,0 +1,244 @@
...
+    variables: list, default=None
+        The list of variables to evaluate. If None, the transformer will evaluate all
+        variables in the dataset.
+
+    bins: int, default=5
+        If the dataset contains numerical variables, the number of bins into which
+        the values will be sorted.
+
+    strategy: str, default='equal_width'
+        Whether to create the bins for discretization of numerical variables using
```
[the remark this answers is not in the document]
Morgan-Sell: Should we use an underscore or hyphen, e.g., equal_width vs equal-width? I feel like I've seen both uses or I may be confusing feature-engine and sklearn semantics.  I'm adopting your pedantic tendencies ;)

feature_engine/prediction/target_mean_prediction.py
```diff
@@ -0,0 +1,244 @@
...
+                "strategy must be 'equal-width' or 'equal-distance'."
+            )
+
+        if not isinstance(ignore_format, bool):
+            raise ValueError("ignore_format takes only booleans True and False")
+
+        self.variables = _check_input_parameter_variables(variables)
+        self.bins = bins
+        self.strategy = strategy
+        self.ignore_format = ignore_format
```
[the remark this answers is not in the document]
Morgan-Sell: Do not initiate any of the variables, i.e., `variables`, `bins`, and `strategy`?

feature_engine/prediction/target_mean_prediction.py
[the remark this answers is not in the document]
Morgan-Sell: Yes, I looked at the `_find_all_variables`. I was slightly confused by your request to "add functionality for when `variables=None` as that function is built into `_find_all_variables`. Will you please elaborate upon your prior comment?

Morgan-Sell:
Muchas gracias por el feedback, @solegalli! I'm still refactoring the codebase. It's a long list ;)

In the meantime, please poke and prod where you see fit!

feature_engine/prediction/target_mean_prediction.py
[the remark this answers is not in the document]
solegalli: self.ignore_format = ignore_format

feature_engine/prediction/target_mean_prediction.py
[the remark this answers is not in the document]
solegalli: trailing underscores are only added to attributes that learn parameters during the fit() method as per sklearn conventions.

tests/conftest.py
Morgan-Sell: Done. What's the benefit of having a separate conftest file?

feature_engine/prediction/target_mean_prediction.py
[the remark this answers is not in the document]
Morgan-Sell: I'm referring to the potential values for the `self.strategy` parameter. The parameter can either be `equal-width` or `equal-distance`.

Should I use (1) `equal_distance` and `equal_width` or (2) `equal-distance` and `equal-width`?

Option 1 uses an underscore. Option 2 uses a hyphen.

I suspect we want to be consistent w/ sklearn and within the feature-engine library.

feature_engine/prediction/target_mean_prediction.py
```diff
@@ -0,0 +1,244 @@
...
+        # NOTES:
+        # - X needs to be a dataframe to be compatible w/ the BaseEncoder()
+        # - X needs to match the shape of the dataframe used in fit()
+
+        # check if is pandas series w/ a name that matches
+        #
+        # if not isinstance(X, pd.Series):
+        #     raise TypeError("fit() method only accepts pandas series.")
+
+        if variable_name not in (self.variables_):
```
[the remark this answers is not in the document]
Morgan-Sell: She nailed it! :)

Morgan-Sell:
Hi @solegalli,

Given that there is a `mean_accuracy_score` method, I suspect we will also use TargetMeanPredictor to transform variables to class labels, e.g. binary classes. If so, should we create a `predict_proba` method?

Morgan-Sell:
Hi @solegalli,

Given that the TargetMeanPredictor class has a `mean_accuracy_score` method, should the class object be able to return classification labels, e.g. binary? If so, should we build a `predict_proba` method?

solegalli:
Hi @Morgan-Sell 

Before I review, could you please try and review the files committed to this PR? I think something went wrong with rebase.

Check out main, pull upstream main to your main branch, then check out your feature branch, rebase main onto it, and then force push.

thank you!

Morgan-Sell:
Hi @solegalli,

I recall this happened when we worked on issue #338 BaseEncoder NaN Detection.

On both occasions, I followed the directions on the "contribute code".

```
git checkout main
git pull --rebase upstream main
git push -f origin main
git checkout myfeaturebranch
git rebase main
```

These directions are slightly different than the ones that you told me when I made the initial mistake in issue #338. 

```
git checkout main 
git pull -rebase upstream main 
git checkout -yourfeaturebranch 
git rebase main
git push -f origin yourfeaturebranch
```
I just implemented the above code and I believe the problem is now resolved.

After careful reading, I believe that I only use `git push -f origin main` if there are conflicting files.

I would not be surprised if I'm the only one who has made this mistake, so no action may be required. I just thought I pass it along as an FYI.

Abrazo!

solegalli:
> Hi @solegalli,
> 
> I recall this happened when we worked on issue #338 BaseEncoder NaN Detection.
> 
> On both occasions, I followed the directions on the "contribute code".
> 
> ```
> git checkout main
> git pull --rebase upstream main
> git push -f origin main
> git checkout myfeaturebranch
> git rebase main
> ```
> 
> These directions are slightly different than the ones that you told me when I made the initial mistake in issue #338.
> 
> ```
> git checkout main 
> git pull -rebase upstream main 
> git checkout -yourfeaturebranch 
> git rebase main
> git push -f origin yourfeaturebranch
> ```
> 
> I just implemented the above code and I believe the problem is now resolved.
> 
> After careful reading, I believe that I only use `git push -f origin main` if there are conflicting files.
> 
> I would not be surprised if I'm the only one who has made this mistake, so no action may be required. I just thought I pass it along as an FYI.
> 
> Abrazo!

The top directions as in docs are to keep your docs in sync.

The bottom directions are to rebase main onto your featurebranch, that is, to keep your feature branch updated with the latest changes. This is important only when the latest changes affect the files you are working on.

feature_engine/prediction/target_mean_prediction.py
```diff
@@ -0,0 +1,271 @@
...
+    ----------
+
+
+    """
+
+    def __init__(
+        self,
+        variables: Union[None, int, str, List[Union[str, int]]] = None,
+        bins: int = 5,
+        strategy: str = "equal-width",
```
solegalli, line 74: I believe we use underscore instead of hyphen in the strategy arguments. Can you check other classes and change the hyphen for underscore? so equal**_**width instead of equal-width?

feature_engine/prediction/target_mean_prediction.py
```diff
@@ -0,0 +1,271 @@
...
+        r2_score : float
+            Linear correlation between 'y' and 'y_pred'.
+
+        """
+        # Should we use the same or different nomenclature as sklearn?
+
+        y_pred = self.predict(X)
+        return r2_score(y, y_pred, sample_weight=sample_weight)
+
+    def mean_accuracy_score(self, X: pd.DataFrame, y: pd.Series,
```
solegalli, line 247: Let's unify methods `r_squared_score` and `mean_accuracy_score` into a method called `score()` so that it is the same name that in the sklearn ClassifierMixin or RegressorMixin.

We should have an additional init param, `regression=True`, that the user needs to set based on its target. If regression is true, then we use r2, if false then we use accuracy.

Also, we need to add the target checks based on the value of the init regression parameter like you did in the discretizer tree in a previous PR.

feature_engine/prediction/target_mean_prediction.py
```diff
@@ -0,0 +1,271 @@
...
+        Returns
+        --------
+        accuracy_score : float
+            Mean accuracy of 'self.predict(X)' and 'y'.
+
+        """
+        # Should we use the same or different nomenclature as sklearn?
+
+        y_pred = self.predict(X)
+        return accuracy_score(y, y_pred, sample_weight=sample_weight)
```
solegalli, line 271: I know that you copied sample_weight=sample_weight from the sklearn code. But I think we need to default to None here.

solegalli (commented):
Hi @Morgan-Sell 

This is looking really, really good. Very well done and thank you so much.

Main next steps are i) expanding the tests and ii) the docstrings.

Could you expand the tests to include the following:

1. test predictor class when all variables are numerical
2. test predictor class when all variables are categorical
3. test predictor class when user passes only 1 variable (test both passing a num var and a cat var)
4. test score method when regression true or false
5. test non fitted error
6. test wrong params as init inputs
7. test error raises when df has nans in fit or transform

Finally, could you expand the docstrings?

I also added 2 minor tweaks in some comments below

Thank you so much for the great work!

Morgan-Sell:
Hi @solegalli,

Should the class object be able to return classification labels, e.g. binary? If so, should we build a predict_proba() method?

feature_engine/prediction/target_mean_prediction.py
Morgan-Sell: `score()` is defined by the following code:

`def score(self, X: pd.DataFrame, y: pd.Series, regression: bool = True,
              sample_weight: float = None) -> float:`

Isn't `sample_weight = None` the default?

solegalli:
> 

Good point on the classification binary vs probability. I did not think of this. And actually, this could simplify our lives a lot!

Sklearn separates classifiers and regressors. They never have the 2 combined. So I am thinking, that we should do the same. So instead of 1 class that does all, we should have the **TargetMeanRegressor** and **TargetMeanClassifier.** The first, inherits the RegressorMixin and the second the ClassifierMixin, so that will automatically give them the score functionality that we need, and we do not need to code it here. And like this, we could have 1 method, predict() for the regressor, and the 2 methods, predict() and predict_proba() for the classifier. 

Would you mind re-accomodating the code for this class break?

On this point, another thing to consider is that the probabilities need to add to 1. So if we are taking the average of various probabilities, we may need to do some normalization at some point.

On the second point, sklearn enforces users to pass a df and estimators do not work with pandas series. So if you pass df['age'] it will raise an error, You have to pass df['age'].to_frame() for sklearn predictors to work. I am thinking we should do the same. But I am not sure why yet. What do you think?

Morgan-Sell:
Hola @solegalli,

I started coding your suggestions. It allows me to think through the steps.

A couple of questions/responses/thoughts:

1.  We'll now have two classes - TargetMeanRegressor and TargetMeanClassifier. By having two distinct classes, are we limiting users by not allowing them to fit dataframes that are comprised of both numerical and categorical variables?

2. When you said that "the probabilities need to add to one", are you referring to the probabilities per observation/row that are returned by the TargetMeanClassifier `predict_proba` method? If so, that makes sense. If not, will you please elaborate?

3.  I like the idea of staying consistent w/ sklearn unless we're improving upon sklearn functionality. feature-engine is the avocado to sklearn's avocado toast! Btw, everyone knows the avocado is the best part ;)


I think our decision to Q1 will have a substantive impact on our path forward. I'll hold off until we decide.

Morgan-Sell:
Halo @solegalli,

How do we apply the MeanEncoder to the TargetMeanClassifier? Or, should we replace the MeanEncoder w/ a GreatestFrequencyEncoder? In this case `predict` would look as follows:

1. Determine the most-frequently-occurring label for each category or discretized bin for each variable.
2. Select and assign the most-frequently-occurring label among all the variables for each specific observation or dataframe row.

I'm imagining the following for `preditct_proba`:
1. Determine the most-frequently-occurring label for each category or discretized bin for each variable.
2. Calculate the percentage of occurrence of each label for each individual observation / dataframe row.
3. Return an array of the percentage of occurences of each label per observation / dataframe row.

I realize we're kind of cheating the `predict_proba` by not counting the weight (percentage of occurrence) in step 1.

Que piensas?

solegalli:
Hi @Morgan-Sell 

Thank you so much for the quick turn-around with the code.

I made a [PR](https://github.com/Morgan-Sell/feature_engine/pull/5) with some code changes.

On this question: "When you said that "the probabilities need to add to one", are you referring to the probabilities per observation/row that are returned by the TargetMeanClassifier predict_proba method? If so, that makes sense. If not, will you please elaborate?"

I need to think this further and sit with some paper and make some calculations to convince myself that this is necessary. 
If the user passes 3 variables, age, marks and city. Then, observation 1 would have 3 mean values of the target coming from each one of the variables, say 0.8 for age, 0.6 for marks and 0.5 for city. So the final output would be the average of (0.8+0.6+0.5)/3

But, by definition, probabilities add up to 1 when taken across all observations. If we have 1 input variable, this is probably the case. But when we take the average across multiple variables, I am not sure it holds. And I am not sure we need to normalize, or how we should normalize. I need to think about this a bit further. Suggestions are welcome.

I think the changes in the PR should answer the remaining questions. If not let me know. 

Thank you!

feature_engine/prediction/base_predictor.py
```diff
@@ -0,0 +1,235 @@
...
+    """
+
+    def __init__(
+        self,
+        variables: Union[None, int, str, List[Union[str, int]]] = None,
+        bins: int = 5,
+        strategy: str = "equal_width",
+    ):
+        # boolean value can be interpreted as an integer
+        if not isinstance(bins, int) or isinstance(bins, bool):
```
solegalli, line 92: bins should just take integers, not booleans, it does not matter if booleans can be interpreted as integers.

Could we change please?

feature_engine/prediction/target_mean_classifier.py
```diff
@@ -0,0 +1,192 @@
...
+
+        Return
+        -------
+        T: array-like of shape (n_samples, n_classes)
+            Returns the log-probability of the sample for each class in the model,
+            where classes are ordered as they are in self.classes_.
+
+        """
+        log_prob = np.log(self.predict_proba(X))
+        if not isinstance(log_prob, np.ndarray) or log_prob.shape[1] != 2:
```
solegalli, line 164: Why do we need this check?

feature_engine/prediction/target_mean_classifier.py
```diff
@@ -0,0 +1,192 @@
...
+            The input samples.
+
+        Return
+        -------
+        T: array-like of shape (n_samples, n_classes)
+            Returns the probability of the sample for each class in the model, where
+            classes are ordered as they are in self.classes_.
+        """
+        prob = self._predict(X)
+        if not isinstance(prob, np.ndarray) or prob.shape[1] != 2:
```
solegalli, line 137: why this check? what is it ensuring? 

Also, is the output of _predict() not a 1-D array, always?

feature_engine/prediction/target_mean_classifier.py
```diff
@@ -0,0 +1,192 @@
...
+        X : pandas dataframe of shape = [n_samples, n_features]
+            The input series which must have the same name as one of the features in the
+            dataframe that was used to fit the predictor.
+
+        Return
+        -------
+        y_pred: ndarray of shape (n_samples,)
+            Vector containing the class labels for each sample.
+        """
+        y_pred = np.where(self.predict_proba(X) > 0.5, 1, 0)
```
solegalli, line 186: is the output of predict_proba not a 2 column array? which probability should be bigger than 0.5?

feature_engine/prediction/target_mean_classifier.py
```diff
@@ -0,0 +1,192 @@
...
+        if not isinstance(y_pred, np.ndarray) or y_pred.ndim != 1:
```
solegalli, line 187: why do we need this check?

feature_engine/prediction/target_mean_regressor.py
```diff
@@ -0,0 +1,124 @@
...
+            The input samples.
+
+        Return
+        -------
+        y_pred: ndarray of shape (n_samples,)
+            Returns predicted values.
+        """
+
+        y_pred = self._predict(X)
+        if not isinstance(y_pred, np.ndarray) or y_pred.ndim != 1:
```
solegalli, line 119: why do we need this check?

feature_engine/prediction/target_mean_regressor.py
```diff
@@ -0,0 +1,124 @@
...
+            raise ValueError(
```
solegalli, line 120: we normally raise errors to help users understand what they are doing wrong. 

But here, we are raising an error that basically checks that our method _predict() is doing the right job. So if _predict() does not, the user won't have any clue what to do. And won't be able to fix it, other than messing with the source code. So it is not useful.

Instead, we need to ensure the output of _predict, is what it should be.

Why did you add this check all along? what did I miss?

feature_engine/variable_manipulation.py
```diff
@@ -283,3 +281,54 @@ def _find_all_variables(
         X[variables]
 
     return variables
+
+
+def _find_categorical_and_numerical_variables(
+        X: pd.DataFrame, variables: Variables = None
+):  # TODO: add typehint to output
```
solegalli, line 288: we need to figure out the typehint here.

solegalli (commented):
Hi @Morgan-Sell 

First of all, apologies for the silence in the last several days. 

I am without internet since last Thursday, and to make things worse, my grandma had covid, and I was with her, so I didn't want to go to my father's or brother's to borrow their internet, until I made a PCR test for covid.

So now I am hopefully back on track.

(Grandma is doing great by the way)

**Back to the PR:**

I think I jumped a bit ahead of my game, because I noticed that you are still working on this PR.

But since I had a look, I have just a very minor nitpick respect to the check for bins (comment below), and then maybe you can help me understand why we need all those checks after the predict_, predict_proba and predict_log_proba?

I am probably missing something there.

And, something that I should have said first: awesome progress with the code, particularly the tests for the regressor. Thank you so much :)

feature_engine/prediction/base_predictor.py
Morgan-Sell: I created the variable `_false_input_params` which is comprised of the following elements.
```
_false_input_params = [
    ("salsa", "arbitrary"),
    ("33", "mean-encoder"),
    ([7], True),
    (False, "prost"),
]
```

The assertion test associated with the last pair `(False, "prost")` was failing, i.e., a ValueError was not raised. To resolve the error, I added `or isinstance(bins, bool)` to the `if` statement to ensure a ValueError was raised in case the user inputs `True` or `False` for `bins` when instantiated the class.

I was surprised by the result. If we remove `or isinstance(bins, bool)` then the code does not pass`test_raises_error_when_wrong_input_params()`.

solegalli: Makes sense!! Thank you!

Morgan-Sell:
Hi @solegalli,

You asked why I created the checks in `predict()`, `predict_proba()`, and `predict_log_proba()`. In the [[MRG] regactors Mean target prediction ](https://github.com/Morgan-Sell/feature_engine/pull/5) PR, there are a number of comments like `# TODO: check that this outputs a numpy array with 2 columns`.

Consequently, I implemented the aforementioned checks. I guess I misunderstood your requests. What did you mean by checking for numpy arrays?

Gracias!

Morgan-Sell:
Hola @solegalli,

I implemented all the variable-manipulation unit tests except for `test_error_find_cat_and_num_vars_datetime_var()`. `_find_categorical_and_numerical_variables()` is returning `datetime_range` as a categorical variable.

I'm stuck! `_find_categorical_and_numerical_variables()` is working on a notebook; however, I cannot translate it to the code base.

What am I missing? :(

tests/test_variable_manipulation.py
```diff
@@ -275,3 +276,102 @@ def test_find_all_variables(df_vartypes):
 
     with pytest.raises(KeyError):
         assert _find_all_variables(df_vartypes, non_existing_vars)
+
+
+def test_find_categorical_and_numeric_variables_one_categorical_variables(
```
solegalli, line 281: could we put all these tests together (into 1) and use parametrize? 

also, can we use a df that already exists? like df_vartypes?

feature_engine/prediction/base_predictor.py
```diff
@@ -0,0 +1,235 @@
...
+        Parameters
+        ----------
+        X : pandas dataframe of shape = [n_samples, n_features]
+            The training input samples.
+
+        y : pandas series of shape = [n_samples,]
+            The target variable.
+        """
+        # check if 'X' is a dataframe
+        _is_dataframe(X)
```
solegalli, line 118: X = is_dataframe(X)

needs to return X

feature_engine/prediction/base_predictor.py
```diff
@@ -0,0 +1,235 @@
...
+        Return
+        -------
+        y_pred: pandas series of shape = [n_samples, ]
+            The mean target value per observation.
+        """
+        # check method fit has been called
+        check_is_fitted(self)
+
+        # check that input is a dataframe
+        _is_dataframe(X)
```
solegalli, line 222: X = is_dataframe(X)

feature_engine/prediction/base_predictor.py
```diff
@@ -0,0 +1,235 @@
...
+        ----------
+        X : pandas dataframe of shape = [n_samples, n_features]
+            The training input samples.
+
+        y : pandas series of shape = [n_samples,]
+            The target variable.
+        """
+        # check if 'X' is a dataframe
+        _is_dataframe(X)
+
```
solegalli, line 119: Hi @Morgan-Sell 

would you be so kind to experiment with adding this check and see if the tests still pass:

```
# Check that X and y have correct shape
X, y = check_X_y(X, y, dtype=None)
```

check_X_y is from sklearn and supposedly checks that X and y fulfill some criteria.

feature_engine/prediction/base_predictor.py
```diff
@@ -0,0 +1,235 @@
...
+            self.variables_numerical_,
+        ) = _find_categorical_and_numerical_variables(X, self.variables)
+
+        # check for missing values
+        _check_contains_na(X, self.variables_numerical_)
+        _check_contains_na(X, self.variables_categorical_)
+
+        # check inf
+        _check_contains_inf(X, self.variables_numerical_)
+
```
solegalli, line 132: if we use Check_X-y we do not need to check for na and inf ourselves. In theory. But I have not tried.

I was wondering if you could try commenting this out and see if the tests still pass?

feature_engine/prediction/target_mean_classifier.py
```diff
@@ -0,0 +1,195 @@
...
+        Parameters
+        ----------
+        X : pandas dataframe of shape = [n_samples, n_features]
+            The training input samples.
+
+        y : pandas series of shape = [n_samples,]
+            The target variable.
+        """
+        # check that y is binary
+        self.classes_ = list(y.unique())
```
solegalli, line 101: can we use unique_labels from sklearn instead?
https://scikit-learn.org/stable/developers/develop.html#rolling-your-own-estimator

feature_engine/prediction/target_mean_classifier.py
```diff
@@ -0,0 +1,195 @@
...
+        if len(self.classes_) > 2:
+            raise ValueError(
+                "This encoder is designed for binary classification only. The target "
+                "has more than 2 unique values."
+            )
+
+        # if target has values other than 0 and 1, we need to remap the values,
+        # to be able to compute meaningful averages.
+        if any(x for x in y.unique() if x not in [0, 1]):
+            y = np.where(y == y.unique()[0], 0, 1)
```
solegalli, line 112: can we use self.classes_ instead of y.unique()?

feature_engine/prediction/target_mean_classifier.py
```diff
@@ -0,0 +1,195 @@
...
+            The input samples.
+
+        Return
+        -------
+        T: array-like of shape (n_samples, n_classes)
+            Returns the probability of the sample for each class in the model, where
+            classes are ordered as they are in self.classes_.
+        """
+        prob = self._predict(X)
+        # TODO: Discuss w/ Soledad
```
solegalli, line 137: remove all the commented

feature_engine/prediction/target_mean_classifier.py
```diff
@@ -0,0 +1,195 @@
...
+
+        Return
+        -------
+        T: array-like of shape (n_samples, n_classes)
+            Returns the log-probability of the sample for each class in the model,
+            where classes are ordered as they are in self.classes_.
+
+        """
+        log_prob = np.log(self.predict_proba(X))
+        # TODO: Discuss w/ Soledad
```
solegalli, line 165: remove all the commented

feature_engine/prediction/target_mean_classifier.py
```diff
@@ -0,0 +1,195 @@
...
+        X : pandas dataframe of shape = [n_samples, n_features]
+            The input series which must have the same name as one of the features in the
+            dataframe that was used to fit the predictor.
+
+        Return
+        -------
+        y_pred: ndarray of shape (n_samples,)
+            Vector containing the class labels for each sample.
+        """
+        y_pred = np.where(self.predict_proba(X) > 0.5, 1, 0)
```
solegalli, line 188: pls change line 188 for this:

`y_pred = np.where(self._predict(X) > 0.5, self.classes_[1], self.classes_[0])`

also pls remove the commented.

tests/conftest.py
```diff
@@ -215,3 +215,23 @@ def df_normal_dist():
     df.columns = ["var"]
 
     return df
+
+
+@pytest.fixture(scope="module")
+def df_enc_categorical_and_numeric():
```
solegalli, line 221: can we use df_vartypes instead of creating a new one? is this just to check the variable selection?

tests/test_prediction/test_target_mean_classifier.py
```diff
@@ -0,0 +1,130 @@
...
+        variables=None,
+        bins=6,
+        strategy="equal_width"
+    )
+
+    transformer.fit(df_pred[["Age", "Height_cm"]], df_pred["Plays_Football"])
+    accuracy_score = transformer.score(
+        df_pred_small[["Age", "Height_cm"]], df_pred_small["Plays_Football"]
+    )
+
```
solegalli, line 80: this is looking really good.

Could we also add asserts for the outputs of predict, predict_proba and predict_log_proba?

tests/test_prediction/test_target_mean_classifier.py
```diff
@@ -0,0 +1,130 @@
...
+        variables=None,
+        bins=4,
+        strategy="equal_frequency"
+    )
+
+    transformer.fit(df_pred[["Studies", "City"]], df_pred["Plays_Football"])
+    accuracy_score = transformer.score(
+        df_pred_small[["Studies", "City"]], df_pred_small["Plays_Football"]
+    )
+
```
solegalli, line 98: this is looking really good.

Could we also add asserts for the outputs of predict, predict_proba and predict_log_proba?

can we also test the transformer by passing a df with 1 variable and a df with 3 or more variables?

tests/test_prediction/test_target_mean_regressor.py
```diff
@@ -0,0 +1,172 @@
...
+        bins=5,
+        strategy="equal_frequency"
+    )
+
+    transformer.fit(df_pred[["City", "Age"]], df_pred["Marks"])
+    r2 = transformer.score(
+        df_pred_small[["City", "Age"]],
+        df_pred_small["Marks"]
+    )
+
```
solegalli, line 99: could we add an assert for the output of predict?

tests/test_prediction/test_target_mean_regressor.py
```diff
@@ -0,0 +1,172 @@
...
+        variables=None,
+        bins=3,
+        strategy="equal_width"
+    )
+
+    transformer.fit(df_pred[["Age", "Height_cm"]], df_pred["Marks"])
+    r2 = transformer.score(
+        df_pred_small[["Age", "Height_cm"]], df_pred_small["Marks"]
+    )
+
```
solegalli, line 116: could we add an assert for the output of predict?

can we also test the transformer by passing a df with 1 variable and a df with 3 or more variables?

solegalli (commented):
Hi @Morgan-Sell 

Well done with the code changes and the tests. This is looking really good.

I wonder if you could experiment with a few functions from sklearn that are used for classifiers and predictors and expand the tests a bit more.

See my comments below.

Thank you!

tests/test_variable_manipulation.py
Morgan-Sell: It seems that the conftest dataframes are either (1) not recognized if they are not imported or (2) stay as a function object.

feature_engine/prediction/base_predictor.py
Morgan-Sell: I played w/ `check_X_y`. The function returns numpy arrays which lack a dtype and cause errors downstream.

I tried transforming `X` and `y` into pandas dataframe and series after `check_X_y` was executed. This caused ` ValueError: The truth value of a DataFrame is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().` to occur for various unit tests.

I think the root of both errors exists when the MeanEncoder is applied.

Morgan-Sell:
Hola @solegalli,

Shoudl we delete the tests `test_target_mean_predictor_transformation()` and `test_regression_score_calculation_with_equal_frequency()` for both the classifier and regressor? They seem redundant now that we are testing predict, predict_proba, predict_log_proba, and score for one, two and >=3 variables.

feature_engine/prediction/base_predictor.py
Morgan-Sell: See prior post.

Morgan-Sell:
Hola @solegalli,

I haven't heard from you in a while. I hope you're busy having fun!

Do you have any thoughts on my above questions or the classes in general? Abrazo!

solegalli:
> Hola @solegalli,
> 
> I haven't heard from you in a while. I hope you're busy having fun!
> 
> Do you have any thoughts on my above questions or the classes in general? Abrazo!

Yep. Sorry. Actually I was working on a different PR. 

This one is a bit out of my comfort zone so I need to do some research to get more familiar with sklearn classifiers and regressors mixins. It'll take me a little longer.

PR #360, opened 2022-01-19
[MRG] create TimeSeriesLagTrasnformer: new feature

### LINKED ISSUE #342: time series forecasting: lag features
This would be a transformer that lags features (pandas.shift())

The transformer would lag all numerical features or those entered by the user, and automatically concatenate them to the df. 
Option to drop the original variables.

To think about: can we make multiple lags within one transformer? lag 1hr, lag2hrs etc, or would it be better to pile up 3 transformers instead?

The previous could be very easily done with a Pipeline, so maybe no need to overcomplicate the class.

```
# Move forward 24 hrs.
tmp = data[variables].shift(freq='24H')

# Rename the variables.
tmp.columns = [v + '_lag_24' for v in variables]

# Add the features to the original data.
data = data.merge(tmp, left_index=True, right_index=True, how='left')
```

Morgan-Sell:
closes #342 
closes #370 

Notes from PR #342 

This would be a transformer that lags features (pandas.shift())

The transformer would lag all numerical features or those entered by the user, and automatically concatenate them to the df.
Option to drop the original variables.

To think about: can we make multiple lags within one transformer? lag 1hr, lag2hrs etc, or would it be better to pile up 3 transformers instead?

The previous could be very easily done with a Pipeline, so maybe no need to overcomplicate the class.

```
# Move forward 24 hrs.
tmp = data[variables].shift(freq='24H')

# Rename the variables.
tmp.columns = [v + '_lag_24' for v in variables]

# Add the features to the original data.
data = data.merge(tmp, left_index=True, right_index=True, how='left')
```


Morgan-Sell:
Hola @solegalli,

A few random questions/thoughts:

1. Should the transformer confirm there is a time series for one of the dataframe columns? If so, do we ask the user to state the time-series column when instantiating the transformer? Do we autodetect the interval measurement used in the time series (I need to brainstorm how to do so.) or ask the user for the measurement's denomination?

2. Does the user state the lag's time interval - e.g., minute, hour, and day - when instantiating the class?

3. If the user claims the lag's time interval, I'm assuming we need to check that it's compatible with the stated time series. How should we define compatibility? Let's imagine that the time series is denominated in 15-minute intervals and the user selects the lag to be one hour, which is equal to four observations. Do we consider this to be compatible? Or, does "compatibility" require the denominations of both the time lag interval and time-series interval to be equivalent? Maybe it's a version 1 and 2. 

4. How do you envision multiple lags being implemented? I'm imagining a world where a user would instantiate a Pipeline with multiple TimeSeriesLagTransformer classes. Is that what you're referring to when you mentioned, `The previous could be very easily done with a Pipeline, so maybe no need to overcomplicate the class.`

5. Should we take a look at the statsmodel package for suggestions?

Lmk what I'm missing! Gracias!

solegalli:
> Hola @solegalli,
> 
> A few random questions/thoughts:
> 
>     1. Should the transformer confirm there is a time series for one of the dataframe columns? If so, do we ask the user to state the time-series column when instantiating the transformer? Do we autodetect the interval measurement used in the time series (I need to brainstorm how to do so.) or ask the user for the measurement's denomination?
> 
>     2. Does the user state the lag's time interval - e.g., minute, hour, and day - when instantiating the class?
> 
>     3. If the user claims the lag's time interval, I'm assuming we need to check that it's compatible with the stated time series. How should we define compatibility? Let's imagine that the time series is denominated in 15-minute intervals and the user selects the lag to be one hour, which is equal to four observations. Do we consider this to be compatible? Or, does "compatibility" require the denominations of both the time lag interval and time-series interval to be equivalent? Maybe it's a version 1 and 2.
> 
>     4. How do you envision multiple lags being implemented? I'm imagining a world where a user would instantiate a Pipeline with multiple TimeSeriesLagTransformer classes. Is that what you're referring to when you mentioned, `The previous could be very easily done with a Pipeline, so maybe no need to overcomplicate the class.`
> 
>     5. Should we take a look at the statsmodel package for suggestions?
> 
> 
> Lmk what I'm missing! Gracias!

1. pandas shift works with the index. It has a param, freq to detect the frequency interval. Our transformer would take pandas shift params as well, and with that, the user would be able to regulate how to move their features. I would not re-invent the wheel :p  
2. the user would instantiate pandas shift arguments, which we would bring forward to the init method of our transformer
3. it is handled by pandas shift.
4. in the param freq, if instead of a string, the user passes a list, for example ['1H', '2H', '3H'], we could loop over this list to concatenate these 3 shifts. I am still not convinced this adds value. Let's start by implementing a transformer with just 1 shift, and then we see.
5. Not at this stage.

I think a good starting point would be to read pandas shift docs and play around with that and a time series dataset with a time index or an integer index to understand how it works

Thank you!

Morgan-Sell:
Hi @solegalli,

I guess I'm just a bit confused as to why we are creating this time-series transformer given that pandas shift() exists.

Morgan-Sell:
Hi @solegalli,

I'm familiar w/ pandas shift. I guess I'm a bit confused as to why we are creating the transformer given that pandas shift() exists.

Also, should allow the user to either `periods` or `frequency` or solely `frequency` when instantiating the class?

solegalli:
@Morgan-Sell github alerted me of a question, but I can't find it. Did you delete it?

Morgan-Sell:
Hi @solegalli,

Yes, I wrote a question then answered it myself. I must be learning ;)

I'm still plugging away.

Disfruta el finde!

Morgan-Sell:
Hola @solegalli,

There seems to be a bug in the `test_check_estimator_selectors.py`. I'm not familiar with the feature-selection transformers. Do you have an idea of the cause of the bug?

Gracias!

feature_engine/timeseries/forecasting/time_lag.py
```diff
@@ -0,0 +1,173 @@
...
+    ----------
+
+    """
+
+    def __init__(
+        self,
+        variables: Union[None, int, str, List[Union[str, int]]] = None,
+        periods: int = 1,
+        freq: str = None,
+        axis: int = 0,
```
solegalli, line 70: for lagged features we are always shifting vertically, so I am not sure we need to have this as an optional value. It should probably always be 0.

feature_engine/timeseries/forecasting/time_lag.py
```diff
@@ -0,0 +1,173 @@
...
+        fill_value: [int, str] = None,
```
solegalli, line 71: I am also not sure we need this param as optional.

feature_engine/timeseries/forecasting/time_lag.py
```diff
@@ -0,0 +1,173 @@
...
+        keep_original: bool = True,
```
solegalli, line 72: normally we call this param drop_original and we default to False. Can we change? so we are compatible with our other transformers?

feature_engine/timeseries/forecasting/time_lag.py
```diff
@@ -0,0 +1,173 @@
...
+            )
+
+        # add check for 'freq' param
+
+        if axis not in (0, 1):
+            raise ValueError(
+                f"'axis' is {axis}. The variable must be 0 or 1."
+            )
+
+        if not isinstance(keep_original, bool):
```
solegalli, line 89: this is the only check we should keep, because it is not pandas.

feature_engine/timeseries/forecasting/time_lag.py
```diff
@@ -0,0 +1,173 @@
...
+            raise ValueError(
+                f"'keep_original' is {keep_original}. The variable must "
+                f"be boolean."
```
solegalli, line 92: remember to add f"Got {drop_original} instead."

feature_engine/timeseries/forecasting/time_lag.py
```diff
@@ -0,0 +1,173 @@
...
+            )
+
+        self.variables = _check_input_parameter_variables(variables)
+        self.periods = periods
+        self.freq = freq
+        self.axis = axis
+        self.fill_value = fill_value
+        self.keep_original = keep_original
+
```
solegalli, line 101: all our classes need a fit method. We should check that X is a df, and that the user either indicated numerical variables, or automatically select numerical variables. And a couple of more attributes from fit like n_features_in

feature_engine/timeseries/forecasting/time_lag.py
```diff
@@ -0,0 +1,173 @@
...
+            the original dataframe plus the transformed variables.
+
+        """
+        # check if 'df' is a dataframe
+        _is_dataframe(df)
+
+        # check if index is a datetime object
+        # Should we check more than first index value?
+        # Could use a list comprehension to check all values, but that will be expensive.
+        if not isinstance(df.index[0], datetime.datetime):
```
solegalli, line 124: pandas shift can also work with non-datetime indexes. So let's let pandas through the errors :)

feature_engine/timeseries/forecasting/time_lag.py
```diff
@@ -0,0 +1,173 @@
...
+            raise ValueError(
+                "Dataframe's index is not a datetime object. Transformer requires"
+                "the index to be a datetime object."
+            )
+
+        # check variables
+        self.variables_ = _find_all_variables(df, self.variables)
```
solegalli, line 131: this goes in fit

feature_engine/timeseries/forecasting/time_lag.py
```diff
@@ -0,0 +1,173 @@
...
+        tmp = df[self.variables_].shift(
+            periods=self.periods,
+            freq=self.freq,
+            axis=self.axis,
+            fill_value=self.fill_value
+        )
+
+        tmp.columns = self.rename_variables()
+
+        if self.keep_original:
```
solegalli, line 142: we need to add the lagged features to the original dataframe first. And then drop the original vars if needed.

feature_engine/timeseries/forecasting/time_lag.py
```diff
@@ -0,0 +1,173 @@
...
+            df = df[self.variables_].merge(
+                tmp, left_index=True, right_index=True, how="left"
+            )
+        else:
+            df = tmp.copy()
```
solegalli, line 147: I don't think this is doing what you want.

The user will pass a dataframe, maybe it has 10 variables, but they only want the lag features of 2 variables. So we need to add the lagged features to the original dataframe with the 10 vars

feature_engine/timeseries/forecasting/time_lag.py
```diff
@@ -0,0 +1,173 @@
...
+        Returns
+        -------
+        variables_lag: list
+            Names of the variables with the time-lag interval that is
+            used in the transformation.
+
+        """
+
+        if self.freq is None:
+            lag_str = f"_lag_{self.periods}pds"
```
solegalli, line 167: not sure pds is clear. Maybe just the period?

Also, the period is an integer, does this convert it to str?

solegalli (commented):
Hi @Morgan-Sell 

Great progress with the code! Thank you!

And apologies again for the silence: internet + covid as per my comment in the other PR.

I think I made a couple of comments that will simplify the code a lot. Have a look and let me know what you think.

Thank you!

feature_engine/timeseries/forecasting/time_lag.py
Morgan-Sell: Are you referring to `fill_value`?

feature_engine/timeseries/forecasting/time_lag.py
Morgan-Sell: Ok, I wasn't sure whether to keep all of the variables or only the original values of the variables that are transformed.

feature_engine/timeseries/forecasting/time_lag.py
Morgan-Sell: I wrote `pds` because I thought it would help the user delineate between time-based lags, e.g. minutes, hours, etc. These lags will be `30m`, `2h`, `1d`, etc. We could write `periods`.

And, yes, the f-string will convert `self.periods` to a string.

feature_engine/timeseries/forecasting/time_lag.py
Morgan-Sell: Should the transformer only transform numerical variables? Do we keep the categorical variables constant?

feature_engine/timeseries/forecasting/time_lag.py
Morgan-Sell: Why do we first add the lagged features to the original dataframe? If `drop_original = True` doesn't the suggested order of operations create an extra step?

feature_engine/timeseries/forecasting/time_lag.py
solegalli: yes.

feature_engine/timeseries/forecasting/time_lag.py
```diff
@@ -0,0 +1,188 @@
...
+
+        Returns
+        -------
+
+        """
+        # check if 'X' is a dataframe
+        _is_dataframe(X)
+
+        # check variables
+        self.variables_ = _find_all_variables(X, self.variables)
```
solegalli, line 118: instead of _find_all_variables, we should use _find_or_check_numerical_variables. This transformer only works with numerical variables.

feature_engine/timeseries/forecasting/time_lag.py
```diff
@@ -0,0 +1,188 @@
...
+
+        # identify numerical variables
+        self.variables_numerical_ = _find_or_check_numerical_variables(self.variables)
```
solegalli, line 121: I don't think we  need this.

feature_engine/timeseries/forecasting/time_lag.py
solegalli: the transformer ignores categorical variables. If user passes one, it would raise an error, which is already part of _find_or_check_numerical_variables functionality.

feature_engine/timeseries/forecasting/time_lag.py
```diff
@@ -0,0 +1,188 @@
...
+            periods=self.periods,
+            freq=self.freq,
+            axis=0,
+            fill_value=self.fill_value
+        )
+
+        tmp.columns = self.rename_variables()
+
+        if self.drop_original:
+            X = tmp.copy()
```
solegalli, line 157: X.drop(self.variables_, axis=1)

feature_engine/timeseries/forecasting/time_lag.py
solegalli: I would not write anything. Otherwise variable names may become too long.

feature_engine/timeseries/forecasting/time_lag.py
Morgan-Sell: Aaahhh... Soy un boludo!

Morgan-Sell:
Hola @solegalli,

I revised the TimeSeriesLagTransformer code base and the tests.

The tests for python 3.8 and 3.9 are failing. The test results state the following - `tests/test_selection/test_check_estimator_selectors.py ........Fatal Python error: Segmentation fault`.

Before I start snooping in places where I probably shouldn't, I thought I should ask whether you've seen this error before. Does it look familiar?

Abrazo!

tests/test_time_series/test_time_lag.py
```diff
@@ -0,0 +1,150 @@
...
+    (True, False, "asado", [1984, 1999]),
+]
+
+
+@pytest.mark.parametrize(
+    "_variables, _periods, _freq, _drop_original",
+    _false_input_params
+)
+def test_raises_error_when_wrong_input_params(
+        _variables, _periods, _freq, _drop_original, df_time
```
solegalli, line 18: Some of these params fail straightaway in __init__, for example `variables` and `drop_original`, and some fail during transform(), like the params that are used by pandas (period, freq).

I am not sure if we need a test, to test that pandas fails when entering the wrong params, pandas should have a check already. @KishManani what is your view on this?

Regarding our params that should fail in the init (variables and drop_original), can we test them in their own test?

Also, note that [3,2] is a valid input for variables in init. 

I suggest we either split the test in 2, one with the param values that fail in fit and one with those that fail in the init.

tests/test_time_series/test_time_lag.py
```diff
@@ -0,0 +1,150 @@
...
+        transformer3 = TimeSeriesLagTransformer(freq=_freq)
+        transformer3.fit(df_time)
+        transformer3.transform(df_time)
+    with pytest.raises(ValueError):
+        transformer4 = TimeSeriesLagTransformer(drop_original=_drop_original)
+        transformer4.fit(df_time)
+        transformer4.transform(df_time)
+
+
+def test_default_params():
```
solegalli, line 38: can we extend this test to evaluate multiple values for each parameter using parametrize? 

For variables try a string, and a list of strings and None. For periods test int, list of int and None, for freq, str, list of str and none.

solegalli (commented):
Hi @Morgan-Sell 

Thank you so much for the code and the tests. This is looking really good.

I was trying to fix the tests that fail in a different PR, and also add functions to simplify the way we write docstrings, so that we don't repeat ourselves over and over.

Those changes are relevant for this PR, so I rebased main and made a [PR to your repo](https://github.com/Morgan-Sell/feature_engine/pull/6/), which is probably why you see so many files being modified.

I also asked my experienced forecasting friend @KishManani to have a look at this PR, and see if we are not missing anything important for this class. 

@KishManani it would be great if you could have a a look at the files **after** @Morgan-Sell merges so you can have a look at the latest version of the code.

Also, I expanded the functionality so that periods and freq can take lists of inputs, so that we can add multiple lag features by calling the transformer just once (instead of piling transformers). @Morgan-Sell we would need to add tests for that functionality.

So in short if 
`transformer =LagFeatures(freq=['1H', '2H'])`

the transformed dataset would have the var_1H and var_2H features added. Does this make sense?

Thank you all!

I really love this class :p

feature_engine/timeseries/forecasting/time_lag.py
```diff
@@ -0,0 +1,174 @@
...
+    """
+
+    Parameters
+    ----------
+    variables: list, default=None
+        The list of numerical variables to transform. If None, the transformer
+        will automatically select all numerical variables.
+
+    periods: int, default=1
+        Number of periods to shift. Can be positive or negative.
```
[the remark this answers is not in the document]
KishManani: I cannot think of a valid use case where you would use a negative lag so I agree with your first point. 

fill-value fills only the NaN values introduced at the start of the series, in particular when you use the `period` argument and do not set the `freq` parameter. The current implementation you have here allows you to shift multiple columns at the same time, in that case you probably wouldn't want to impute the missing values introduced at the start of the time series in different columns with the same value. But the current Pandas API does allow the user to do this. If you wanted to lag one column and also fill the missing values at the beginning then it could be convenient to expose `fill-value`. In practice you'd likely use the `fillna` method of a dataframe use backfilling.

tests/test_time_series/test_time_lag.py
KishManani: I agree that a test here is not necessary for the reasons you've provided.

Morgan-Sell: Ok, so the plan is to only test `variables` and `drop_original`? In that case, should we delete the `fit()` and `transform()` methods?

tests/test_time_series/test_time_lag.py
Morgan-Sell: How do we test for multiple values when we are testing the default params? There's only one default value per param.

solegalli: I would say, let's change the test name and test the default paramaters and other permitted values for the params as well.

feature_engine/timeseries/forecasting/time_lag.py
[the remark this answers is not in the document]
solegalli: @Morgan-Sell 

to test the variables param, we have now a generic test that we can apply to all transformers. The test is [here](https://github.com/feature-engine/feature_engine/blob/main/feature_engine/estimator_checks.py#L102).

We also have [this other test](https://github.com/feature-engine/feature_engine/blob/main/feature_engine/estimator_checks.py#L31) that runs the above and other multiple tests that we need to run on all our transformers.

And here you have [an example](https://github.com/feature-engine/feature_engine/blob/main/tests/test_selection/test_check_estimator_selectors.py#L46) on how to run those tests for all selector transformers. For this module, at the moment we have only 1 class, but we will have more. We should do the same.

In short, we need a script like the test_check_estimator_selectors for the forecasting module, that runs the check_estimator test from sklearn (first test in the script) and the check_estimator from feature engine (second test in this script).

If you are feeling dareful, we could also create a check for the param drop_original in the estimator_checks and run it from the master check scropt. Because other transformers also have this parameter, and would benefit from this test.

tests/test_time_series/test_time_lag.py
solegalli: @Morgan-Sell 

to test the variables param, we have now a generic test that we can apply to all transformers. The test is [here](https://github.com/feature-engine/feature_engine/blob/main/feature_engine/estimator_checks.py#L102).

We also have [this other test](https://github.com/feature-engine/feature_engine/blob/main/feature_engine/estimator_checks.py#L31) that runs the above and other multiple tests that we need to run on all our transformers.

And here you have [an example](https://github.com/feature-engine/feature_engine/blob/main/tests/test_selection/test_check_estimator_selectors.py#L46) on how to run those tests for all selector transformers. For this module, at the moment we have only 1 class, but we will have more. We should do the same.

In short, we need a script like the test_check_estimator_selectors for the forecasting module, that runs the check_estimator test from sklearn (first test in the script) and the check_estimator from feature engine (second test in this script).

If you are feeling dareful, we could also create a check for the param drop_original in the estimator_checks and run it from the master check scropt. Because other transformers also have this parameter, and would benefit from this test.

feature_engine/timeseries/forecasting/time_lag.py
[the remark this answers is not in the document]
Morgan-Sell: All done ;)

tests/test_time_series/test_time_lag.py
Morgan-Sell: I am feeling intrepid!  I'll embark on this adventure later ;)

Morgan-Sell: Ok, well, it seems like someone already developed the sklearn and feature-engine check_estimator tests.

A couple of questions about the `drop_original` param check:

- Are you envisioning a check function that assesses whether `drop_original` is a boolean variable? Is it doing anything else?
- What/where is the master check script?

feature_engine/timeseries/forecasting/lag_features.py
```diff
@@ -0,0 +1,292 @@
...
+        tmp.columns = self.get_feature_names_out()
+
+        X = X.merge(tmp, left_index=True, right_index=True, how="left")
+
+        if self.drop_original:
+            X = X.drop(self.variables_, axis=1)
+
+        return X
+
+    def get_feature_names_out(self, input_features=None):
```
KishManani, line 240: If this method is only meant to be used internally when .predict() is called should this be a private method? Is there a use case where a user would want to run this method?

feature_engine/timeseries/forecasting/lag_features.py
```diff
@@ -0,0 +1,292 @@
...
+                for feature in input_features
+            ]
+        elif self.freq is not None:
+            feature_names = [
+                str(feature) + f"_lag_{self.freq}" for feature in input_features
+            ]
+        elif isinstance(self.periods, list):
+            feature_names = [
+                str(feature) + f"_lag_{pr}"
+                for pr in self.perdiods
```
KishManani, line 274: Typo here which raises an AttributeError. Should be `periods` rather than `perdiods`.

feature_engine/timeseries/forecasting/lag_features.py
```diff
@@ -0,0 +1,292 @@
...
+            df_ls = []
+            for fr in self.freq:
+                tmp = X[self.variables_].shift(
+                    freq=fr,
+                    axis=0,
+                )
+                df_ls.append(tmp)
+            tmp = pd.concat(df_ls, axis=1)
+
+        elif isinstance(self.periods, list):
```
KishManani, line 214: The user is still able to pass a list for both `periods` and `freq`.  With this implementation the list passed to `periods` is ignored and the list passed to `freq` takes precedence. 

`LagFeatures(periods=[1, 2, 3, 4], freq=['15T','30T']).fit(df).transform(df)`

Still returns a valid output despite the fact that incompatible inputs have been provided. 

There needs to be an agreement as to what the behaviour should be if a list is passed to both `periods` and `freq`. The most reasonable options imo would be:

1. A cartesian product of the two options (so you generate `len(periods)*len(freq)` new lagged features)
2. You require that the length of the lists for both parameters are the same and mimic the same behaviour as for when you do not provide a list of arguments (i.e., you map the first element in `periods` and you create a lag by the amount specified in the first element in `freq` and so on). 
3. Raise an error and don't allow a list to be passed for both the `freq` and `periods` argument.

@solegalli - what do you think? I think option 2 or 3 is probably the more intuitive and expected behaviour.

feature_engine/timeseries/forecasting/lag_features.py
KishManani: Also suggests that tests are needed where the input arguments are lists and combinations of lists and ints.

KishManani: I have corrected the typo in this PR: https://github.com/feature-engine/feature_engine/pull/370

However, we still probably need to add some additional tests as per the comment above.

feature_engine/timeseries/forecasting/lag_features.py
solegalli: Sklearn has rolled out this method for all feature extractors and it can be accessed through the pipeline. 

It is meant to return all the features of the dataset if input_features=None, or the features that were created from a certain feature or features if input_features is a list.
https://scikit-learn.org/stable/glossary.html#term-get_feature_names_out

(I think our function here does not do that correctly)

We already got 2 issues from 2 different users requesting this functionality. Plus we should roll it out to adhere to sklearn conventions. We have another PR to roll this out in other transformers: #298

feature_engine/timeseries/forecasting/lag_features.py
solegalli: @KishManani good point. 

I think the current implementation does not do both. Because it says **if** freq is list, do this, **elif** period is list do that.

I was mostly inclined to give the param `freq` priority over `periods`. So, if you pass a list to freq, we ignore periods (and mention that in the docs). Mostly because I do not see a case where you would want both. If you pass the frequency, you kind of know very well what you want, so why would you pass a list of freq and a list of periods?

But then, I haven't lagged a lot of features myself (lol), so if you can see a case where a user would want to pass both, happy to change.

For simplicity, we could also go with 3, because this class would already simplify the user's life a lot, so if they want features with a list of freq and a list of periods, they may as well call the transformer twice, once for each.

@KishManani thoughts?

tests/test_time_series/test_time_lag.py
solegalli: yes, that was me, lol.

the check needs to check that if it is set to false, the original variables are retained, and if set to true, the original variables are dropped after the transformation.

I wonder if we should also check within the same test, that it takes the allowed param. have a think and let me know your views :)

Morgan-Sell: Ok, I understand the desired test. Are you thinking that we use include that test in `test_different_types_for_init_params()`?

If so, it seems better to create its own test b/c `test_different_types_for_init_params()` does not require `fit()` and `transform()`. I assume it would be cleaner for someone to understand what's occurring if it were a separate unit test.

feature_engine/timeseries/forecasting/lag_features.py
KishManani: @solegalli thanks for highlighting the comment in the docstring. I missed that. It says the following in the docstring `If freq is not None, then this parameter overrides the parameter periods`. I don't think this is true. The current implementation only overrides `periods` when `freq` is a `list`. For example the following:

`LagFeatures(periods=3, freq='2D').fit(df).transform(df)`

would shift the index by 6 days (which is also the expected behaviour and the same as what you get with `.shift()`. This is an example where `freq` is not `None` and did not override the `periods` parameter. 

Given the above example there is a way in which it could be sensible to have a `list` for both `periods` and `freq`. But in my opinion this is redundant because we could have just provided in the example above `freq='6D'`. 

Likewise the following example could generate confusion:

`LagFeatures(periods=[1,2,3], freq='D').fit(df).transform(df)`

We might expect this to lag by 1 day, 2 days, and 3 days. But the `freq` parameter would be completely ignored in the current implementation in this example.

A simpler solution might be to restrict the user to only allow them to provide inputs for `periods` or `freq` but not both and raise an error if this is tried. If the user wanted to lag by 3 days they could just use `freq='3D'` rather than `periods=3` and `freq='D'`.

solegalli: Enlightening. I did not know about this. That's why we need the expert :) Thank you @KishManani 

I agree and 'll fix accordingly.

Morgan-Sell: I implemented the check that raises an error if both `periods` and `freq` have values. I also built a test for the corresponding check.

I changed the `periods` default value to `None`. It was `1`. Otherwise, an error is raised whenever the user solely inputs a value for `freq`. It seems that this would be a common practice.

feature_engine/timeseries/forecasting/lag_features.py
```diff
@@ -0,0 +1,313 @@
...
+        if input_features is None:
+            # Create all lag features.
+            input_features_ = self.variables_
+        else:
+            if not isinstance(input_features, list):
+                raise ValueError(
+                    f"input_features must be a list. Got {input_features} instead."
+                )
+            # Create just indicated lag features.
+            input_features_ = input_features
```
solegalli, line 268: Add a control to check that the user entered variable belongs to self.variables_

Otherwise, this function will output names for lag features that were not created. For example if user passes a string with categorical variables.

solegalli:
@KishManani 

I think we've got the main functionality now. Would you like to have a final look?

feature_engine/timeseries/forecasting/lag_features.py
solegalli: @Morgan-Sell  I discovered a bug here, so I will probably make another PR to your repo tomorrow :/

docs/user_guide/timeseries/forecasting/LagFeatures.rst
```diff
@@ -0,0 +1,257 @@
...
+
+.. currentmodule:: feature_engine.timeseries.forecasting
+
+LagFeatures
+===========
+
+The :class:`LagFeatures` adds lag features to the dataframe. A lag feature is a feature
+with information about a prior time step.
+
+In forecasting, past values of the variable we want to forecast, are likely to be
```
KishManani, line 11: Suggest change from:

> In forecasting, past values of the variable we want to forecast, are likely to be predictive. 

to

> When forecasting the future values of a time series the past values of that time series are likely to be predictive. 

Reason: Grammatically (I think the comma usage in the first sentence is wrong) correct and clearer imo.

feature_engine/timeseries/forecasting/lag_features.py
```diff
@@ -0,0 +1,313 @@
...
+
+        # Check if input data contains same number of columns as dataframe used to fit.
+        _check_input_matches_training_df(X, self.n_features_in_)
+
+        # check if dataset contains na
+        if self.missing_values == "raise":
+            _check_contains_na(X, self.variables_)
+            _check_contains_inf(X, self.variables_)
+
+        # if freq is not None, it overrides periods.
```
KishManani, line 195: Is the change to ensure that if a list is passed to both `freq` and `periods` that this method raises error happening in a different PR?

solegalli: We are not raising an error. The logic should just not use both and ensures that if freq is given, either list or string, it will override periods. And we mention that in the docstrings above.

solegalli:
@KishManani is the review done?

I've got from you the change in the wording in the docs, and the question on the error, which we are not raising. 

From my side, I need to add a variable check in the get_feature_names_out.

Is that all @KishManani or do you need more time?

Thank you!

KishManani:
Another minor comment. If you pass a series such as this:

`LagFeatures(variables=['y'], freq=['1MS', '2MS', '3MS']).fit_transform(df['y'])`

it will raise a TypeError because df['y'] is a pandas Series rather than a Dataframe. Do you want to be able to support passing a Series like this and returning a dataframe of lags?

solegalli:
@Morgan-Sell @KishManani 

I went on and enforced dataframes to have indexes with unique values and no NaN to be compatible with this transformer.
Otherwise we duplicate rows when we merge.

I also added an argument in the init, to give the option to the user to sort the index or not before lagging the features.

I then added tests and updated the docstrings.

Are we good to go?

Morgan-Sell:
@Morgan-Sell @KishManani 

Wow! You two have been having a lot of fun w/o me :(

I like the idea of requiring the user to provide unique time intervals. I worked in renewable energy and dealt with a lot of time-series data (I wasn't using Python) both energy production and wholesale electricity markets. 

I would like to know if the dataset was comprised of duplicate time intervals.  I cannot envision a scenario in which I would've developed a model/anlayis using duplicate time intervals. In the case of energy production, the time-interval duplication would inform me that there may be something wrong w/ the SCADA - the system that collects energy generation and other relevant data.

feature_engine/timeseries/forecasting/lag_features.py
```diff
@@ -0,0 +1,370 @@
...
+
+        # Check that the index contains unique values.
+        if X.index.is_unique is False:
+            raise NotImplementedError(
+                "The dataframe's index does not contain unique values. "
+                "Only dataframes with unique values in the index are compatible "
+                "with this transformer."
+            )
+
+        if X.index.isnull().sum() > 0:
```
solegalli, line 178: this check needs to come before the check for unique values.

Morgan-Sell:
@solegalli @KishManani 

Is the PR completed?

solegalli:
Oh, yes it is, big time!

Thank you guys, merging now.

@Morgan-Sell @KishManani

PR #369, opened 2022-02-01
add stacking selector functionality and docstrings

### LINKED ISSUE #309: Adjust variables when piling up selectors
The variables argument works very well when having a single selector but then "piling up" selectors in a pipeline it does not account for the features removed in the previous steps.

For example if I want to have the pipeline running on only variables starting with "f_" I cannot run the following code because "f_A" is not defined anymore at the second step.

image

Removing the variables in the second selector does not help as the feature called 'E' is also removed

image

I believe a selection at the beginning of the fit method would allow this feature. Something like:
```
if self.variables:
      self.variables = [var for var in self.variables if car in X.columns]
``` 

**Describe alternatives you've considered**
The alternative is to make the selection at the entry of the pipeline. Although, to me, this is not optimal as most dataframes do not only contain the features but also additional information needed further (could be a target or groups not used for modelling but for analysis).

solegalli:
closes #361 
closes #309 

Hi @gverbock 

I rebased main to solve the pytest errors and add docstring functionality and then cherry picked some of the changes in #361.

I still need to cherry-pick the test changes, and we need to add the confirm_variables functionality to the remaining selectors (except the target mean because we need to fix that in a different PR)


solegalli:
I think I brought all the changes in PR #361 to this one now.

solegalli:
Hi @gverbock 

It would be great if you can pick up this changes in a new branch and then make a PR to the branch here, `stack_sels`. 

Also, since all classes have the functionality for `confirm_variables`, instead of adding 1-2 tests per class, can we add 1 generic test in the script `check_estimator.py` at the root of the package and then call that test from [check_estimator_selectors](https://github.com/feature-engine/feature_engine/blob/main/tests/test_selection/test_check_estimator_selectors.py) ?

gverbock:
OK. I see that you are indeed quite far in solving the issues. I will make a PR to this branch.

solegalli:
> * not sure if there is somethi

Hi @gverbock 

I am not sure why I did not receive a notification for your last message. I am really sorry I reply only today :_(

I believe I have now added all the necessary changes. Needs a final review to merge.

solegalli:
Still needs revision:

- remove tests for confirm variables in individual test files (we now have a generic test in estimator_checks.py.
- see if we can add a test in check_confirm_variables in estimator_checks.py initialising the transformers with variables=None.

I think that will be all.

PR #373, opened 2022-02-08
Add tests for confirm_variables

gverbock:


gverbock:
@solegalli  I created this PR to stack_sels to work on the tests for `confirm_variable` (with a first test push).  
Would it be handy to close #361 because if I got it right, it has been fully replaced by #369 and this is a PR for the latter.

solegalli:
> @solegalli I created this PR to stack_sels to work on the tests for `confirm_variable` (with a first test push). Would it be handy to close #361 because if I got it right, it has been fully replaced by #369 and this is a PR for the latter.

It will close automatically when we merge #369

PR #372, opened 2022-02-08
adds get_feature_names_out() method and feature_names_in_ attribute

solegalli:
closes #298 
closes #316 
closes #317 
closes #364 
closes #366 
closes #371 
closes #374 

**Done**

- transformation
- discretization

**TODO**

- [x] encoders
- [x] imputation
- [x] creation - unify variable creation procedure
- [x] datetime - common checks
- [x] outliers
- [x] preprop
- [x] selection
- [x] wrappers
- [x] after #360 to unify some files and remove TODOs
- [x] finish logic RelativeFeatures
- [x] complete documentation new creation transformers
- [x] add new transformers in documentation index
- [x] after #358 and update TargetMeanSelector
- [x] check documentation builds and reads well
- [x] add changes in what's new section for next release

Also in this PR:

- rollout of generic tests for all transformers
- decision tree discretizer now accepts cv constructible
- improve sklearntransformerwrapper functionality (with supported transformers)
- Sklearntransformerwrapper now has inverse_transform method
- deprecates creators and adds new classes

In #390, merged onto this branch:

closes https://github.com/feature-engine/feature_engine/issues/227
closes https://github.com/feature-engine/feature_engine/issues/238
closes https://github.com/feature-engine/feature_engine/issues/296
closes https://github.com/feature-engine/feature_engine/pull/334
closes https://github.com/feature-engine/feature_engine/pull/358

FYI @janrito @hectorpatino

PR #374, opened 2022-02-12
add inverse_transform functionality to sklearnTransformerWrapper

### LINKED ISSUE #371: add inverse_transform functionality to sklearnTransformerWrapper
Please add inverse_transform functionality to the sklearnTransformerWrapper .

Morgan-Sell:
Closes #371 

Notes from #371:

The sklearn transformer wrapper wraps an sklearn estimator to apply its functionality to a subset of variables and return a dataframe (instead of the default numpy array of sklearn).

We need to add the methods inverse_transform and get_feature_names out to this class.

Those methods would do whatever the sklearn transformer that is being wrapped does. So basically, we would be calling the methods from the sklearn transformer.

The thing is, not every sklearn transformer has those methods, and we don't know which transformer the user wants to wrap. So in the method we implement here we should check if the sklearn transformer has a callable inverse_transform or get_feature_names_out, then we return the method of the estimator. If not, we return a print statement, saying "the transformer does not have this method implemented" or something like that.

### DIFF
--- a/feature_engine/wrappers/wrappers.py
+++ b/feature_engine/wrappers/wrappers.py
@@ -129,10 +129,8 @@ class SklearnTransformerWrapper(BaseEstimator, TransformerMixin):
 
     def inverse_transform(self, X: pd.DataFrame) -> pd.DataFrame:
         """
-        Apply the transformation to the dataframe.
+        Apply the reverse the transformation of a dataframe.
         Only the selected features will be modified.
-
-
 
         Args:
             X: Pandas DataFrame to perform desired transformation
@@ -143,28 +141,22 @@ class SklearnTransformerWrapper(BaseEstimator, TransformerMixin):
         # check that input is a dataframe
         X = _is_dataframe(X)
 
-        # Check that input data contains same number of columns as
-        # the dataframe used to fit the imputer.
-        _check_input_matches_training_df(X, self.input_shape_[1])
+        # A OneHotEncoder-transformed dataframe has the original variables.
+        if isinstance(self.transformer, OneHotEncoder):
+            X = X[self.variables]
 
-        try:
-            if isinstance(self.transformer, OneHotEncoder):
-                ohe_results_as_df = pd.DataFrame(
-                    data=self.transformer.inverse_transform(X[self.variables]),
-                    columns=self.transformer.get_feature_names(self.variables),
-                )
-                X = pd.concat([X, ohe_results_as_df], axis=1)
+        else:
+            # Check that input data contains same number of columns as
+            # the dataframe used to fit the imputer.
+            _check_input_matches_training_df(X, self.input_shape_[1])
+            X[self.variables] = (self
+                                 .transformer
+                                 .inverse_transform(X[self.variables])
+                                 )
+        return X
 
-            else:
-                X[self.variables] = (self
-                                     .transformer
-                                     .inverse_transform(X[self.variables])
-                                     )
-        except:
 
-        pass
-
-    def get_feature_names(self, X: pd.DataFrame) -> list:
+def get_feature_names(self, X: pd.DataFrame) -> list:
         """
 
 
--- a/feature_engine/wrappers/wrappers.py
+++ b/feature_engine/wrappers/wrappers.py
@@ -129,14 +129,18 @@ class SklearnTransformerWrapper(BaseEstimator, TransformerMixin):
 
     def inverse_transform(self, X: pd.DataFrame) -> pd.DataFrame:
         """
-        Apply the reverse the transformation of a dataframe.
+        Reverse the transformation of a dataframe.
         Only the selected features will be modified.
 
-        Args:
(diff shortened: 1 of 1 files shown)


Morgan-Sell:
Hi @solegalli,

Why does the `transform` method make an exception for `OneHotEncoder`?

solegalli:
Hi @Morgan-Sell 

Point 1: I don't recall

Point 2: yes. We should probably extend the fit and transform method, because atm I don't think this is working with PolynomialFeatures or MissingIndicator for example.

Point 2, again: I would probably leave the Tif transformer out of scope for this wrapper and make a separate wrapper for that.

Point 3: I would prefer not to use try\except. Just `if is_callable()`

Point 4: why do we need to rename variables?

Point 5: makes sense. I haven't really thought of it. We need to do some research and testing. I leave that with you.

Point 6: sure thanks!

Morgan-Sell:
2) Do you have a prefer approach in resolving the issue for PolynomialFeatures() and MissingIndicator()? 

It seems that the error is generated by `X[self.variables] = self.transformer.transform(X[self.variables])` because these transformers return an array with a different shape. Should we should `isinstance` for these two transformers? If the transformer is PolynomialFeatures() or MissingIndicator(), one solution is to overwrite `X= self.transformer.transform(X[self.variables])`.

3) Do you mean `if callable()`? If so, I'm not too familiar with this built-in function. I see that it should return if the class has a `__call__` method or of a type that has a which indicates callability such as in functions, methods, etc?

Why does the following code return `False` given that `inverse_transform()` is compatible with the StandardScaler() class?

```
sub = df[["Age", "Marks"]].copy()
scaler = StandardScaler()
scaler.fit(sub)
df_scaled = scaler.transform(sub)
callable(scaler.inverse_transform(df_scaled))
```

4) Given that the variables are transformed, they are technically not the same as the prior variable and I thought it could helpful to remind the user. It was just a thought. It seems the best approach is to be consistent with other Python libraries.

solegalli:
2) We should expand [line 118](https://github.com/feature-engine/feature_engine/blob/af151406cb692f04448d5fdf3ae5a4fa36f23181/feature_engine/wrappers/wrappers.py#L117) to include polynomialfeatures and missing indicator.

3) for the use of callable check the second answer of this [stackoverflow thread](https://stackoverflow.com/questions/25295327/how-to-check-if-a-python-class-has-particular-method-or-not)

feature_engine/wrappers/wrappers.py
```diff
@@ -125,3 +125,61 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+        Returns:
+        X_new : pandas dataframe of shape = [n_samples, n_features]
+
+        """
+        # check that input is a dataframe
+        X = _is_dataframe(X)
+
+        # A OneHotEncoder-transformed dataframe has the original variables.
+        if isinstance(self.transformer, OneHotEncoder):
+            X = X[self.variables]
```
solegalli, line 150: self.variables may be a subset of the variables. We want to drop the features that are output of get_feature_names

feature_engine/wrappers/wrappers.py
```diff
@@ -125,3 +125,61 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+            A dataframe that has already been transformed using a
+            sklearn transformer.
+
+        Returns:
+        X_new : pandas dataframe of shape = [n_samples, n_features]
+
+        """
+        # check that input is a dataframe
+        X = _is_dataframe(X)
+
```
solegalli, line 147: in the inverse_transform we want to check callable(self.estimator.transform) . If it exists, we apply it. If not, we return a Not implemented error

feature_engine/wrappers/wrappers.py
```diff
@@ -125,3 +125,61 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+            # the dataframe used to fit the imputer.
+            _check_input_matches_training_df(X, self.input_shape_[1])
+            X[self.variables] = (self
+                                 .transformer
+                                 .inverse_transform(X[self.variables])
+                                 )
+        return X
+
+
+def get_feature_names(self, X: pd.DataFrame) -> list:
```
solegalli, line 163: we dont need this

feature_engine/wrappers/wrappers.py
```diff
@@ -125,3 +125,61 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+        Args:
+            X: Pandas DataFrame to perform desired transformation
+
+        Returns:
+            List of column names
+        """
+
+        pass
+
+    def _check_if_sklearn_method_exist(self):
```
solegalli, line 177: we dont need this

solegalli:
@Morgan-Sell 

heads-up: I am adding the functionality discussed here in PR #372 

Will be closing this PR soon.

PR #380, opened 2022-02-25
time series forecasting: window features

### LINKED ISSUE #343: time series forecasting: window features
The transformer should create computations over windows of past values of the features, and populate them at time t, t being the time of the forecast.

It uses pandas rolling, outputs several comptutations, mean, max, std, etc, and pandas shift to move the computations to the right row.

```
tmp = (data[variables]
       .rolling(window='3H').mean()  # Average the last 3 hr values.
       .shift(freq='1H')  # Move the average 1 step forward
       )

# Rename the columns
tmp.columns = [v + '_window' for v in variables]

data = data.merge(tmp, left_index=True, right_index=True, how='left')
```

Morgan-Sell:
Closes #343. 

Notes from #343:

The transformer should create computations over windows of past values of the features, and populate them at time t, t being the time of the forecast.

It uses pandas rolling, outputs several comptutations, mean, max, std, etc, and pandas shift to move the computations to the right row.

```
tmp = (data[variables]
       .rolling(window='3H').mean()  # Average the last 3 hr values.
       .shift(freq='1H')  # Move the average 1 step forward
       )

# Rename the columns
tmp.columns = [v + '_window' for v in variables]

data = data.merge(tmp, left_index=True, right_index=True, how='left')
```

### DIFF
--- a/feature_engine/timeseries/forecasting/window_features.py
+++ b/feature_engine/timeseries/forecasting/window_features.py
@@ -1,0 +1,41 @@
+# Authors: Morgan Sell <morganpsell@gmail.com>
+# License: BSD 3 clause
+
+from typing import List, Optional, Union
+
+import pandas as pd
+from sklearn.base import BaseEstimator, TransformerMixin
+from sklearn.utils.validation import check_is_fitted
+
+from feature_engine.dataframe_checks import (
+    _check_contains_inf,
+    _check_contains_na,
+    _check_input_matches_training_df,
+    _is_dataframe,
+)
+from feature_engine.docstrings import (
+    Substitution,
+    _drop_original_docstring,
+    _feature_names_in_docstring,
+    _fit_not_learn_docstring,
+    _fit_transform_docstring,
+    _missing_values_docstring,
+    _n_features_in_docstring,
+    _variables_numerical_docstring,
+)
+from feature_engine.validation import _return_tags
+from feature_engine.variable_manipulation import (
+    _check_input_parameter_variables,
+    _find_or_check_numerical_variables,
+)
+
+
+@Substitution(
+    variables=_variables_numerical_docstring,
+    missing_values=_missing_values_docstring,
+    drop_original=_drop_original_docstring,
+    feature_names_in_=_feature_names_in_docstring,
+    n_features_in_=_n_features_in_docstring,
+    fit=_fit_not_learn_docstring,
+    fit_transform=_fit_transform_docstring,
+)


Morgan-Sell:
@solegalli,

For the above code to work, `variables` must be a list. Otherwise, we cannot add the `_window` suffix to the transformed variable because df[variables] is a pandas series. How would you like to handle it?

- The user could be required to submit a list.
- In the fit(), we could check if `self.variables` is a list type, if not, we make change the type to a list.

Is there a case where variables should be allowed to be an integer?

solegalli:
Hi @Morgan-Sell 

Pleas check our variable_manipulation functions. They've got what you need.

solegalli:
> 

Yes, we leave all the checks to pandas.

feature_engine/timeseries/forecasting/window_features.py
```diff
@@ -0,0 +1,333 @@
...
+
+        self.variables = _check_input_parameter_variables(variables)
+        self.window = window
+        self.function = function
+        self.freq = freq
+        self.sort_index = sort_index
+        self.missing_values = missing_values
+        self.drop_original = drop_original
+
+    def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
```
solegalli, line 162: the fit method is identical to lag features.

Could we move this to a base forecasting class? for this transformer and the lag features?

feature_engine/timeseries/forecasting/window_features.py
```diff
@@ -0,0 +1,333 @@
...
+                "with this transformer."
+            )
+
+        # check if dataset contains na
+        if self.missing_values == "raise":
+            _check_contains_na(X, self.variables_)
+            _check_contains_inf(X, self.variables_)
+
+        if self.sort_index is True:
+            X.sort_index(inplace=True)
```
solegalli, line 253: up to this line, the transform method is identical to lag features, maybe we should make it part of a base class.  together with the fit method

feature_engine/timeseries/forecasting/window_features.py
```diff
@@ -0,0 +1,333 @@
...
+
+        # if freq is not None, it overrides periods.
+        if self.freq is not None:
+            tmp = (X[self.variables_]
+                   .rolling(window=self.window).apply(self.function)
```
solegalli, line 258: @KishManani

the param windows takes strings and integers. Is this enough? should we consider anything else?

feature_engine/timeseries/forecasting/window_features.py
```diff
@@ -0,0 +1,333 @@
...
+        tmp.columns = self.get_feature_names_out(self.variables_)
+
+        X = X.merge(tmp, left_index=True, right_index=True, how="left")
+
+        if self.drop_original:
+            X = X.drop(self.variables_, axis=1)
+
+        return X
+
+    def get_feature_names_out(self, input_features: Optional[List] = None) -> List:
```
solegalli, line 277: Is this method identical to lag features? if yes, then let's move it to the base class.

tests/test_time_series/test_forecasting/test_window_features.py
```diff
@@ -0,0 +1,227 @@
...
+    assert transformer.freq == _freqs
+
+
+@pytest.mark.parametrize("_freqs", [5, "asado", [1, 2, 4], False])
+def test_error_when_non_permitted_param_freq(_freqs):
+    with pytest.raises(ValueError):
+        WindowFeatures(periods=None, freq=_freqs)
+
+
+def test_get_feature_names_out(df_time):
```
solegalli, line 56: is this test identical to the one for lag features? if yes, let's move it to the common tests file, and pass the transformer as an argument with parametrize.

The common test file is where we check the tests from check_estimator.

feature_engine/timeseries/forecasting/window_features.py
```diff
@@ -0,0 +1,333 @@
...
+        See Also
+        --------
+        pandas.shift
+        """
+
+    def __init__(
+            self,
+            variables: List[str] = None,
+            window: Union[str, int] = 1,
+            function: Callable = np.mean,
```
solegalli, line 112: I have two questions for you guys @Morgan-Sell  and @KishManani

First, in other transformers I call this parameter "operations" instead of "functions". I'd like to unify the way we call parameters that do similar things throughout the package. So my question is: which is a better name for this parameter? "operations", "functions", something else, or it does not really matter? If it is not important, I would prefer "operations" so I don't have to deprecate in the other transformer.

And second, I see that @Morgan-Sell is passing numpy functions here. But pandas agg() takes strings. So, should this parameter take strings, numpy functions, or we don't care because in any case, pandas takes care of this check? and we just allow everything?

Quick note: whenever functionality is already implemented in packages that have great developer support, like pandas, sklearn and scipy, I tend to leave "the work" to them (that is, their methods and classes), because we are a tiny community in comparison, so maintaining for us is harder.

Thank you!

feature_engine/timeseries/forecasting/window_features.py
```diff
@@ -0,0 +1,333 @@
...
+            periods: int = 1,
+            freq: str = None,
+            sort_index: bool = True,
+            missing_values: str = "raise",
+            drop_original: bool = False,
+    ) -> None:
+
+        if not (isinstance(window, int) or isinstance(window, str)) or isinstance(window, bool):
```
solegalli, line 120: Like in my previous comment, I would leave all checks of parameters that will be used by pandas, to pandas. That is, periods, freq and window.

feature_engine/timeseries/forecasting/window_features.py
```diff
@@ -0,0 +1,333 @@
...
+            raise ValueError(
+                f"function must be np.mean, np.std, np.median, np.min, np.max, or np.sum. "
+                f"Got {function} instead."
+            )
+        if periods is None and not isinstance(freq, str):
+            raise ValueError(
+                f"freq must be a string. Got {freq} instead."
+            )
+
+        if missing_values not in ["raise", "ignore"]:
```
solegalli, line 142: this parameter and the next one are also in lag features. I would make them part of a base class, see my next comment.

solegalli (commented):
Hi @Morgan-Sell 

Thank you so much for such an incredible amount of work!

This is looking really good. Thank you so much.

I made a few comments that are mainly organizational. 

Much of the code in this class is identical to LagFeatures, so I think we should move the bits that are similar to a base forecasting class, because it will be easier to maintain.

The second is regarding reorganizing the tests. Those tests that are similar for window and lag features, should also be moved to a common file we already have in the forecasting test folder. 

Would you mind having a look at this?

Thanks again!

feature_engine/timeseries/forecasting/window_features.py
Morgan-Sell: It's different. It includes the window in the name, e.g. var_A_window_2_freq_3d and var_B_window_3_periods_2.

I guess we could move the following code to the BaseForecast class:

```
check_is_fitted(self)

        # create names for all window features or just the indicated ones.
        if input_features is None:
            input_features_ = self.variables_
        else:
            if not isinstance(input_features, list):
                raise ValueError(
                    f"input_features must be a list. Got {input_features} instead."
                )
            if any([f for f in input_features if f not in self.variables_]):
                raise ValueError(
                    "Some features in input_features were not transformed. This method only "
                    "provides the names of the transform features with this method."
                )
            # create just indicated window features
            input_features_ = input_features
```

tests/test_time_series/test_forecasting/test_window_features.py
Morgan-Sell: It's different because "window_X" is included in the transformed variables. See the comment above.

feature_engine/timeseries/forecasting/window_features.py
Morgan-Sell: Just to confirm, are  `missing_values` and `drop_original` the only init params to be included in BaseForecast init?

Morgan-Sell:
hola @solegalli,

Are you envisioning the following tests to be moved to the test_check_estimator_forecasting.py?

- test_permitted_param_periods
- test_error_when_non_permitted_param_periods
- test_permitted_param_freq
- test_error_when_non_permitted_param_freq
- test_sort_index

The other tests involve `window`, `function`, and feature names. As of now, the variables that are transformed using WindowFeatures include the suffix `window_X`.

Morgan-Sell:
@solegalli @KishManani 

A few questions regarding the common forecasting tests:

- WindowFeatures is failing the `test_check_estimator_from_feature_engine` test in test_check_estimator_forecasting.py. It seems to be because WindowFeatures does not return an error when `check_error_if_y_not_passed`. Is this test appropriate for WindowFeatures? The `y` variable is optional in the `fit()`. 
- What is causing `sklearn.utils.check_estimators` to skip certain tests for LagFeatures? Why doesn't the same occur for WindowFeatures?

KishManani:
I would also include an option to allow expanding windows: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.expanding.html

feature_engine/timeseries/forecasting/window_features.py
KishManani: I think it covers most use cases. However, in Pandas there's also the option to provide a custom class to define windows (see Pandas rolling docs (here)[BaseIndexer subclass](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.rolling.html]). If the effort required to support it is high, it may not be worth it and can be added in the future.

feature_engine/timeseries/forecasting/window_features.py
```diff
@@ -0,0 +1,333 @@
...
+        y: pandas Series, default=None
+            y is not needed in this transformer. You can pass None or y.
+        """
+        # check input dataframe
+        X = _is_dataframe(X)
+
+        # We need the dataframes to have unique values in the index and no missing data.
+        # Otherwise, when we merge the window features we will duplicate rows.
+
+        if X.index.isnull().sum() > 0:
```
KishManani: Is there any reason for doing this over this:

```
if X.index.isnull().any():
```

if not I believe my suggestion is the preferred method for such checks.

feature_engine/timeseries/forecasting/window_features.py
KishManani: I think we need to consider what kind of transformations we want to accept. `agg` is orders of magnitude faster for  operations like sum and mean relative to `apply` (like ms for `agg` vs seconds for `apply`). Also the `agg` API is a bit more flexible in allowing you to specify multiple different transformations for different variables using a dictionary (see second highest comment (here)[https://stackoverflow.com/questions/21828398/what-is-the-difference-between-pandas-agg-and-apply-function]).

feature_engine/timeseries/forecasting/window_features.py
solegalli: This is what I've done for this parameter:

- removed all init checks. We leave that to pandas. Like this, this param can take whatever pandas accepts. 
- in addition, I also allow passing a a list with window values, so that several window features can be calculated with one transformer.
- in theory, the user could pass a list of callables, but I have not tested this process. @KishManani do you have an example of a useful feature that could be created with a bespoke function? we can use this in a test and more importantly in the documentation.

feature_engine/timeseries/forecasting/window_features.py
solegalli: I modified this method in https://github.com/Morgan-Sell/feature_engine/pull/9

feature_engine/timeseries/forecasting/window_features.py
solegalli: In https://github.com/Morgan-Sell/feature_engine/pull/9 I modified the functionality of the class so that it works with `agg` instead of `apply`. 

Agg can take strings representing each one of the rolling() allowed aggregation functions. That is convenient.

What the current functionality is not allowing, is for the user to pass a bespoke function. Which might be upsetting.

We have one of 2 options:

1. release like this and wait for user feedback
2. try to think how we can allow the use of bespoke functions

For 2, it would be helpful if @KishManani has an example of what this could be?

feature_engine/timeseries/forecasting/window_features.py
solegalli: i noticed that variables should also be part of the common tests. Added that in https://github.com/Morgan-Sell/feature_engine/pull/9

feature_engine/timeseries/forecasting/window_features.py
solegalli: fixed.

solegalli:
> I would also include an option to allow expanding windows: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.expanding.html

Regarding expanding windows, I tried to make this optional, but pandas.expanding windows takes an argument less than rolling windows, so it gets messy with the docs.

Also, it gets even more messy now that the parameter that is missing in pandas expanding is the one that allows the user to create multiple features for different window sizes.

So I think, the best is, to have expanding windows in a different class.

On that note, I am unsure as to how useful expanding windows as implemented in pandas, are to create features for forecasting. In my mind, I would like expanding windows with limits, and not extending all the way from the first observations. For example, I would like the ast 3,4, 5, and 6 months of data, but most likely not all data from the beginning. And this is not covered in pandas expanding which is just brute force from the first observations.

With WindowFeatures, if we pass the window values in a list, we can, in essence, create expanding windows as the ones I describe.

So question for @KishManani is, is pandas.expanding really useful in forecasting? If yes, we should create a separate issue for a separate class.

And the same for the other pandas functions like weighted windows since we are here :p

feature_engine/timeseries/forecasting/window_features.py
solegalli: > few questions regarding the co

feature_engine checks are not failing for me in this pr https://github.com/Morgan-Sell/feature_engine/pull/9

sklearn check_estimators tests are failing. And not sure why yet.

solegalli:
Hi @Morgan-Sell @KishManani 

Thank you so much for your valuable contributions to this class.

I made a PR to Morgan's repo here: https://github.com/Morgan-Sell/feature_engine/pull/9 where I address most of the questions and issues that came up in this PR thread.

The main things outstanding are:

1. expand the tests for when `window` is a list, flagged with TODO (@Morgan-Sell )
2. fix test for sort_index, flagged with TODO (@Morgan-Sell )
3. fix sklearn check_estimator (@solegalli )
4. expand the docs user guide, flagged with TODO (@solegalli )
5. create a new issue for expanding windows (@KishManani )
6. test passing a base indexer to window (if @KishManani has an example, we can take it from there)
7. see if we can allow bespoke functions instead of just `add` with strings (team thinking)

I put some tentative names next to each issue. Would this be alright for all of you? 

@Morgan-Sell if you go ahead with merging and the first changes, I 'll pick up afterwards with the sklearn tests.

Thank you all!

solegalli:
@KishManani 

When recreating the notebooks for section 3, I realise that it would be convenient to have the features created for timestamps not present in the train data.

For example, if last timestamp of train data is 2020-01-01, and say we have daily data. If we create lag 1 day features, we would have the value for timestamp 2020-01-02, however, the current implementation of lag features (and also window features) don't add these values, because they would add values up to the timestamp in the train data.

Is this desired? or would we want the future value as well?

I am thinking for multistep forecasting, I want to pass the value of the feature at time t, and want the transformer to return t+1, this would not work if the dataframe contains 1 row only at the moment.

thoughts?

Morgan-Sell:
Hi @solegalli and @KishManani,

I'm back! I was visiting family on the east coast. Thanks for all the great work while I was gone!

@solegalli,

1. I'm game to run point on:
```
- expand the tests for when window is a list, flagged with TODO (@Morgan-Sell )--
- fix test for sort_index, flagged with TODO (@Morgan-Sell )
```

2. Your abovementioned idea of having a window function for missing data is interesting! In energy, it is not uncommon for a SCADA to miss recording time sequences. Consequently, there are missing values for energy generation, irradiation, etc. I see the window function that you are proposing to serve as a `fillna` in such scenarios

feature_engine/timeseries/forecasting/window_features.py
KishManani: I don't have a good example to hand. My recommendation is to implement it only if an issue is raised with a good use case.

feature_engine/timeseries/forecasting/window_features.py
KishManani: Bespoke functions to pass to `agg` is anything that aggregates the time series into a scalar value (e.g., a quantile, the kurtosis, the entropy).

KishManani:
https://github.com/feature-engine/feature_engine/pull/380#issuecomment-1062057682

> So question for @KishManani is, is pandas.expanding really useful in forecasting? If yes, we should create a separate issue for a separate class.

> And the same for the other pandas functions like weighted windows since we are here :p

Indeed, creating windows of various sizes is different from expanding windows. Expanding windows is the expansion of a window from the start and contains all data points. The main feature I've seen created using an expanding window is taking the mean of the entire history. In my experience these features tend not to be as important as looking at averages over recent windows. Worth adding a separate class then. 

Weighting can be helpful. For, example in giving more weight to recent time periods compared to ones in the past. I've normally achieved this by passing a custom function into `agg` rather than specifying `win_type`.

feature_engine/timeseries/forecasting/window_features.py
KishManani: Also, allowing a custom function allows you to do weighted averages too:

```
weights = np.array([0.1, 0.2, 0.3, 0.4])
df.rolling(window=4).agg(lambda x: np.sum(x * weights))
```

KishManani:
https://github.com/feature-engine/feature_engine/pull/380#issuecomment-1063116124

We will need to project features into the future at predict time for forecasting. However, I don't think that logic should belong to these transformers. I think it is simpler and more intuitive that the transformers should return the values of the lag and window features for the time periods provided in the inputted dataframe in the `.fit_transform()` method.

tests/test_time_series/test_forecasting/test_window_features.py
```diff
@@ -0,0 +1,586 @@
...
+                    "module_temp_window_3_sum",
+                    "module_temp_window_3_mean",
+                ],
+                axis=1
+            )
+        )
+    )
+
+
+def test_sort_index(df_time):
```
KishManani, line 572: This test could be simplified to the following:

```
from pandas.testing import assert_frame_equal
def test_sort_index(df_time):
    # Shuffle dataframe
    Xs = df_time.sample(frac=1)

    transformer = WindowFeatures(sort_index=False)
    df_tr = transformer.fit_transform(Xs)
    assert_frame_equal(df_tr[transformer.variables_], Xs[transformer.variables_])

    transformer = WindowFeatures(sort_index=True)
    df_tr = transformer.fit_transform(Xs)
    assert_frame_equal(
        df_tr[transformer.variables_], Xs[transformer.variables_].sort_index()
    )
```

feature_engine/timeseries/forecasting/window_features.py
```diff
@@ -0,0 +1,270 @@
...
+    periods: int, list of ints, default=1
+        Number of periods to shift. Can be a positive integer. See param `periods` in
+        pandas `shift`.
+
+    freq: str, list of str, default=None
+        Offset to use from the tseries module or time rule. See parameter `freq` in
+        pandas `shift()`.
+
+    sort_index: bool, default=True
+        Whether to order the index of the dataframe before creating the lag features.
```
KishManani, line 90: Typo: I think it should say `... window features.` rather than `... lag features.`

feature_engine/timeseries/forecasting/window_features.py
```diff
@@ -0,0 +1,270 @@
...
+        X = super().transform(X)
+
+        if isinstance(self.window, list):
+            df_ls = []
+            for win in self.window:
+                tmp = (
+                    X[self.variables_]
+                    .rolling(window=win)
+                    .agg(self.functions)
+                    .shift(periods=self.periods, freq=self.freq)
```
KishManani, line 187: What is the use case to shift the rolling window feature by greater than one period? If there isn't one, then removing it could help simplify this class.

feature_engine/timeseries/forecasting/window_features.py
```diff
@@ -0,0 +1,270 @@
...
+                    "Some features in input_features were not used as input for the "
+                    "window features. You can only get the names of the window features "
+                    "with this function."
+                )
+            # create just indicated window features
+            input_features_ = input_features
+
+        if isinstance(self.window, list):
+            feature_names = [
+                str(feature) + f"_window_{win}" + f"_{agg}"
```
KishManani, line 247: This could be simplified to: 

```
f"{feature}_window_{win}_{agg}"
```

feature_engine/timeseries/forecasting/base_forecast.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+@Substitution(
+    missing_values=_missing_values_docstring,
+    drop_original=_drop_original_docstring,
+    feature_names_in_=_feature_names_in_docstring,
+    fit=_fit_not_learn_docstring,
+    n_features_in_=_n_features_in_docstring,
+)
+class BaseForecast(BaseEstimator, TransformerMixin):
```
KishManani, line 38: Minor point: I think naming this class BaseForecast could be misleading. It's not a base forecast and doesn't produce forecasts. The docstring is clear though. A more explicit but verbose name could be: `BaseForecastTransformer`, `BaseTimeseriesTransformer`.

feature_engine/timeseries/forecasting/window_features.py
solegalli: I think it is a fair point.

I just worry that some users will be using pandas `periods` mostly for shift, and then obliging them to use `freq` instead must be disorienting/frustrating.

Also, supposedly, pandas also shifts and rolls even if we do not have a datetime in the index. So in those cases, freq won't be useful.

PR #384, opened 2022-03-11
Fixes CV split bug in SelectByShuffling 

### LINKED ISSUE #383: Wrong reference for SelectByShuffling
If I am right, the calculation of the "scorer" between the unshuffled and the shuffled data set is not fair. 

The initial model performance is computed on the test set only:

```
# store initial model performance
self.initial_model_performance_ = model["test_score"].mean()
```

while the shuffled performance is computed on the full (shuffled) data set

```
# determine the performance with the shuffled feature
performance = np.mean(
            [scorer(m, X_shuffled, y) for m in model["estimator"]]
)
```

This leads to negative values for the performance drift 
```
performance_drift = self.initial_model_performance_ - performance
```

A mapping of CV model to the fold used for test could be a solution.

gverbock:
closes #383 
@solegalli I started working on it.


solegalli:
I made a small change to the class you are working on, and also fixed some of the tests that are failing. 

You may want to fetch and rebase the changes :)

feature_engine/selection/shuffle_features.py
```diff
@@ -169,6 +170,8 @@ def fit(self, X: pd.DataFrame, y: pd.Series):
 
         if isinstance(y, pd.Series):
             y = y.reset_index(drop=True)
+        else:
```
solegalli, line 173: if y is a pd.Series already, why do we need to re-assign?

solegalli (commented):
Nice one with the `check_cv`. I wasn't aware of this function.

Let me know when the tests are finished.

I am aware that the tests failing in the `get_feature_names` branch is a road blocker. So I am working on that as we speak. And hope to have it ready by Wednesday.

Thank you for the contribution!

feature_engine/selection/shuffle_features.py
gverbock: I must say that I am a bit puzzled by this one. My reasoning was: if there is a condition on y being and pd.Series then y could in some cases also not be a Series. Because I make a selection using `.iloc` I though of forcing y to be a pd.Series.
However when I do that it the unit tests fails and throw an error I cannot reproduce on my development notebook. I now removed it but I must say I am not 100% confident this is the right choice.

feature_engine/selection/shuffle_features.py
```diff
@@ -211,7 +217,10 @@ def fit(self, X: pd.DataFrame, y: pd.Series):
 
             # determine the performance with the shuffled feature
             performance = np.mean(
-                [scorer(m, X_shuffled, y) for m in model["estimator"]]
+                [
+                    scorer(m, X_shuffled.iloc[idx], y.iloc[idx])
+                    for m, idx in zip(model["estimator"], validation_indices)
```
solegalli, line 222: nice one with zip. This is a method that I should use more :)

tests/test_selection/test_shuffle_features.py
```diff
@@ -41,11 +41,12 @@ def test_sel_with_default_parameters(df_test):
     pd.testing.assert_frame_equal(sel.transform(X), Xtransformed)
 
 
-def test_regression_cv_3_and_r2(load_diabetes_dataset):
+def test_regression_cv_3_and_r2(df_test, load_diabetes_dataset):
```
solegalli, line 44: I don't think we are using df_test here. Are you still working on the tests and I got ahead of my game? If so, apologies.

solegalli (commented):
Hi @gverbock 

thank you for the additional changes.

I remember that we had to reset the index of X to avoid problems when doing the cv. And because we reset the index of X, we then also had to reset y if it was a pd.Series.

If you think we can avoid resetting both indexes, please do improve the code.

I have now updated the `get_feature_names` branch enough for all tests to pass. So if you could rebase once last time, then, we should be able to see if the tests pass here as well and merge :)

Let me know when you are done with the changes, so i don't add annoying comments or questions if you are half way :)

Thank you!

tests/test_selection/test_shuffle_features.py
gverbock: That's true. This is a left over from a check. I'll remove it.

solegalli:
Thank you @gverbock !!

PR #390, opened 2022-03-16
improves select by target mean functionality

### LINKED ISSUE #227: expand tests selection by target mean performance
the tests available right now are a bit high level. It does not test internal functionality, for example, the calculation of the performances. It does not test threshold defaulting to None. And also, it does not test when we pass a subset of categories to the selector.

### LINKED ISSUE #238: Very URGENT: SelectByTargetMeanPerformance()  Value Error
This function throws the following

 **Value Error: Input contains NaN, infinity or a value too large ford type('float64').** None of these seem to be the case.

selector = SelectByTargetMeanPerformance(
    variables=categorical, # the variables to examine
    scoring='r2_score', # the metric to use for the performance evaluation
    threshold=0.1, # the minimum performance threshold for a variable to be selected
    cv=2, # the cross-validation fold
    random_state=0)

selector.fit(X_train, y_train)

Data:

[X_train.xlsx](https://github.com/solegalli/feature_engine/files/6142687/X_train.xlsx)
[y_train.xlsx](https://github.com/solegalli/feature_engine/files/6142688/y_train.xlsx)

Error Snapshot:

[error.pdf](https://github.com/solegalli/feature_engine/files/6142624/error.pdf)

### LINKED ISSUE #296: Improve SelectByTargetMeanPerformance functionality

solegalli:
closes #227 
closes #238 
closes #296 
closes #334 
closes #358 

FYI @Morgan-Sell 

I am bringing the changes in #358 to this new PR, made to the branch `get_feature_names` which has a lot of code improvements.

- [x] one last review to the code
- [x] update user guide for target mean selection 
- [x] build and revise documentation
- [ ] add user guide for target mean classifier and regressor?


solegalli:
@Morgan-Sell 

I think we managed to make all changes required to have these classes up and running. Next step would be to incorporate them into the feature selection class, which was the whole point of the exercise, lol.

Would you mind having a look at these 11 files and letting me know if I missed anything?

Morgan-Sell:
Hola @solegalli,

You've been quite busy! I'll review the code.

How do I see which branch this is? I think I may have deleted it from my local CPU. I thought it was `mean_target_prediction`. I see that you created a new branch called `sel_t_mean`.

I've lost track of the workflow ;)

PR #405, opened 2022-03-31
extends functionality of DatetimeFeatures to extract features from the dataframe index

dodoarg:
Here's a draft to implement the functionality. I might have forgotten about some test cases and there might be better ways to refactor the logic overall

I tried to keep the logic as consistent as possible with the rest of the cases (e.g. find_or_check_dt_variables dumps "index" into ["index"] as it would do with any other string etc) although the fact that we need to access .index rather than a (set of) column(s) introduces some "boilerplate" code

I did think of having _check_contains_na check for nans in the index but I found out some check estimator tests would fail then, so I just added that logic under an if statement in the datetime transformer (more boilerplate, ugh)

Also, this is probably a trivial note, but as things stand there's no coverage for the case where an actual variable (not the index) is named "index". I suppose we might just leave that as it is as it sounds like a very unrealistic name for a column anyway

Let me know how we can improve it

### DIFF
--- a/feature_engine/datetime/datetime.py
+++ b/feature_engine/datetime/datetime.py
@@ -220,12 +220,7 @@ class DatetimeFeatures(BaseEstimator, TransformerMixin):
                 raise TypeError("The dataframe index is not datetime.")
 
             if self.missing_values == "raise":
-                if X.index.isnull().any():
-                    raise ValueError(
-                        "The dataframe index contains missing data. "
-                        "Check and remove those before using this transformer "
-                        "or set missing_values to False."
-                    )
+                self._check_index_contains_na(X.index)
 
             self.variables_ = None
 
@@ -279,12 +274,7 @@ class DatetimeFeatures(BaseEstimator, TransformerMixin):
         # special case index
         if self.variables_ is None:
             if self.missing_values == "raise":
-                if X.index.isnull().any():
-                    raise ValueError(
-                        "The dataframe index contains missing data. "
-                        "Check and remove those before using this transformer "
-                        "or set missing_values to False."
-                    )
+                self._check_index_contains_na(X.index)
 
             for feat in self.features_to_extract_:
                 X[FEATURES_SUFFIXES[feat][1:]] = FEATURES_FUNCTIONS[feat](
@@ -417,6 +407,14 @@ class DatetimeFeatures(BaseEstimator, TransformerMixin):
                 feature_names = self.feature_names_in_ + feature_names
 
         return feature_names
+
+    def _check_index_contains_na(self, index: pd.Index):
+        if index.isnull().any():
+            raise ValueError(
+                "The dataframe index contains missing data. "
+                "Check and remove those before using this transformer "
+                "or set missing_values to False."
+            )
 
     def _more_tags(self):
         tags_dict = {"variables": "datetime"}
--- a/feature_engine/datetime/datetime.py
+++ b/feature_engine/datetime/datetime.py
@@ -42,10 +42,10 @@ from feature_engine.variable_manipulation import (
 class DatetimeFeatures(BaseEstimator, TransformerMixin):
     """
     DatetimeFeatures extracts date and time features from datetime variables, adding
-    new columns to the dataset. DatetimeFeatures is able to extract datetime
-    information from existing datetime or object-like variables.
+    new columns to the dataset. DatetimeFeatures can extract datetime information from
+    existing datetime or object-like variables or from the dataframe index.
 
-    DatetimeFeatures uses pandas.to_datetime to convert object variables to datetime
+    DatetimeFeatures uses `pandas.to_datetime` to convert object variables to datetime
     and pandas.dt to extract the features from datetime.
 
     The transformer supports the extraction of the following features:
@@ -75,8 +75,8 @@ class DatetimeFeatures(BaseEstimator, TransformerMixin):
 
     Parameters
     ----------
-    variables: list, default=None
-        The list of variables to extract date and time features from.
+    variables: str, list, default=None
+        List wit the variables from which date and time information will be extracted.
         If None, the transformer will find and select all datetime variables,
         including variables of type object that can be converted to datetime.
         If "index", the transformer will extract datetime features from the
@@ -86,8 +86,8 @@ class DatetimeFeatures(BaseEstimator, TransformerMixin):
         The list of date features to extract. If None, the following features will be
         extracted: "month", "year", "day_of_week", "day_of_month", "hour",
         "minute" and "second". If "all", all supported features will be extracted.
-        Alternatively, you can pass a list with the names of the supported features
-        you want to extract.
+        Alternatively, you can pass a list with the names of the features you want to
+        extract.
 
     drop_original: bool, default="True"
         If True, the original datetime variables will be dropped from the dataframe.
@@ -96,29 +96,37 @@ class DatetimeFeatures(BaseEstimator, TransformerMixin):
         Indicates if missing values should be ignored or raised. If 'raise' the
         transformer will return an error if the the datasets to `fit` or `transform`
         contain missing values. If 'ignore', missing data will be ignored when
-        performing the feature extraction.
+        performing the feature extraction. Missing data is only evaluated in the
+        variables that will be used to derive the date and time features. If features
+        are derived from the dataframe index, missing data will be checked in the
+        index.
 
     dayfirst: bool, default="False"
-        Specify a date parse order if arg is str or its list-likes. If True, parses
-        dates with the day first, eg 10/11/12 is parsed as 2012-11-10.
+        Specify a date parse order if arg is str or is list-like. If True, parses
+        dates with the day first, e.g. 10/11/12 is parsed as 2012-11-10. Same as in
+        `pandas.to_datetime`.
 
     yearfirst: bool, default="False"
-        Specify a date parse order if arg is str or its list-likes.
+        Specify a date parse order if arg is str or is list-like.
+        Same as in `pandas.to_datetime`.
 
-        - If True parses dates with the year first, eg 10/11/12 is parsed as 2010-11-12.
+        - If True parses dates with the year first, e.g. 10/11/12 is parsed as
+          2010-11-12.
         - If both dayfirst and yearfirst are True, yearfirst is preceded.
 
     utc: bool, default=None
         Return UTC DatetimeIndex if True (converting any tz-aware datetime.datetime
-        objects as well).
+        objects as well). Same as in `pandas.to_datetime`.
 
     Attributes
     ----------
     variables_:
-        List of variables from which date and time features will be extracted.
+        List of variables from which date and time features will be extracted. If None,
+        features will be extracted from the dataframe index.
 
     features_to_extract_:
-        The date and time features that will be extracted from each variable.
+        The date and time features that will be extracted from each variable or the
+        index.
 
     {feature_names_in_}
 
@@ -342,13 +350,12 @@ class DatetimeFeatures(BaseEstimator, TransformerMixin):
 
         Parameters
         ----------
-        input_features: list, default=None
-            Input features. If `input_features` is `None`, then the names of all the
-            variables in the transformed dataset (original + new variables) is returned.
-            Alternatively, only the names for the datetime features derived from
-            input_features will be returned.
-            If the features were derived from the dataframe index, pass 'index'
-            to obtain the datetime features or None for all output features."
+        input_features: str, list, default=None
+            Input features. If `None`, then the names of all the variables in the
+            transformed dataset (original + new variables) is returned. If 'index',
+            then the name of the features created from the index is returned. If list
+            with feature names, the names for the datetime features derived from
+            the variables in the list will be returned."
 
         Returns
         -------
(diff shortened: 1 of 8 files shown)


feature_engine/datetime/datetime.py
```diff
@@ -209,7 +211,14 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
 
         # check if datetime variables contains na
         if self.missing_values == "raise":
-            _check_contains_na(X, self.variables_)
+            if self.variables_ == ["index"]:
```
solegalli, line 214: my idea was that the argument `variables` in the init, can take 3 values:

- None: defaults to finding all dt variables (already implemented)
- Variable list: checks if variables entered by user are or can be datetime (already implemented)
- "index", taken as a string, which will extract datetime features from the index (new in this PR)

Then, here I would tailor the message to say, there were nan in the index, for example.

feature_engine/datetime/datetime.py
```diff
@@ -252,24 +261,44 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
 
-        # reorder variables to match train set
-        X = X[self.feature_names_in_]
+            else:
+                _check_contains_na(X, self.variables_)
 
-        # convert datetime variables
-        datetime_df = pd.concat(
-            [
+        if self.variables_ == ["index"]:
```
solegalli, line 274: to extract features from the index, do we absolutely need to make it a column?

 Can we not do df.index.month? or something like this? I saw it the other day on a stackoverflow thread.

But maybe it does not fit with how we structured the class?

feature_engine/datetime/datetime.py
```diff
@@ -280,14 +309,21 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
             )
 
         # create new features
-        for var in self.variables_:
+        if self.variables_ == ["index"]:
```
solegalli, line 312: If we extract features from the index, I am wondering whether we should just name the features as the datetime they represent? we are not creating features from multiple variables, so in principle, there would be no need to add the variable name at the front to distinguish them from anything else?

feature_engine/variable_manipulation.py
```diff
@@ -221,14 +221,23 @@ def _find_or_check_datetime_variables(
             raise ValueError("No datetime variables found in this dataframe.")
 
     elif isinstance(variables, (str, int)):
+        if variables == "index":
```
solegalli, line 224: I would not modify this function. I would handle the logic from the datetime class, because this is probably one of the only classes that will work with the index in this way.

tests/test_datetime/test_datetime_features.py
```diff
@@ -148,7 +159,11 @@ def test_raises_error_when_nan_in_transform_df(df_datetime):
     transformer.fit(df_datetime)
     # dataset containing nans
     with pytest.raises(ValueError):
-        DatetimeFeatures().transform(dates_nan)
+        transformer.transform(dates_nan)
+    transformer = DatetimeFeatures(variables="index")
```
solegalli, line 163: I see, here is when the problem you raised would come up. If the user has a variable called "index", but we are now using this string to indicate the df index, then they would not be able to extract features from this variable.

I think this is very unlikely, but the user still could pass ["index"] and then, they could extract features from the column called index. And if they pass "index" as a string, then, the class would use the index of the df. And we could make this clear in the docstrings. 

I think, as a first release, this could be enough, and then we hear user feedback. Unless you think otherwise?

solegalli (commented):
Hi @dodoarg 

Thank you so much for the quick turnaround! Very helpful and very kind of you :)

I was thinking in making fewer changes for this addition. I would only modify the transformer, and not the variables manipulation function. I was wondering if that would be possible?

Would you mind having a look at my comments and see if they make sense?

Thank you!

dodoarg:
I think I've tackled pretty much all issues you raised @solegalli ? 

- I removed the "index_" prefix for feature columns extracted from the index

- The part where the DatetimeIndex is dumped into a pd.Series was necessary because the way you access to datetime functions on an index object looks like `df.index.year` i.e. it "bypasses" the call to the dt module. Since our logic calls a set of lambda functions (FEATURES_FUNCTIONS) that all look like x.dt.year I thought it would add less boilerplate code to put the index data into a Series so as to access those functions the same way rather than rewrite all them without the dt part just for this specific case

feature_engine/datetime/datetime.py
```diff
@@ -204,12 +208,33 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
...
-        # check if datetime variables contains na
-        if self.missing_values == "raise":
-            _check_contains_na(X, self.variables_)
+            if not (
+                is_datetime(X.index)
+                or (
+                    not is_numeric(X.index) and _is_categorical_and_is_datetime(X.index)
+                )
+            ):
+                raise TypeError("Index is not datetime.")
```
solegalli, line 220: Could we re-phrase the error as

"The dataframe index is not datetime."

feature_engine/datetime/datetime.py
```diff
@@ -204,12 +208,33 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
...
+
+            if self.missing_values == "raise":
+                if X.index.isnull().any():
+                    raise ValueError(
+                        "Index contain NaN. Check and "
+                        "remove those before using this transformer."
```
solegalli, line 226: Could we re-phrase this error message?

"The dataframe index contains missing data. Check and remove those before using this transformer or set `missing_values` to False."

feature_engine/datetime/datetime.py
```diff
@@ -204,12 +208,33 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
...
+                    )
+
+            self.variables_: List[Union[str, int]] = ["index"]
```
solegalli, line 229: If we are using the index to extract features, I would leave `variables_` to None. 

I think Feature-engine users would expect to find the variables that are considered for the transformation in this attribute. And in this case, no variable is. Because we are using the index. Also, if we set this to ["index"], it would clash with the unlikely situation in which the user has a variable called "index".

feature_engine/datetime/datetime.py
```diff
@@ -250,13 +275,36 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
         # Check if input data contains same number of columns as dataframe used to fit.
         _check_input_matches_training_df(X, self.n_features_in_)
 
+        # special case index
+        if self.variables_ == ["index"]:
```
solegalli, line 279: We could use either self.variables_ is None or self.variables =="index" without the underscore.

feature_engine/datetime/datetime.py
```diff
@@ -250,13 +275,36 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+            if self.missing_values == "raise":
+                if X.index.isnull().any():
+                    raise ValueError(
+                        "Index contain NaN. Check and "
+                        "remove those before using this transformer."
```
solegalli, line 284: could we rephrase as in fit()? 

Also wondering, we are repeating code in fit and transform, should we capture it in a private method in this class? 

```
       if self.missing_values == "raise":
            self._check_contains_na()
```

feature_engine/datetime/datetime.py
```diff
@@ -250,13 +275,36 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+                    )
+
+            for feat in self.features_to_extract_:
+                X[FEATURES_SUFFIXES[feat][1:]] = FEATURES_FUNCTIONS[feat](
+                    pd.Series(
+                        pd.to_datetime(
```
solegalli, line 290: converting something to datetime can take a lot of time if the dataset is big. Could we convert only if it is necessary instead of always?

solegalli: I notice that we do the same when the user passes variables. We convert to datetime all the time. Does pandas check that it is not datetime before doing the conversion? or converts everytime one says so, even though it could already be datetime?

Because in this case, it might be worth splitting the functionality. If not here, in a separate PR.

solegalli (commented):
Hi @dodoarg 

Thank you so much for the quick fixes on the code. This is really helpful.

I've got a couple of minor cosmetic requests, and then a question about pd.to_datetime() that might be worth exploring.

Would you be able to have a look?

Thank you!

feature_engine/datetime/datetime.py
solegalli: Answering to my own question, to_datetime() does check if it is datetime already.

https://github.com/pandas-dev/pandas/blob/v1.4.1/pandas/core/tools/datetimes.py#L1027
https://github.com/pandas-dev/pandas/blob/06d230151e6f18fdb8139d09abf539867a8cd481/pandas/core/tools/datetimes.py#L289

So I don't think we need to change this. Apologies.

I also made some manual checks :p 

```
start = time.time()
data['Date_Time'] = pd.to_datetime(data['Date_Time'])
end = time.time()
print("The time of execution of above program is :", end-start)
The time of execution of above program is : 0.9556782245635986
```

And if i repeat that, after it was already converted to datetime, it takes a fraction:
`The time of execution of above program is : 0.013006925582885742`

dodoarg:
Should be better now. Notice that when setting self.variables_ as None in the index case it occurred to me that I had to adapt get_feature_names_out to the index case

solegalli:
Thank you so much @dodoarg !!!! This will do!

Merging on Monday.

Have a nice weekend.

feature_engine/datetime/datetime.py
```diff
@@ -250,13 +271,32 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
         _check_input_matches_training_df(X, self.n_features_in_)
 
+        # special case index
+        if self.variables_ is None:
+            if self.missing_values == "raise":
+                self._check_index_contains_na(X.index)
+
+            for feat in self.features_to_extract_:
+                X[FEATURES_SUFFIXES[feat][1:]] = FEATURES_FUNCTIONS[feat](
+                    pd.Series(
```
solegalli, line 281: I just noticed that with this logic, we would be converting the index to datetime for every feature. 

Could we not convert it one, capture the series outside the loop, and then enter the loop using the already prepared series?

feature_engine/datetime/datetime.py
```diff
@@ -250,13 +271,32 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+                            X.index,
+                            dayfirst=self.dayfirst,
+                            yearfirst=self.yearfirst,
+                            utc=self.utc,
+                        ),
+                        index=X.index,
+                    )
+                )
+
+            return X
```
solegalli, line 292: could we have just 1 return at the end of the method and not break the logic in 2?

feature_engine/datetime/datetime.py
```diff
@@ -250,13 +271,32 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+
         # check if dataset contains na
         if self.missing_values == "raise":
             _check_contains_na(X, self.variables_)
 
         # reorder variables to match train set
         X = X[self.feature_names_in_]
```
solegalli, line 299: This line should also apply for the index features. Is the final of the transform checks, before starting any transform logic. It ensures the df passed to transform has the columns in the same order than the one passed in fit.

feature_engine/datetime/datetime.py
```diff
@@ -301,6 +342,8 @@ def get_feature_names_out(self, input_features: Optional[List] = None) -> List:
             variables in the transformed dataset (original + new variables) is returned.
             Alternatively, only the names for the datetime features derived from
             input_features will be returned.
+            If the transformer was fitted on the dataframe index, you may only pass
+            `index` as input_features, or leave it as `None`.
```
solegalli, line 346: Could we reprhase these 2 sentences as "If the features were derived from the dataframe index, pass 'index' to obtain the datetime features or None for all output features."

solegalli (commented):
Hi @dodoarg 

I took a second look at the PR just now, and I am wondering if we could make the loop more efficient by taking the pandas series out of the loop?

What do you think?

feature_engine/datetime/datetime.py
dodoarg: Sure. My bad.

feature_engine/datetime/datetime.py
dodoarg: If you'd rather have what follows under an else statement yea we can do that

feature_engine/datetime/datetime.py
dodoarg: Right. It probably didn't matter as far as the test go since we're not ever changing the existing columns but the index only when variables="index"

tests/test_datetime/test_datetime_features.py
dodoarg: Yea no I agree. If anything if someone comes up with a dataset where a column not only is named "index" but contains datetime info (??) it's probably their fault in terms of bad naming practices and even lack of common sense haha

dodoarg:
Should have made quick work of that in my break

dodoarg:
is this ready to merge @solegalli ?

Issue #406, opened 2022-04-04
DropFeatures fails when features_to_drop is a string 


noahjgreen295:
**Describe the bug**
`DropFeatures` [type hint and documentation says](https://feature-engine.readthedocs.io/en/1.2.x/api_doc/selection/DropFeatures.html#:~:text=Parameters,from%20the%20dataframe) `features_to_drop` can take `str or list` as parameter, but instantiation fails when `features_to_drop` is a string.

**To Reproduce**
`transformer=DropFeatures(features_to_drop="some_string")`

**Expected behavior**
`DropFeatures` should instantiate successfully and drop the single feature specified by the string name when `transform()` or `fit_transform()` is called.

**Screenshots**
Error message:
`ValueError: features_to_drop should be a list with the name of the variablesyou wish to drop from the dataframe.`

**Additional context**
This is a simple fix, I am happy to take care of it. For convenience I would also convert the string to a single-element list when storing the attribute, for consistency.

solegalli:
Thank you!

noahjgreen295:
no prob, on it now!

noahjgreen295:
OK, PR #407 here!

PR #407, opened 2022-04-04
Fix DropFeatures fail when features_to_drop is string, fixes issue 406

noahjgreen295:
Fix to issue #406 , where `DropFeatures` raises exception when instantiating with `features_to_drop` as string, even though docs and type hinting says string is supported.

### DIFF
--- a/feature_engine/selection/drop_features.py
+++ b/feature_engine/selection/drop_features.py
@@ -58,9 +58,6 @@ class DropFeatures(BaseSelector):
                 "you wish to drop from the dataframe."
             )
 
-        if isinstance(features_to_drop, str):
-            features_to_drop = [features_to_drop]
-
         self.features_to_drop = features_to_drop
 
     def fit(self, X: pd.DataFrame, y: pd.Series = None):
--- a/feature_engine/selection/drop_features.py
+++ b/feature_engine/selection/drop_features.py
@@ -52,9 +52,9 @@ class DropFeatures(BaseSelector):
 
     def __init__(self, features_to_drop: List[Union[str, int]]):
 
-        if not isinstance(features_to_drop, list) or len(features_to_drop) == 0:
+        if not isinstance(features_to_drop, (str, list)) or len(features_to_drop) == 0:
             raise ValueError(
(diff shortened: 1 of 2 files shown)


feature_engine/selection/drop_features.py
```diff
@@ -52,12 +52,15 @@ class DropFeatures(BaseSelector):
...
 
-        if not isinstance(features_to_drop, list) or len(features_to_drop) == 0:
+        if not isinstance(features_to_drop, (str, list)) or len(features_to_drop) == 0:
             raise ValueError(
-                "features_to_drop should be a list with the name of the variables"
+                "features_to_drop should be a list with the name of the variables "
                 "you wish to drop from the dataframe."
             )
 
+        if isinstance(features_to_drop, str):
```
solegalli, line 61: To follow sklearn conventions, we must not re-assign user' entered values in the init parameters. 

Here, we need to let feature_to_drop be a string, and make it a list when we create features_to_drop_ in the fit method.

I am surprised that the `check_estimator` from sklearn did not pick this up. I was expecting the test to fail :_(

feature_engine/selection/drop_features.py
```diff
@@ -52,12 +52,15 @@ class DropFeatures(BaseSelector):
 
     def __init__(self, features_to_drop: List[Union[str, int]]):
 
-        if not isinstance(features_to_drop, list) or len(features_to_drop) == 0:
+        if not isinstance(features_to_drop, (str, list)) or len(features_to_drop) == 0:
             raise ValueError(
-                "features_to_drop should be a list with the name of the variables"
+                "features_to_drop should be a list with the name of the variables "
                 "you wish to drop from the dataframe."
```
solegalli, line 58: since we are working on this class... can we add f"Got {features_to_drop} instead." to the error message?

Thank you!

solegalli (commented):
Thank you @noahjgreen295 !!

A tiny change requested, to follow conventions...

noahjgreen295:
OK made both changes!

- Undid reassignment of string to list
- Addition to exception message

solegalli:
Thank you!!!

PR #410, opened 2022-04-07
add check_y and check_x_y to check that indexes of X and y match when encoding

### LINKED ISSUE #365: transformers that take y should check X.shape[0] == y.shape[0].

### LINKED ISSUE #376: Encoders that are f(X, y) can produce nan results when y has non-standard index and X becomes an np.ndarray

solegalli:
closes #377 
closes #376 
closes #365 

@noahjgreen295 @bmreiniger

I am bringing the changes from #377 to this new PR at the back of former PR #408

Regarding the X and y checks, I am bringing in sklearn functionality to check y and then to check X and y (taken from their check_X_y function). I think that decreases the amount of new things we need to code.

I think I captured most of what we discussed in #377 , and I brought forward the tests and ideas from @noahjgreen295 as they were or slightly reformatted and also the suggestions from @bmreiniger.

Would you guys have a look and let me know what you think?

Let me know if you are OK to merge.

Thank you so much for raising and also for the support on this issue.

### DIFF
--- a/feature_engine/dataframe_checks.py
+++ b/feature_engine/dataframe_checks.py
@@ -89,7 +89,7 @@ def check_X(X: Union[np.generic, np.ndarray, pd.DataFrame]) -> pd.DataFrame:
 def check_y(
         y: Union[np.generic, np.ndarray, pd.Series, List],
         multi_output: bool = False,
-        y_numeric: bool = True,
+        y_numeric: bool = False,
 ) -> pd.Series:
     """
     Checks that y is a series, or alternatively, if it can be converted to a series.
@@ -117,7 +117,12 @@ def check_y(
         raise ValueError("y cannot be None.")
 
     elif isinstance(y, pd.Series):
-        _check_y(y, multi_output=multi_output, y_numeric=y_numeric)
+        if y.isnull().any():
+            raise ValueError("y contains NaN infinity values.")
+        if y.dtype != "O" and not np.isfinite(y).all():
+            raise ValueError("y contains infinity values.")
+        if y_numeric and y.dtype == "O":
+            y = y.astype("float")
         y = y.copy()
 
     else:
--- a/feature_engine/dataframe_checks.py
+++ b/feature_engine/dataframe_checks.py
@@ -7,6 +7,7 @@ from typing import List, Union
 import numpy as np
 import pandas as pd
 from scipy.sparse import issparse
+from sklearn.utils.validation import _check_y
 
 
 def check_X(X: Union[np.generic, np.ndarray, pd.DataFrame]) -> pd.DataFrame:
@@ -83,6 +84,80 @@ def check_X(X: Union[np.generic, np.ndarray, pd.DataFrame]) -> pd.DataFrame:
         )
 
     return X
+
+
+def check_y(
+    y: Union[np.generic, np.ndarray, pd.Series, List],
+    multi_output: bool = False,
+    y_numeric: bool = True,
+) -> pd.Series:
+    """
+    Checks that y is a series, or alternatively, if it can be converted to a series.
+
+    Parameters
+    ----------
+    y : pd.Series, np.array, list
+
+    multi_output : bool, default=False
+        Whether to allow 2D y (array). If false, y will be
+        validated as a vector. y cannot have np.nan or np.inf values if
+        multi_output=True.
+
+    y_numeric : bool, default=False
+        Whether to ensure that y has a numeric type. If dtype of y is object,
+        it is converted to float64. Should only be used for regression
+        algorithms.
+
+    Returns
+    -------
+    y: pd.Series
+    """
+
+    if y is None:
+        raise ValueError("y cannot be None.")
+
+    elif isinstance(y, pd.Series):
+        _check_y(y, multi_output=multi_output, y_numeric=y_numeric)
+        y = y.copy()
+
+    else:
+        y = _check_y(y, multi_output=multi_output, y_numeric=y_numeric)
+        y = pd.Series(y)
+
+    return y
+
+
+def _check_pd_X_y(
+    X: Union[np.generic, np.ndarray, pd.DataFrame],
+    y: Union[np.generic, np.ndarray, pd.Series, List],
+    multi_output: bool = False,
+    y_numeric: bool = True,
+) -> (pd.DataFrame, pd.Series):
+    """
+    Ensures X and y are compatible pandas DataFrame and Series. If both are pandas
+    objects, checks that their indexes match. If any is a numpy array, converts to
+    pandas object with compatible index.
+
+    This transformer ensures that we can concatenate X and y using `pandas.concat`,
+    functionality needed in the encoders.
+
+    Parameters
+    ----------
+    X: Pandas DataFrame or numpy ndarray
+    y: Pandas Series or numpy ndarray
+
+    Raises
+    ------
+    ValueError: if X and y are pandas objects with inconsistent indexes.
+    TypeError: if X is sparse matrix, empty dataframe or not a dataframe.
+    TypeError: if y can't be parsed as pandas Series.
+
+    Returns
+    -------
+    X: Pandas DataFrame
+    y: Pandas Series
+    """
+    pass
 
 
 def _check_X_matches_training_df(X: pd.DataFrame, reference: int) -> None:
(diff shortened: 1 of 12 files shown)


noahjgreen295:
Hi @solegalli - I'm sorry I'm just having git some trouble getting this branch into my cloned repo. I did the following:
`% git pull upstream checkky`
`% git switch checkky`
I'm pretty sure that what I have now locally is the correct branch (if not, let me know).
I am now trying to put a copy into my cloned repo, to push changes to there, but when I try `git push`, I get this error:
`remote: Permission to feature-engine/feature_engine.git denied to noahjgreen295.
fatal: unable to access 'https://github.com/feature-engine/feature_engine.git/': The requested URL returned error: 403`

Am I doing this the right way? Sorry for the basic git question - thanks!

feature_engine/dataframe_checks.py
```diff
@@ -85,6 +86,108 @@ def check_X(X: Union[np.generic, np.ndarray, pd.DataFrame]) -> pd.DataFrame:
...
+    objects, checks that their indexes match. If any is a numpy array, converts to
+    pandas object with compatible index.
+
+    This transformer ensures that we can concatenate X and y using `pandas.concat`,
+    functionality needed in the encoders.
+
+    Parameters
+    ----------
+    X: Pandas DataFrame or numpy ndarray
+    y: Pandas Series or numpy ndarray
```
bmreiniger, line 147: some parameters not documented

feature_engine/dataframe_checks.py
```diff
@@ -85,6 +86,108 @@ def check_X(X: Union[np.generic, np.ndarray, pd.DataFrame]) -> pd.DataFrame:
...
+    Parameters
+    ----------
+    y : pd.Series, np.array, list
+
+    multi_output : bool, default=False
+        Whether to allow 2D y (array). If false, y will be
+        validated as a vector. y cannot have np.nan or np.inf values if
+        multi_output=True.
+
+    y_numeric : bool, default=False
```
bmreiniger, line 106: default in docstring disagrees with code default

feature_engine/dataframe_checks.py
```diff
@@ -85,6 +86,108 @@ def check_X(X: Union[np.generic, np.ndarray, pd.DataFrame]) -> pd.DataFrame:
...
+    Returns
+    -------
+    y: pd.Series
+    """
+
+    if y is None:
+        raise ValueError("y cannot be None.")
+
+    elif isinstance(y, pd.Series):
+        _check_y(y, multi_output=multi_output, y_numeric=y_numeric)
```
bmreiniger, line 120: by not assigning, you get all the checks without the conversion, but e.g. the numeric conversion also doesn't happen...

bmreiniger (commented):
I was going to suggest using sklearn's `check_X_y` and/or `check_array` in a fashion similar to your usage of `_check_y` (could reduce some duplication in your `check_X`, and uses a public sklearn method instead of a private one), but there's a lot going on there for converting to array/sparse that would be a waste.  Hrm...

Otherwise, minor comments.

solegalli:
> Hi @solegalli - I'm sorry I'm just having git some trouble getting this branch into my cloned repo. I did the following: `% git pull upstream checkky` `% git switch checkky` I'm pretty sure that what I have now locally is the correct branch (if not, let me know). I am now trying to put a copy into my cloned repo, to push changes to there, but when I try `git push`, I get this error: `remote: Permission to feature-engine/feature_engine.git denied to noahjgreen295. fatal: unable to access 'https://github.com/feature-engine/feature_engine.git/': The requested URL returned error: 403`
> 
> Am I doing this the right way? Sorry for the basic git question - thanks!

You need to push the branch to your own repo, and make a PR from there to the main repo's checkky branch.

feature_engine/dataframe_checks.py
solegalli: Good catch, thank you! Refactoring.

feature_engine/dataframe_checks.py
```diff
@@ -85,6 +86,108 @@ def check_X(X: Union[np.generic, np.ndarray, pd.DataFrame]) -> pd.DataFrame:
...
+    -------
+    X: Pandas DataFrame
+    y: Pandas Series
+    """
+
+    def _check_X_y(X, y):
+        X = check_X(X)
+        y = check_y(y, multi_output=multi_output, y_numeric=y_numeric)
+        check_consistent_length(X, y)
+        return X, y
```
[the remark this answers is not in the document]
solegalli: Yes, makes sense. The thing is, if I ran this function first, then X and y would be a pandas df and series, so I can't really check for the different scenarios any more. That's why I took this convoluted approach instead. 

How would you approach it?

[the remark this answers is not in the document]
bmreiniger: Oh, of course. This is fine then. You could also maybe define "the" index in cases,  _then_ do the checks/conversions?

solegalli:
Thank you @bmreiniger @noahjgreen295 for raising and your support on this issue. Merging.

PR #412, opened 2022-04-11
Add "input_features" param to all get_feature_names_out() method [Scikit-learn compatibility]

### LINKED ISSUE #402: get_feature_names_out does not work with the pipeline
I think this is because the pipeline passes by default `input_features`, and not all transformers allow the parameter input_features at the moment. 

We need to fix this and add a couple of tests to ensure it works with the pipeline.

Hey @janrito @hectorpatino 

I would really appreciate if you could help me fix it :)

Morgan-Sell:
Closes #402 

Notes from #402:

**From 29-Mar-22:**

I think this is because the pipeline passes by default input_features, and not all transformers allow the parameter input_features at the moment.

We need to fix this and add a couple of tests to ensure it works with the pipeline.

**From 11-Apr-22:**

The idea is that now, all classes have the method get_feature_names_out() that return the name of the features in the transformed dataset. That is, the dataset that results of applying the transform() method.

A bunch of those methods in the classes take no input. That makes the method not compatible with the Scikit-learn pipeline.

So basically, we need to go, class per class, and for those where the get_feature_names_out takes no input, we need to modify it so that it can take the parameter input_features which should default to None. This parameter should have no function through (we need to make this clear in the docstrings), but we need it to default to None to make it compatible with the pipeline.

Silly I know.

One thing to notice, if the class itself does not have a get_feature_names_out, it is because it is inheriting the functionality from the parent class. So actually, it is not as difficult as it looks.


Morgan-Sell:
@solegalli,

Now that `input_features` is a parameter, should we add code to check the parameter's value (e.g., `None` returns all feature names) and confirm the provided string values exist?

feature_engine/base_transformers.py
```diff
@@ -154,9 +155,18 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
 
         return X
 
-    def get_feature_names_out(self) -> List:
+    @Substitution(
```
solegalli, line 158: This is genius!  I didn't know that I could also use this class to substitute in methods! Thank you!

feature_engine/datetime/datetime.py
```diff
@@ -343,19 +344,17 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
 
         Parameters
         ----------
-        input_features: str, list, default=None
-            Input features. If `None`, then the names of all the variables in the
-            transformed dataset (original + new variables) is returned. If 'index',
-            then the name of the features created from the index is returned. If list
-            with feature names, the names for the datetime features derived from
-            the variables in the list will be returned."
+        {input_features}
```
solegalli, line 357: I would not substitute this method for a common string, because datetime features has its own particular functionality. I would leave the original docstring.

feature_engine/docstrings.py
```diff
@@ -42,6 +42,14 @@ def __call__(self, obj):
         parameters or performing the transformation.
         """
 
+_input_features_docstring = """input_features: str, list, default=None
+        Input features. If `None`, then the names of all the variables in the
+        transformed dataset (original + new variables) is returned. If 'index',
+        then the name of the features created from the index is returned. If list
+        with feature names, the names for the datetime features derived from
+        the variables in the list will be returned."
+    """.rstrip()
```
solegalli, line 51: This doscstring is correct for the class datetimeFeatures that can obtain features from the index as well. All other transformers cannot.

I would make a generic docstring that applies to all classes that do not create features:

```
_input_features_docstring = """input_features: str, list, default=None
        Input features. If `None`, then the names of all the variables in the
        transformed dataset (original + new variables) is returned. If list
        with feature names, the features in the list will be returned."
    """.rstrip()
```

Also note that lines 46 to 50 need a "tab" indentation respect to current state, otherwise they appear bold in the documentation.

feature_engine/encoding/base_encoder.py
```diff
@@ -236,17 +239,44 @@ def inverse_transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
             The feature names.
         """
         check_is_fitted(self)
 
-        return self.feature_names_in_
+        if input_features is None:
+            # return all feature names
+            feature_names = self.feature_names_in_
+
+        else:
```
solegalli, line 265: this same functionality should be added in the base_transformer

feature_engine/base_transformers.py
```diff
@@ -154,9 +155,18 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+    )
+    def get_feature_names_out(
+            self, input_features: Optional[Union[List, str]] = None
+    ) -> List:
         """Get output feature names for transformation.
 
+        Parameters
+        ----------
+        {input_features}
+
```
solegalli, line 169: add functionality added in base_encoder to check input_features

feature_engine/imputation/base_imputer.py
```diff
@@ -87,17 +88,46 @@ def _get_feature_names_in(self, X):
...
         feature_names_out: list
             The feature names.
         """
         check_is_fitted(self)
 
-        return self.feature_names_in_
+        if input_features is None:
+            # return all feature names
+            feature_names = [
+                f for f in self.feature_names_in_ if f not in self.features_to_drop_
```
solegalli, line 113: I dont think these classes have a fetures_to_drop attribute

feature_engine/imputation/base_imputer.py
```diff
@@ -87,17 +88,46 @@ def _get_feature_names_in(self, X):
...
+            ]
+
+        else:
```
solegalli, line 116: we will be repeating this bit of code over and over, I wonder if it would be best to make an abstraction and then just call that function.

feature_engine/wrappers/wrappers.py
```diff
@@ -354,17 +364,30 @@ def get_feature_names_out(self) -> List:
...
-        if self.transformer_.__class__.__name__ in _TRANSFORMERS:
-            feature_names = self.feature_names_in_
+        if input_features is None:
+            if self.transformer_.__class__.__name__ in _TRANSFORMERS:
+                feature_names = self.feature_names_in_
 
-        if self.transformer_.__class__.__name__ in _CREATORS:
-            added_features = self.transformer_.get_feature_names_out(self.variables_)
-            feature_names = list(self.feature_names_in_) + list(added_features)
+            if self.transformer_.__class__.__name__ in _CREATORS:
```
solegalli, line 371: why the indentation?

solegalli (commented):
Hi @Morgan-Sell 

This is looking good thank you!

I have a couple of suggestions:

- let's not change the docstrings of the classes that create features, and as such, have dedicated get_feature_names_out code
- we need to add the checks for input features in all the classes you are working with, so it might be wise to create a function and call that function instead of repeating the code
- in estimator_checks.py there is a test for the get_feature_names_out that is then applied to all transformers we need to extend that test to do the following:

1. test that when user passes input_features it returns the names of the features passed by the user
2. when get_feature_names_out is called from a pipeline it returns the expected output.

For 2, we need to wrap the encoder with a pipeline and call get_feature_names_out from there

Thanks a lot for getting started on this!

feature_engine/imputation/base_imputer.py
Morgan-Sell: Make sense. Should it go in variable_manipulation.py?

Morgan-Sell:
hola @solegalli,

What do you mean by `let's not change the docstrings of the classes that create features, and as such, have dedicated get_feature_names_out code`?

Don't all the transformers create new features?

feature_engine/imputation/base_imputer.py
Morgan-Sell: When/why do we create `self.features_to_drop_` for certain classes and no others?

solegalli: that attribute is only for the feature selection classes

feature_engine/imputation/base_imputer.py
solegalli: > hola @solegalli,
> 
> What do you mean by `let's not change the docstrings of the classes that create features, and as such, have dedicated get_feature_names_out code`?
> 
> Don't all the transformers create new features?

Nope. Most transformers would transform the original features. Only transformers in the creation, timeseries and datetime modules create new features. And the one hot encoder.

Morgan-Sell: Ok, got it. We're all good. The `get_feature_names_out` docstrings for the creation, time-series, and datetime modules and one-hot encoder are the initial, non-generic docstrings.

Morgan-Sell:
Hola @solegalli,

We're getting there. I need to dig deeper into the check_estimator code. It seems like I incorrectly implemented the changes for `check_get_feature_names_out()`. Although, it's strange because when I execute `pytest estimator_checks.py`, the code returns all test passed.

feature_engine/base_transformers.py
```diff
@@ -154,17 +157,26 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+        input_features=_input_features_docstring,
+    )
+    def get_feature_names_out(
+            self, input_features: Optional[Union[List, str]] = None
+    ) -> List:
         """Get output feature names for transformation.
 
+        Parameters
+        ----------
+        {input_features}
```
solegalli, line 170: I think we could make the entire docstring (not just input_features) part of the `_input_features_docstring` docstring in the docstrings module. It is the same everywhere I think.

feature_engine/base_transformers.py
```diff
@@ -154,17 +157,26 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+
         Returns
         -------
         feature_names_out: list
             The feature names.
         """
         check_is_fitted(self)
-
-        return self.feature_names_in_
+        feature_names = check_input_features(input_features, self.variables_)
```
solegalli, line 178: we could probably just return the function without capturing it previously in a variable.

Have you read anything that says that one is better practice than the other?

feature_engine/estimator_checks.py
```diff
@@ -231,6 +238,17 @@ def check_get_feature_names_out(estimator):
                 "date2",
             ]
```
solegalli, line 240: We need to add same test as above (when input_features is None), but using pipe right here.

feature_engine/estimator_checks.py
```diff
@@ -231,6 +238,17 @@ def check_get_feature_names_out(estimator):
...
 
+            # user passes input features
+            features = ["var_3", "var_5", "var_7", "cat_var2", "date1"]
```
solegalli, line 242: I am not sure this is going to work, because when we set up the estimators, we are not defining the parameter `variables`. Also, we are making a test for all transformers, those that work with numerical and those that work with categorical, so we can't set the variables param universally for all. We need to think a bit what's the best way. What would you suggest?

feature_engine/parameter_checks.py
```diff
@@ -34,3 +34,76 @@ def _define_numerical_dict(dict_: Optional[dict]) -> Optional[dict]:
         raise TypeError("The parameter can only take a dictionary or None")
 
     return dict_
+
+
+def check_input_features(
```
solegalli, line 39: good place to add this function!

feature_engine/parameter_checks.py
```diff
@@ -34,3 +34,76 @@ def _define_numerical_dict(dict_: Optional[dict]) -> Optional[dict]:
...
+        input_features: Union[List, str], all_variables: List
+) -> Union[List, str]:
+    """
+    Checks if input_features is None; otherwise, check whether input_variables
+    is a list of features that exist within transformed dataset.
+
+    Parameters
+    ----------
+    input_features : Union[List, str]
```
solegalli, line 48: we don't use typehint in the docstrings. We normally write

`input_features: List, str, default=whatever_the_default is is any`

feature_engine/parameter_checks.py
```diff
@@ -34,3 +34,76 @@ def _define_numerical_dict(dict_: Optional[dict]) -> Optional[dict]:
...
+        The features to check whether they
+
+    all_variables: List
+        All the variables associated with the transformed dataset.
```
solegalli, line 52: maybe replace "associated" by "in"?

feature_engine/parameter_checks.py
```diff
@@ -34,3 +34,76 @@ def _define_numerical_dict(dict_: Optional[dict]) -> Optional[dict]:
...
+
+    Raises
+    ------
+    ValueError
+        If input_features is not a list or all the features are not included in the transformed
+        variables.
+
+    Returns
+    -------
+    res: bool
```
solegalli, line 62: The result is actually feature_names which is a list. Could we update the docstrings?

feature_engine/parameter_checks.py
```diff
@@ -34,3 +34,76 @@ def _define_numerical_dict(dict_: Optional[dict]) -> Optional[dict]:
...
+        # Return features requested by user.
+        if not isinstance(input_features, list):
+            raise ValueError(
+                f"input_features must be a list. Got {input_features} instead."
+            )
+        if _check_input_features_in_variables(input_features, all_variables):
+            raise ValueError(
+                "Some features in input_features were not used to extract new "
+                "variables. Pass either None, or a list with the features that "
+                "were used to create date and time features."
```
solegalli, line 81: We need to replace the error message:

"Some features in input_features were not transformed by this transformer. "
"Pass either None, or a list with the features that were transformed by this transformer."

feature_engine/parameter_checks.py
```diff
@@ -34,3 +34,76 @@ def _define_numerical_dict(dict_: Optional[dict]) -> Optional[dict]:
...
+    all_variables: List
+        All the variables associated with the transformed dataset.
+
+    Returns
+    -------
+    res: bool
+        Returns True if all the input_features are included in variables; otherwise,
+        returns False.
+    """
+    res = any([feature for feature in input_features if feature not in all_variables])
```
solegalli, line 108: this is just a one liner... is it worthy to make an entire function with docstrings and all? I think we can just add this line to the function above.

feature_engine/selection/base_selector.py
```diff
@@ -117,19 +119,43 @@ def _get_feature_names_in(self, X):
...
+        else:
+            if not isinstance(input_features, list):
+                raise ValueError(
+                    f"input_features must be a list. Got {input_features} instead."
+                )
+            if any([f for f in input_features if f not in self.variables_]):
+                raise ValueError(
+                    "Some features in input_features were not used to extract new "
+                    "variables. Pass either None, or a list with the features that "
+                    "were used to create date and time features."
```
solegalli, line 156: We need to change the wording as per previous comment.

feature_engine/wrappers/wrappers.py
```diff
@@ -357,15 +369,18 @@ def get_feature_names_out(self) -> List:
         if self.transformer_.__class__.__name__ in _TRANSFORMERS:
             feature_names = self.feature_names_in_
```
solegalli, line 370: we need to modify this behaviour based on input_features

feature_engine/wrappers/wrappers.py
```diff
@@ -357,15 +369,18 @@ def get_feature_names_out(self) -> List:
...
             added_features = self.transformer_.get_feature_names_out(self.variables_)
             feature_names = list(self.feature_names_in_) + list(added_features)
 
-        if self.transformer_.__class__.__name__ in _SELECTORS:
+        elif self.transformer_.__class__.__name__ in _SELECTORS:
             feature_names = [
                 f for f in self.feature_names_in_ if f not in self.features_to_drop_
             ]
 
+        else:
```
solegalli, line 381: we don't need this else, we need to modify the ifs and elifs behaviours.

If input_features is None, the behaviour is the current one, but if input_features is a list, we need to check that all the elements of that list are within self.variables_ and if yes, then return added_features passing input_features instead of self.variables_

solegalli (commented):
Hi @Morgan-Sell 

Enormous amount of work! Thank you very much. We are getting there true. This is looking really good.

I made a few suggestions for changes here and there, and then we need to expand the tests.

If you run pytest check_estimators.py all the tests will pass because in essence no test is run. Pytest only runs functions that start with the word "test" and there arent any in that file. 

the functions from check_estimator.py are called from the test files in the tests folder. And those tests are the one failing, and I tried to point out why they are failing in one of my comments.

If you can make the changes I suggest, and then try to expand the get_feature_names tests for the sklearn wrapper and the selectors, and see if you manage to implement a suggestion for the check_estimator.py, then we take it from there?

Thank you very much for all the work!

feature_engine/base_transformers.py
Morgan-Sell: Do you mean changing the above code to `return check_input_features(input_features, self.variables_)`?

I never came across anything that suggested creating a number variable vs executing a function in a `return` statement. Also, I did some Googling; I didn't find anything.

feature_engine/estimator_checks.py
Morgan-Sell: Do you mean to add the following code in the `if` statement associated w/ `# selection transformers`?

```
# transformer is used in a pipeline
assert pipe.get_feature_names_out(
    input_features=features
) == features
```

feature_engine/estimator_checks.py
Morgan-Sell: One way would be to create 3 global variables: 1) `_transformers_categorical`; 2) `_transformers_numerical`; and 3) `_transformers_categorical_and_numerical`.

I imagine that knowing whether a transformer is numerical or categorical will be important in other scenarios.

In the case of `check_get_feature_names_out`, we change the code to accommodate the 3 new global variables.

With regards to not defining the `variables` parameter, don't all of the classes create a `self.variables_` that correspond w/ the transformer - i.e., categorical, numerical, or both - when no `variables` parameter is provided when the class is instantiated?

feature_engine/parameter_checks.py
Morgan-Sell: Makes sense. I thought you wanted to create a function using the logic. I guess I misunderstood ;)

feature_engine/wrappers/wrappers.py
Morgan-Sell: Are you envisioning changing the above code from `feature_names = self.feature_names_in_` to `feature_names=self.get_feature_names_out(input_features=input_feature)`?

It looks like the classes included in `_TRANSFORMERS` are scikit-learn transformers. However, not all of the transformers, e.g. OrdinalEncoder, do have a `get_feature_names_out()` method.

Do we check if the class uses `get_feature_names_out()`? If the method exists we apply the abovementioned revised code and, if not, we keep `feature_names = self.feature_names_in_`.

feature_engine/wrappers/wrappers.py
Morgan-Sell: I tried to implement your suggestion. Hopefully, it is almost correct.

Morgan-Sell:
hola @solegalli,

Thanks for the feedback! I have a few follow-up questions and I still need to review some of my code. Hopefully, I'll get around to it in the next couple of days.

On a separate, but related note, why do not all classes have `get_feature_names_out`? For example, the method does not exist within BaseDiscretiser?

PR #414, opened 2022-04-14
organize docstrings into module and submodules

### LINKED ISSUE #396: organise docstrings into module
At the moment, most docstrings are contained in the _docstrings.py module.

We should create a folder called docstrings, and within it, create separate .py for attributes in the init, fit, transform etc.

We need to think if we also want to split it in imputers, encoders, etc. But we should probably centralize docstrings in this module instad of having it in base classes or within modules.

solegalli:
closes #396


solegalli:
FYI @Morgan-Sell 

Please rebase main onto your get_feature_names_out branch.

These changes are relevant for your PR.

PR #454, opened 2022-05-12
[Feature Creation] Decision tree creates a new feature by combining numerous variables

### LINKED ISSUE #107: create new features by combining variables with decision trees
New variables are created by combination of user indicated variables with decision trees. Example: if user passes 3 variables to transformer, a new feature will be created fitting a decision tree with this tree variables and the target.

To think about:
Should we make the transformer so that it combines variables in groups of 2s and 3s, etc? Say the user passes 5 variables, should we create features combining all possible groups of 2s, all possible groups of 3s, all possible groups of 4s and all 5?

Need to think a bit. I know that we do combine a few variables with trees to create new ones, particularly for use in linear models. But this brute force of combining everything with everything for the sake of combining, I have not seen in organisations where models will be used to score customers. So maybe not ideal. Also, increases computational cost, which is not in the spirit of feature-engine.

We want simple, understandable features, for solid, understandable models. Thoughts?

Morgan-Sell:
Closes #107 

Notes from #107: 

New variables are created by combination of user indicated variables with decision trees. Example: if user passes 3 variables to transformer, a new feature will be created fitting a decision tree with this tree variables and the target.

To think about:
Should we make the transformer so that it combines variables in groups of 2s and 3s, etc? Say the user passes 5 variables, should we create features combining all possible groups of 2s, all possible groups of 3s, all possible groups of 4s and all 5?

Need to think a bit. I know that we do combine a few variables with trees to create new ones, particularly for use in linear models. But this brute force of combining everything with everything for the sake of combining, I have not seen in organisations where models will be used to score customers. So maybe not ideal. Also, increases computational cost, which is not in the spirit of feature-engine.


Morgan-Sell:
halo @solegalli,

A few questions:

- Did we make a final decision on whether the class should create new features from all the possible permutations of the user-selected variables? I guess we could create an `all_permutations` init param. Although, I am not convinced of the param's value. I can't think of a use case; however, my experience is limited.
- Should the class allow the user to choose from all of the sklearn decision-tree init params - e.g. `max_depth` and `min_sample_leaf` - to prevent overfitting? Or, do we want to limit the user to 1 or 2 params?
- I plan to limit the variables to numerical; unless, the categorical variable is encoded. Do you agree?
- Should the class apply to both regression and classification?

solegalli:
An idea would be:

These parameters in the init:
variables = variable list (as always)
output_features = None, integer, list of integers, tuple

So if I pass three variables in the list: [var1, var2, var3] and:

- 1 in the output_features: return new features based on predictions of the three based on each variable individually, 3 new features
- 2: make all possible combinations of 2 variables: (var1, var2), (var1, var3), (var2,var3), 3 new features in this example
- 3 make all possible combinations of 3: in this case only 1 possible combination (var1,var2,var3), 1 new feature in this example
- 4 or greater raise an error, more combinations than variables in list

If i pass a list, say [1,2], then we return the output of 1 and 2 as above.
If None, then return all possible 1s, 2s, and 3s in this case, if the list contained more variables, then it would also be 4s and 5s

Alternatively, the user can pass a tuple with tuples (var1, (var,var2), (var1,var2,var3)) indicating how to combine the variables.

Morgan-Sell:
hola @solegalli,

Espero que estes disfrutando vacay!

When/why would a person apply the decision-tree transformer to one variable?

Morgan-Sell:
hi @solegalli,

how are you?! I guess you're busy writing your new book! Do you have any thoughts on my earlier comments?

Abrazo!

feature_engine/creation/decision_tree.py
```diff
@@ -0,0 +1,328 @@
...
+        drop_original: bool = False,
+    ) -> None:
+
+        if (
+            not isinstance(variables, list)
+            or not all(isinstance(var, (int, str)) for var in variables)
+            or len(set(variables)) != len(variables)
+        ):
+            raise ValueError(
+                "variables must a list of strings or integers comprise of "
```
solegalli, line 136: typo?

feature_engine/creation/decision_tree.py
```diff
@@ -0,0 +1,328 @@
...
+                f"distinct variables. Got {variables} instead."
+            )
+
+        if (
+                not isinstance(output_features, (int, list, tuple))
+                and output_features is not None
+        ):
+            raise ValueError(
+                f"output_features must an integer, list or tuple. Got "
```
solegalli, line 145: typo?

feature_engine/creation/decision_tree.py
```diff
@@ -0,0 +1,328 @@
...
+                f"{output_features} instead."
+            )
+
+        if isinstance(output_features, int):
+            if output_features > len(variables):
+                raise ValueError(
+                    "If output_features is an integer, the value cannot be "
+                    f"greater than the length of variables. Got {output_features} "
```
solegalli, line 153: number for lenght

feature_engine/creation/decision_tree.py
```diff
@@ -0,0 +1,328 @@
...
+                    "for output_features and {len(variables)} for the length "
```
solegalli, line 154: and {len(variables)} variables."

feature_engine/creation/decision_tree.py
```diff
@@ -0,0 +1,328 @@
...
+
+        if isinstance(output_features, list):
+            if (
+                    max(output_features) > len(variables)
+                    or not all(
+                            isinstance(feature, int) for feature in output_features
+                        )
+            ):
+                raise ValueError(
+                    "output_features must be a list solely comprised of integers "
```
solegalli, line 166: remove "solely comprised " and replace lenght by number

feature_engine/creation/decision_tree.py
```diff
@@ -0,0 +1,328 @@
...
+                    "and the maximum integer cannot be greater than the length of "
+                    f"variables. Got {output_features} for output_features and "
+                )
+
+        if isinstance(output_features, tuple):
+            num_combos = 0
+            for n in range(1, len(variables) + 1):
+                num_combos += len(list(combinations(variables, n)))
```
solegalli, line 174: could we add a small comment to explain what this is doing? I am so not familiar with itertools, it would help me.

feature_engine/creation/decision_tree.py
```diff
@@ -0,0 +1,328 @@
...
+            if (
+                    not all(
+                        isinstance(feature, (str, tuple)) for feature in output_features
+                    )
+                    or len(output_features) > num_combos
+            ):
+                raise ValueError(
+                    "output_features must a tuple solely comprised of tuples and "
```
solegalli, line 182: this error seems to contradict the previews error. Should output_features be a tuple of integers or a tuple of tuples? is it possible to put the 2 checks together?

feature_engine/creation/decision_tree.py
```diff
@@ -0,0 +1,328 @@
...
+                    f"the maximum number of tuples cannot exceed {num_combos}. Got "
+                    f"{output_features} instead."
+                )
+
+        if not isinstance(regression, bool):
+            raise ValueError(
+                f"regression must be a boolean value. Got {regression} instead."
+            )
+
+        if not isinstance(max_depth, int):
```
solegalli, line 192: it can also take None. I think we need to fix

feature_engine/creation/decision_tree.py
```diff
@@ -0,0 +1,328 @@
...
+            raise ValueError(
+                f"max_depth must be an integer. Got {max_depth} instead."
+            )
+
+        super().__init__(missing_value, drop_original)
```
solegalli, line 197: this class can't work with missing values, so we probably need to remove this.

feature_engine/creation/decision_tree.py
```diff
@@ -0,0 +1,328 @@
...
+            The training input samples. Can be the entire dataframe, not just
+            the variables to transform.
+
+        y: pandas Series or np.array = [n_samples,]
+            The target variable that is used to train the decision tree.
+        """
+        # common checks and attributes
+        # TODO: We don't need to check_variables b/c BaseCreation fit() performs action
+        X = super().fit(X, y)
+        self._check_dependent_variable_not_fitted_by_estimator(X, y)
```
solegalli, line 220: remove

feature_engine/creation/decision_tree.py
```diff
@@ -0,0 +1,328 @@
...
+
+        self.variable_combinations_ = self._create_variable_combinations()
+        self.variable_combination_indices_ = {}
+        self.fitted_estimators_ = {}
```
solegalli, line 224: do we need to expose this attribute? is it useful to the user to have a dictionary of fitted estimators? for brainstorming. Otherwise we should add an _ at the beginning.

feature_engine/creation/decision_tree.py
```diff
@@ -0,0 +1,328 @@
...
+        y: pandas Series or np.array = [n_samples,]
+            The target variable that is used to train the decision tree.
+        """
+        # common checks and attributes
+        # TODO: We don't need to check_variables b/c BaseCreation fit() performs action
+        X = super().fit(X, y)
+        self._check_dependent_variable_not_fitted_by_estimator(X, y)
+
+        self.variable_combinations_ = self._create_variable_combinations()
+        self.variable_combination_indices_ = {}
```
solegalli, line 223: what is the difference between these 2? if called indeces I expect integers, but I think these are variable names?

feature_engine/creation/decision_tree.py
```diff
@@ -0,0 +1,328 @@
...
+            else:
+                for num in self.output_features:
+                    variable_combinations += list(combinations(self.variables, num))
+
+            # transform all elements to lists to slice X dataframe
+            variable_combinations = [list(var) for var in variable_combinations]
+
+        return variable_combinations
+
+    def _create_new_features_names(self):
```
solegalli, line 300: for convention rename to get_feature_names_out() and it should also be able to take input_features=None.

It also returns a list instead of a dictionary, not sure how / if that could be made to fit with the current logic?

tests/test_creation/test_decision_tree.py
```diff
@@ -0,0 +1,125 @@
+import numpy as np
+import pandas as pd
+import pytest
+
+from feature_engine.creation import DecisionTreeCreation
+
+
+def test_create_variable_combinations(df_creation):
```
solegalli, line 8: great test!

solegalli (commented):
Hi @Morgan-Sell 

First of all, apologies for the delay. I know that it is frustrating when you put time into something and you don't get feedback soon enough. I am sorry. 

Not an excuse, but as a bit of an explanation, I felt Feature-engine was taking too much of my time, and I needed to balance it out a bit with those activities that actually pay my bills :p 

I'll try to answer your questions:

**Parameters for the tree**: max_depth is the most important, and we need to include the random_state. Maybe we leave the rest out for the first version of the transformer?

BaseCreation was designed thinking of MathFeatures and RelativeFeatures that can't take strings, and also need to check for duplication. I don't think we have such a function in the manipulation script. Maybe see how common the functionality is across these transformers and try to make a function?

I don't think BaseCreation is checking for variables in fit. Did you have a look?

I added some comments below. There is a lot going on, and to help auntie Sole understand, would you be so kind to add some short comments with what is expected from the code blocks? I put some hints where I needed some help to follow along.

I was thinking of using itertools, so you seem to be on the right track. I will have a look again and see if I can think of anything else.

Thanks a lot!

feature_engine/creation/decision_tree.py
```diff
@@ -0,0 +1,328 @@
...
+        X: pandas dataframe of shape = [n_samples, n_features]
+            The training input samples. Can be the entire dataframe, not just
+            the variables to transform.
+
+        y: pandas Series or np.array = [n_samples,]
+            The target variable that is used to train the decision tree.
+        """
+        # common checks and attributes
+        # TODO: We don't need to check_variables b/c BaseCreation fit() performs action
+        X = super().fit(X, y)
```
solegalli, line 219: In fact, BaseCreation checks for nan so not sure this base class is the best?

feature_engine/creation/decision_tree.py
```diff
@@ -0,0 +1,328 @@
...
+
+        If the user passes None, then all possible combinations will be
+        created. This is analogous to the user passing an integer that is equal
+        to the number of provided variables when the class is initiated.
+
+    regression: boolean, default = True
+        Select whether the decision tree is performing a regression or
+        classification.
+
+    max_depth: integer, default = True
```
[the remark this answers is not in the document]
Morgan-Sell: Type-o! It should've said 3 like it doe in `init()`

feature_engine/creation/decision_tree.py
```diff
@@ -0,0 +1,328 @@
...
+
+
+    """
+    def __init__(
+        self,
+        variables: List[Union[str, int]] = None,
+        # TODO: What is the correct typing hint?
+        output_features: Union[int, List[int], Tuple[tuple, ...]] = None,
+        regression: bool = True,
+        max_depth: int = 3,
```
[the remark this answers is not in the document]
Morgan-Sell: fixed. docstring was incorrect.

feature_engine/creation/decision_tree.py
Morgan-Sell: It was type-o

feature_engine/creation/decision_tree.py
Morgan-Sell: `self.variable_combinations` is a **list of lists** of all the variable combinations to be used in the decision tree based on the value of `output_features`.

`self.variable_combination_indices_` is a **dictionary** with a key that has a unique estimator id and a value that is one of the lists from `self.variable_combinations`.

Example:
```
{'estimator_0': ['Age'],
 'estimator_1': ['Marks'],
 'estimator_2': ['Avg_5k_run_minutes'],
 'estimator_3': ['Height_cm'],
 'estimator_4': ['Age', 'Marks'],
 'estimator_5': ['Age', 'Avg_5k_run_minutes'],
 'estimator_6': ['Age', 'Height_cm'],
 'estimator_7': ['Marks', 'Avg_5k_run_minutes'],
 'estimator_8': ['Marks', 'Height_cm'],
 'estimator_9': ['Avg_5k_run_minutes', 'Height_cm'],
 'estimator_10': ['Age', 'Marks', 'Avg_5k_run_minutes'],
 'estimator_11': ['Age', 'Marks', 'Height_cm'],
 'estimator_12': ['Age', 'Avg_5k_run_minutes', 'Height_cm'],
 'estimator_13': ['Marks', 'Avg_5k_run_minutes', 'Height_cm']}
```

`self._fitted_estimators_` is a dictionary with the same keys and values as fitted decision trees. The keys/indices enable the mapping between the variable combinations and their corresponding fitted decision trees.

For some reason, I wanted to use a dictionary and a list/iterable can't be a key. However, I'm now thinking that we merge `self.variables_combination_indices_` and `self._fitted_estimators' into a list of tuples. Each tuple is comprised of 2 elements: a variable combination and its corresponding fitted decision tree. For now, let's call the merged variable `variable_combos_and_fitted_estimators`.

The goal of `variable_combos_and_fitted_estimators` is to ensure that DecisionTreeFeatures uses the correct variables when using the fitted estimator to predict new values in `transform()`.

Morgan-Sell:
No pasa nada, @solegalli! Bills are known to impact one's priorities ;)

I apologize for the lack of notes in the init()! My type-os didn't help!

The user has more optionality with the `output_features` parameter than normal init params. I want to ensure we put them on the right path early in their feature-engine adventure. Consequently, I created all of those checks.

I'm still working on incorporating all of your comments. Also, I will simplify the code by merging `self.variables_combination_indices_` and `self._fitted_estimators_` into a list of tuples.

solegalli:
> hen/why would a person apply the decision-tree transformer to one vari

Just catching up with old questions: if the variable has a non-linear relationship with a target, transforming that variable with a decision tree would, at least, create a monotonic relationship between the transformed variable and the target. Useful in cases where you want to stick with linear models. This is a good point to add to the user guide. Thanks for asking!

feature_engine/creation/decision_tree.py
solegalli: - I think having variable_combinations_ and variable_combinaction_indexes_ is reduntant.  
- Thinking out loud: can we have an attribute called `output_features_` that contains a tuple of tuples with the variable combinations (the combos). And then the hidden attribute `_estimators` (no need to add `fitted`) with a list of estimators fitted for those variable groups?

That might simplify things and also shorten the name of the attributes, that always helps :p

solegalli:
> ser has more option

Great! Thank you so much! give me a shout when it's done :)

feature_engine/creation/decision_tree.py
Morgan-Sell: BaseTransformer does too. I changed the base classes to BaseEstimator and TransformerMixin.

Morgan-Sell:
hi @solegalli,

In one of your comments, you said that decision trees can handle categorical variables. I thought sklearn's decision tree can handle categorical variables if they are one-hot encoded. Consequently, all variables must be numerical. If so, should the class find only numerical variables?

See [sklearn docs](https://scikit-learn.org/stable/modules/tree.html). Excerpt: _"Able to handle both numerical and categorical data. However scikit-learn implementation does not support categorical variables for now."_

Either way, I think we should find the variables (either numerical or numerical/categorical) in the init(). This will allow the class to check whether the user provided the correct values for `output_features` when the param is a tuple and the user provides variable names. Otherwise, the code compares the `output_features` and `None`.

solegalli:
> 

I think this is just too complicated. We offer a lot of functionality to encode categorical variables. It is very easy for the user to add a categorical encoder ahead of using this transformer. And they can encode the variables however they like. So I would not go down this route.

I would stick with sklearn functionality. They only support numerical variables.

Morgan-Sell:
hola @solegalli,

I think I'm heading down a rabbit hole. Before I do, I would like to hear your thoughts.

The `init()` performs a few checks to confirm that `output_features` is compatible with `variables`. E.g., Is the size of the variable combination compatible with the number of variables provided in the `init()`?

When `variables` is not `None`, the code is working. However, when `variables` is `None`, then `output_features` and `variables` cannot be compared because `_find_or_check_numerical_variables` has been executed.

Should we implement the `output_features` checks in `fit()` so the code only performs the checks once, i.e., after `self.variables_` is created? If so, would the best approach be to create `_check_output_features_is_permittied()` that performs all necessary checks?

solegalli:
Hi @Morgan-Sell 

I made a PR: https://github.com/Morgan-Sell/feature_engine/pull/10

with some changes. I think I answer your question with the changes in the logic. If in doubt, simpler is better ;)

solegalli:
Actually, I don't think I've done a great job, now that I am reading your question more carefully.

Let's move all the tests for output_features to the fit method.

Maybe create a hidden method called `_validate_strategy()`, that contains all relevant checks. And then we run _validate_strategy() in fit, after the _find_or_check_numerical_variables().

Good catch, well done @Morgan-Sell 

Thank you very much.

solegalli:
Hi @Morgan-Sell 

I've seen you made a lot of commits. Is this work in progress? Do you still need to update the tests? They are all failing :_(

I am on holidays from Thursday till August. So if you don't hear from me... you know why ;)

Cheers

Morgan-Sell:
hi @solegalli.

Ahh.... a month-long vacation! Hopefully, the US will adopt such traditions one day ;)

I'm still working on this class. I do have one question.

The following test is failing:
```
FAILED tests/test_creation/test_check_estimator_creation.py::test_check_estimator_from_sklearn[estimator6] - ValueError: No numerical variables found in this dataframe. Please check variable format with pandas dtypes.
```

Do you know if sklearn's `check_estimator` tests a dataframe without numerical variables? [check_estimator docs](https://scikit-learn.org/stable/modules/generated/sklearn.utils.estimator_checks.check_estimator.html) has limited information.

If so, then the DecisionTreeFeatures should raise an error. Other feature-engine classes skip certain sklearn checks, which makes sense given that all of sklearn's checks are unlikely to be appropriate for each class. How do we select to omit certain tests?

Morgan-Sell:
Hi @solegalli,

I'm embarrassed to say this, but I'm stumped by these errors. Hopefully, we can discuss the errors when you return.

Issue #428, opened 2022-04-25
`OrdinalEncoder` could output -1 for unseen categories


solegalli:
The OrdinalEncoder has an errors argument which can either raise an error or output NaNs when encountering new categories. For this particular class, it'd make sense to output -1 when a new category is encoded instead of generating NaNs.

datacubeR:
@solegalli I was taking a look at this Issue and I noticed `transform` method is inherited from `CategoricalMethodsMixin`. So in order to fix this, a whole transform method should override the one in `CategoricalMethodsMixin`? Or is there a better way to do this?

solegalli:
We did this for the CountEncoder already. I think the best would be to have a look at how we implemented there?

PR #456, opened 2022-05-12
add option to encode unseen categories with zeros for `CountFrequencyEncoder`

### LINKED ISSUE #417: `CountFrequencyEncoder` could output zeros for new categories
The `CountFrequencyEncoder` has an `errors` argument which can either raise an error or output NaNs when encountering new categories. For this particular class, it'd make sense (perhaps even as a default behavior) to output zeros when a new category is encoded instead of generating NaNs.

david-cortes:
Fixes #417 

This PR adds an option for `CountFrequencyEncoder` to output an encoding result of zero when calling `transform` on new data that has categories that were not present in the training data, which makes sense for this particular class since unseen categories would have a count of zero by definition.


solegalli:
Will have a look when I am back from hols, apologies for the delay

feature_engine/encoding/count_frequency.py
```diff
@@ -145,15 +148,18 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
         self._get_feature_names_in(X)
 
         self.encoder_dict_ = {}
+        dct_init = defaultdict(lambda: 0) if self.errors == "zeros" else {}
 
         # learn encoding maps
         for var in self.variables_:
             if self.encoding_method == "count":
-                self.encoder_dict_[var] = X[var].value_counts().to_dict()
+                self.encoder_dict_[var] = X[var].value_counts().to_dict(dct_init)
```
solegalli, line 156: what does this do?

solegalli (commented):
Hi @david-cortes 

Thanks a lot for the changes! And apologies for the delay. I am just back from holidays and had to fix those failing tests as they were blocking all PRs.

On that front, could you please rebase the lastest version of main to this PR?

I am thinking of expanding the options to encode unseen data to more encoders:
CountFrequencyEncoder by 0 (this PR)
OrdinalEncoder by -1 (this issue #428)
TargetEncoder by mean of target for all observations (no issue yet)
WoE and PR we need to think.

So instead of calling it "zeroes" could we call the parameter "encode"?
I add more details throughout the code.

Please let me know your thoughts and if you could implement this. Thank you!

david-cortes:
Redone with the new base classes for categoricals.

feature_engine/encoding/count_frequency.py
david-cortes: It starts from a base dictionary which gets updated with the counts.

feature_engine/encoding/count_frequency.py
```diff
@@ -1,6 +1,7 @@
 # Authors: Soledad Galli <solegalli@protonmail.com>
 # License: BSD 3 clause
 
+from collections import defaultdict
```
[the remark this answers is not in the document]
david-cortes: It's a different type of dictionary that outputs zeros for entries that it doesn't know. I think it is a very good candidate to use for this PR since it makes it involve very little changes, and it's probably faster than doing checks after encoding.

feature_engine/encoding/base_encoder.py
```diff
@@ -184,31 +184,31 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
         # check if NaN values were introduced by the encoding
-        if X[self.encoder_dict_.keys()].isnull().sum().sum() > 0:
-
-            # obtain the name(s) of the columns have null values
-            nan_columns = (
-                X[self.encoder_dict_.keys()]
-                .columns[X[self.encoder_dict_.keys()].isnull().any()]
-                .tolist()
-            )
+        if self.errors != "zeros":
```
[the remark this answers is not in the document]
david-cortes: Changed it for this particular transformer.

feature_engine/encoding/count_frequency.py
```diff
@@ -145,15 +148,18 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
...
         for var in self.variables_:
             if self.encoding_method == "count":
-                self.encoder_dict_[var] = X[var].value_counts().to_dict()
+                self.encoder_dict_[var] = X[var].value_counts().to_dict(dct_init)
 
             elif self.encoding_method == "frequency":
                 n_obs = float(len(X))
-                self.encoder_dict_[var] = (X[var].value_counts() / n_obs).to_dict()
+                self.encoder_dict_[var] = (
+                    X[var].value_counts() / n_obs
```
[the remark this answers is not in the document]
david-cortes: Changed.

feature_engine/encoding/base_encoder.py
```diff
@@ -79,10 +79,14 @@ def __init__(
         variables: Union[None, int, str, List[Union[str, int]]] = None,
         ignore_format: bool = False,
         errors: str = "ignore",
+        supports_errors_encode: bool = False
```
solegalli, line 82: hi @david-cortes 

We can't really add parameters unless they are strictly necessary, because that turns classes less user friendly.

I am trying to think what would be the best way forward, given that at the moment, all classes take "raise" and "ignore" and we want the Count frequency to take the extra parameter.

I think, probably the best way is to take this functionality out of the base class and into a function, that takes the allowed strings as inputs.

Unless you have a different idea?

feature_engine/encoding/_docstrings.py
```diff
@@ -15,9 +15,15 @@
...
 _errors_docstring = """errors: string, default='ignore'
         Indicates what to do, when categories not present in the train set are
-        encountered during transform. If 'raise', then rare categories will raise an
-        error. If 'ignore', then rare categories will be set as NaN and a warning will
-        be raised instead.
+        encountered during transform. If 'raise', then unseen categories will raise an
+        error. If 'ignore', then unseen categories will be set as NaN and a warning will
+        be raised instead. If 'encode', then unseen categories will be encoded according
+        to the default strategy from the transformer, provided that it supports it.
+    """.rstrip()
```
solegalli, line 22: At the moment, most classes accept only "raise" and "ignore". So I would not change the original docstring.

feature_engine/encoding/_docstrings.py
```diff
@@ -15,9 +15,15 @@
...
+
+_errors_docstring_with_encode = _errors_docstring + """
```
solegalli, line 24: at the moment, only the count frequency has the "encode" extra functionality. And when we expand this to other classes, each one will do something different. So we would probably have to right independent docstrings for each one of them. I am not sure that having a centralized copy is suitable in this case. I would just re-word it in the class.

tests/test_encoding/test_count_frequency_encoder.py
```diff
@@ -145,20 +146,52 @@ def test_error_if_input_df_contains_categories_not_present_in_fit_df(
...
+        encoder.fit(df_enc)
+        encoder.transform(df_enc_rare)
+
 
 def test_fit_raises_error_if_df_contains_na(df_enc_na):
     # test case 4: when dataset contains na, fit method
-    with pytest.raises(ValueError):
-        encoder = CountFrequencyEncoder()
-        encoder.fit(df_enc_na)
+    for errors in ["raise", "ignore", "encode"]:
```
solegalli, line 159: why do we need this change? The NA check comes before the unseen categories check.

tests/test_encoding/test_count_frequency_encoder.py
```diff
@@ -145,20 +146,52 @@ def test_error_if_input_df_contains_categories_not_present_in_fit_df(
...
+            encoder.fit(df_enc_na)
 
 
 def test_transform_raises_error_if_df_contains_na(df_enc, df_enc_na):
     # test case 4: when dataset contains na, transform method
-    with pytest.raises(ValueError):
-        encoder = CountFrequencyEncoder()
-        encoder.fit(df_enc)
-        encoder.transform(df_enc_na)
+    for errors in ["raise", "encode"]:
```
solegalli, line 167: interesting, here you skipped the "ignore", so I guess, "ignore" would ignore both already existing nan and the newly introduced.... it feels like a code smell in our original source code...

solegalli (commented):
Hi @david-cortes 

Thank you very much for the code changes.

The code looks good, and I think it could be ready to go. 

Your tests made me think that there might be something weird with how we handle existing NAN and newly introduced NaN, I would like to hear your thoughts when creating the tests?

Other than that, my main concern is around adding a new parameter to the base class. I would prefer we did not do that.

I suggested replacing the functionality in the main class by an external function. That would mean changing slightly the code in all other encoding classes. I would like to hear your thoughts on this. Maybe you have a better idea?

thanks a lot!

feature_engine/encoding/_docstrings.py
david-cortes: But not all of the clases would end up supporting the "encode" option.

feature_engine/encoding/base_encoder.py
david-cortes: Yes, that also would do.

tests/test_encoding/test_count_frequency_encoder.py
david-cortes: It should still fail when passing "encode" if the fit input has NAs.

david-cortes:
I am not sure what would be the right way to handle explicit missing values that are not from unseen categories.

I think ideally whether to handle them should be configured through a different option than this one, and ideally one of the ways of handling them would be to encode them into a separate category "missing" or similar. However pandas functions and methods don't lend themselves well towards handling missing values and that would require a lot more modifications.

feature_engine/encoding/_docstrings.py
david-cortes: Fixed.

solegalli:
Hi @david-cortes 

Sorry for the delay. Please see this PR: https://github.com/david-cortes/feature_engine/pull/1

I mostly reorganized the code you wrote.

The conflict is probably through the main branches not being in sync. If you can't resolve it let me know.

Thank you so much for the great contribution!

david-cortes:
Updated, but I'm not sure if the merge conflict solving I made here preserved the changes that you wanted.

feature_engine/base_transformers.py
```diff
@@ -8,17 +8,16 @@
 from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.utils.validation import check_is_fitted
```
solegalli, line 9: can we remove this file from the PR? I think it got accidentally the indentation changed.

feature_engine/encoding/base_encoder.py
```diff
@@ -2,28 +2,22 @@
...
-    _check_contains_na,
-    _check_X_matches_training_df,
-    check_X,
-)
-from feature_engine.encoding._docstrings import (
-    _errors_docstring,
-    _ignore_format_docstring,
-    _variables_docstring,
-)
+from feature_engine.dataframe_checks import (_check_contains_na,
```
solegalli, line 7: can we revert to the original indentation please? so it is coherent with the rest of the codebase?

feature_engine/encoding/base_encoder.py
```diff
@@ -78,14 +72,9 @@ def __init__(
         self,
         variables: Union[None, int, str, List[Union[str, int]]] = None,
         ignore_format: bool = False,
-        errors: str = "ignore",
+        errors: str = "ignore"
```
solegalli, line 75: can we please keep the comma after the parameter?

tests/test_encoding/test_count_frequency_encoder.py
```diff
@@ -1,5 +1,8 @@
+import warnings
+
 import pandas as pd
 import pytest
+from numpy import nan
```
solegalli, line 5: can we use isort here?

solegalli (commented):
Hi @david-cortes 

Yes, this is looking great.

Unfortunately, the codestyle change. We've got an indentation pattern that we do not normally use in the codebase.

Could you please have a look and change it back to what it looked like in my PR?

Thank you!

david-cortes:
@solegalli What linter are you using for indentation?

solegalli:
black

david-cortes:
Reformatted with black and isort.

solegalli:
Hi @david-cortes 

Not sure all the files were reformatted. Also, can we remove the base transformer from this PR?

Thank you!

david-cortes:
If I run either `black` or `isort` on this package, it will reformat pretty much all the files, not just the ones connected with this PR.

david-cortes:
The base class `CategoricalInitExpandedMixin` is used by other transformers too.

david-cortes:
Updated.

solegalli:
Thank you! merging as we speak.

Issue #457, opened 2022-05-12
Potentially unneeded data copying


david-cortes:
There is a function `check_X` used throughout the library which produces a copy of the data in order to avoid potential in-place modifications:
https://github.com/feature-engine/feature_engine/blob/248465e111416f88a4784b5e2e13a83d47e1f00e/feature_engine/dataframe_checks.py#L50

This is inefficient since it generates an extra copy of the whole data, and as a given transformer such as `CountFrequencyEncoder` processes that data further, a third copy will be generated until it replaces the earlier copy. It is possible to modify data frames column-by-column without having to produce an extra copy of the data and without affecting the original object that is passed by the user by using `pd.DataFrame.assign`, which could potentially result in a significant decrease in memory usage in large datasets and a small speed up.

solegalli:
Hi @david-cortes 

Do you have some code that compares the time implementation of copy() vs assign() that you could share? or a link to more details about the efficiency of the 2 methods?

david-cortes:
I don't have any comparison, but the underlying issue is rather simple: `.copy()` makes a deep copy of the whole data, `assign` updates it column-by-column without creating a full copy.

david-cortes:
Just learned that nowadays pandas's `assign` also copies the data underneath: https://github.com/pandas-dev/pandas/blob/4bfe3d07b4858144c219b9346329027024102ab6/pandas/core/frame.py#L4509

Thus, using `assign` would not actually accomplish anything here.

solegalli:
Thanks @david-cortes for following this through!

hkothari:
Hi @solegalli I was just wondering if you could elaborate on the concern which triggered you to put into the `.copy()` call. Is it out of an abundance of caution or do you have transformers which actually do mutate the given DataFrame? I ask because we're using `feature_engine` in a production use case in a low latency situation and from our profiling the copies actually take up a substantial amount of time. 

Generally speaking unless you're using inplace operations, I would assume that no changes you make can actually modify the DF anyways and it seems like inplace has fallen out of vogue lately anyways because under the hood for most use cases it actually does a copy so the perf is effectively the same. If you are doing inplace operations (which from a quick search it seems like maybe there are some), you could downscope the copies to just those use cases perhaps?

Am I totally off the mark here or is this something that might be feasible. If so, I'd be happy to take a pass at putting together a PR for this.

solegalli:
Hi @hkothari 

[...]

PR #458, opened 2022-05-12
Add new transformer `MatchCategories` to preprocessing module

### LINKED ISSUE #418: Would be useful to have a simple categorical levels encoder
It would be useful if there could be a class that would encode `object` / `category` and similar dtypes into `pd.Categorical`, while remembering the categories / encodings / levels that were used when calling `transform` later, but without applying any further transformation like one-hot (akin to R's `factor(data, levels)`).

david-cortes:
Fixes #418

This PR introduces a transformer `CategoryEncoder` which simply encodes columns as categorical dtype remembering the encodings / levels that were used for each variable.

As per the issue comments, I put it under `preprocessing`, but it is based on the base categorical encoder that the transformers in `encoding` use.

There's a few things that I was not sure about:
* I see that some modules use absolute imports (`from feature_engine... import ...`). I am not sure if that was intentional or not. This PR uses relative imports for internal objects.
* I see some transformers modify the data directly by using `df[column]`, creating an extra copy of the data beforehand which is inefficient (opened an issue about it: https://github.com/feature-engine/feature_engine/issues/457). For this PR instead I used `df.assign`, but it's still incurring an extra copy due to calling the input-checking methods.


solegalli:
Hi @david-cortes 

Of the top of my head, the .copy() is intentional and not to over-write the users dataframe.

I am away for 2 weeks, so will have a look when I am back. Apologies.

feature_engine/encoding/base_encoder.py
```diff
@@ -39,22 +39,32 @@ class BaseCategoricalTransformer(BaseEstimator, TransformerMixin):
...
+
+    allow_missing: bool
+        Whether to allow missing values in the columns handled by the transformer.
     """
 
     def __init__(
         self,
         variables: Union[None, int, str, List[Union[str, int]]] = None,
         ignore_format: bool = False,
+        allow_missing: bool = False
```
solegalli, line 51: If we add a parameter to this class, this parameter will be available to all encoders. Can we not add the param to the class that will use it only?

The least parameters in a class the better (more user friendly) so I would not want to add unneeded param to all transformers.

feature_engine/encoding/base_encoder.py
```diff
@@ -183,6 +195,13 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
                 else:
                     X[feature] = X[feature].astype("float")
 
+        self._check_nas_in_result(X)
+        return X
+
+    def _check_nas_in_result(self, X):
```
solegalli, line 201: this method is only used in the new class, correct? can we not place it there instead?

solegalli (commented):
Hi @david-cortes 

Thanks a lot for the contribution. 

I think the new class looks good. I am not completely sure that I would like to modify the base categorical class much, that would have a knock on effect on all encoders. Could we move the methods and params to the new transformer instead?

If there are bits of the base class that should be avoided, maybe we can think of breaking the methods down more and see how to make it fit.

Also, CategoryEncoder is a bit misleading, this transformer does not change the values of the variables, just the type. So I would call it MatchCategories or something on those line to keep the spirit of the preprocessing module and not to induce users to think that they will have the values of their categories changed into numbers (aka encoding).

Let me know your thoughts.

feature_engine/preprocessing/category_encoder.py
```diff
@@ -0,0 +1,121 @@
...
+
+        Returns
+        -------
+        X_new: pandas dataframe of shape = [n_samples, n_features].
+            The dataframe containing variables encoded as pandas categorical dtype.
+        """
+        X = self._check_transform_input_and_state(X)
+
+        for feature, levels in self.encoder_dict_.items():
+            X = X.assign(**{feature: pd.Categorical(X[feature], levels)})
```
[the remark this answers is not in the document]
david-cortes: In this particular case, there's little difference, but if the `check_X` function were changed so as not to copy the data, this would avoid modifying the input in place. See https://github.com/feature-engine/feature_engine/issues/457

feature_engine/encoding/base_encoder.py
david-cortes: Issue is, most of what the transformers do happens in the base class. If it were to be done outside of the base class, if would involve duplicating a lot of code.

solegalli: Can we not make this new class inherit the BaseCategorical and modify the functionality in the new class instead of in the base class? and if not, maybe make a child class of BaseCategorical and then modify that one instead?

BaseCategorical is used by all categorical transformers. So we can't really add a parameter that is only going to be used by one class. We need to find an alternative way.

feature_engine/preprocessing/category_encoder.py
```diff
@@ -0,0 +1,121 @@
...
+    dtype, or to prevent unseen categories from reaching a further transformer
+    or estimator in some pipeline, for example.
+
+    Parameters
+    ----------
+    {variables}
+
+    {ignore_format}
+
+    {errors}
```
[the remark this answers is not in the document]
david-cortes: Yes, it also has to decide whether to output missing values for unseen categories or to fail when it encounters them.

feature_engine/encoding/base_encoder.py
david-cortes: Issue is, since the part that differs is right in the middle of the base class' methods, adding a new subclass would involve duplicating lots of the code. Not sure if that's Ok for you.

solegalli:
Hi @david-cortes 

FYI this PR: https://github.com/david-cortes/feature_engine/pull/2

restructures the base encoder so that you can inherit in this new class without adding the extra parameters.

Hope this helps

Cheers!

david-cortes:
Updated, but I'm again not very sure if this is how you intended the changes to be.

feature_engine/encoding/base_encoder.py
```diff
@@ -39,22 +39,32 @@ class CategoricalInitMixin:
     {variables}.
 
     {ignore_format}
+
+    allow_missing: bool
```
solegalli, line 43: this parameter should only be added to the MatchColumns class

feature_engine/encoding/base_encoder.py
```diff
@@ -124,7 +138,8 @@ def _check_or_select_variables(self, X: pd.DataFrame):
             self.variables_ = _find_all_variables(X, self.variables)
 
         # check if dataset contains na
-        _check_contains_na(X, self.variables_)
+        if not self.allow_missing:
```
solegalli, line 141: I created the _fit method and removed the check_na from this method, so you can use it in MatchVariables.

In short, this logic that you are using here, you need to move it to MatchVariables.

feature_engine/encoding/base_encoder.py
solegalli: this method needs to be moved to MatchVariables

feature_engine/preprocessing/match_categories.py
```diff
@@ -0,0 +1,159 @@
...
+    """
+
+    def __init__(
+        self,
+        variables: Union[None, int, str, List[Union[str, int]]] = None,
+        ignore_format: bool = False,
+        errors: str = "raise",
+    ) -> None:
+        check_parameter_errors(errors, ["ignore", "raise"])
+        super().__init__(variables, ignore_format, allow_missing=errors != "raise")
```
solegalli, line 83: the logic for missing values needs to be moved here.

Also, we normally call this parameter "missing" and takes values "raise" and "ignore". So for consistency, could we use the same names?

feature_engine/preprocessing/match_categories.py
```diff
@@ -0,0 +1,159 @@
...
+        ----------
+        X: pandas dataframe of shape = [n_samples, n_features]
+            The training dataset. Can be the entire dataframe, not just the
+            variables to be transformed.
+
+        y: pandas Series, default = None
+            y is not needed in this encoder. You can pass y or None.
+        """
+        X = check_X(X)
+        self._check_or_select_variables(X)
```
solegalli, line 100: after this method, you need to add the logic if self.missing=="raise": then do something

solegalli (commented):
Hi @david-cortes 

Thank you for the changes. 

We are almost there.

I introduced some changes in the base_encoder, so that you can implement the check for missing data **only** in the class you are creating.

So please, **do not add more parameters or methods** to the base_encoder. Whatever MatchColumns needs, needs to be just in that class.

Do you think you could do a few more changes to accommodate that?

See my comments throughout the code.

Thank you!

david-cortes:
Redid the PR based on the current master branch - please take a look to see if this is how the relationships between classes were meant to be.

feature_engine/preprocessing/match_categories.py
david-cortes: In the other categorical encoders (such as `CountFrequency`), it it called "errors".

david-cortes:
Actually, from a look at the other commits that you had pushed for this PR, I'm now noticing that it differed a bit from the current master branch and have some confusion over which changes should be kept. Were methods like `_fit` intended to be used for this PR, or were they meant to be removed from an earlier PR?

solegalli:
Hi @david-cortes 

Did you had a chance to have the look at the changes I made in the PR I made to your repo? Because there you would have a better view of what was "new".

I modified the base_encoder so that you could accommodate the allow_missing functionality. For this reason, I created the _fit() method, that would be used by all encoders, and left the _check_categorical_variables() without the check for na, so that you could use it in this class, without re-writing all the code again.

Most of my changes have disappeared now :_(

I suggest you take a look at the PR you merged to your repo, hopefully you will see what was meant to be retained. If not, maybe the best would be to checkout the branch from which I made the PR to your repo and start from there.

Hope this helps
Cheers

david-cortes:
Redone based on the earlier branch, but this time removing the additional parameter in the base class. Although, I am still not sure that `_check_nas_in_result` is better left as a duplicated piece of code in `MatchCategories` instead of a method in the base class.

solegalli:
Hi @david-cortes 

It's looking good. We need to add the documentation now. We need to call this class from:

- docs/index
- docs/api/index
- docs/user_guide/index

and within the api and user_guide folders we need to add an rst file, in the first one just to retrieve the docstrings, and in the second one with a demo of the class.

Would that be ok? You could look at the ones from MatchColumns as inspiration.

david-cortes:
Added examples.

PR #474, opened 2022-06-12
fix bug for python 3.10 tests in CI

solegalli:
FYI @SangamSwadiK


SangamSwadiK:
> FYI @SangamSwadiK

I'll check more on this.

PR #476, opened 2022-06-13
fix numpy typing error with version 1.22 and above

solegalli:
FYI
@SangamSwadiK


solegalli:
I am really puzzled. mypy works locally with numpy 1.22.4 using np.typing. But in circleci, even though the numpy version seems to be the same, it does not find the module typing.

solegalli:
@SangamSwadiK 

I am inclined to merge these changes. This resolved mypy failing in my local environment.

Issue #429, opened 2022-04-25
`CountFrequencyEncoder` could have a parameter to group categories with few observations


solegalli:
Useful to handle rare categories in highly cardinal variables.

If a category is present in less than a certain threshold of observations it should be replaced by a certain value. Check page 91 of Alice Zhengs book and the CountEncoder from category encoders.

SangamSwadiK:
Hey, @solegalli, I would like to work on this issue if possible.

solegalli:
Great! Go for it!

Do you have an idea of how to go about it? or shall we discuss here first?

SangamSwadiK:
> Great! Go for it!
> 
> Do you have an idea of how to go about it? or shall we discuss here first?

Hi, Is the book name "Feature engineering for machine learning" ?, I think ill be able to better understand and discuss about it after going through the book.

solegalli:
This is the book:
https://www.amazon.de/-/en/Alice-Zheng/dp/1491953241

solegalli:
Yes, that is correct :)

SangamSwadiK:
Okay Cool !
Ill do it and submit a PR.

PR #480, opened 2022-06-15
(In progress) Backoff binning rare labels in CountFrequency encoder

SangamSwadiK:
**Reference Issues/PRs**:
Issue : #429  (Add functionality for threshold)

**What does this implement/fix?**
Adds functionality to group categories with few observations.

Hi, @solegalli I've added functionality for Backoff binning, threshold is a list that takes values(threshold) for each feature mentioned and bins them accordingly. I haven't added the tests yet but will add them.

Please let me know any improvements that you think could be done here. I'll implement them.

### DIFF
--- a/feature_engine/encoding/count_frequency.py
+++ b/feature_engine/encoding/count_frequency.py
@@ -78,6 +78,10 @@ class CountFrequencyEncoder(CategoricalInitExpandedMixin, CategoricalMethodsMixi
 
     {errors}
 
+    threshold: int,float or None, default=None
+        With threshold you have the option to perform backoff binning for all
+        columns for the given threshold.
+
     Attributes
     ----------
     encoder_dict_:
@@ -121,21 +125,44 @@ class CountFrequencyEncoder(CategoricalInitExpandedMixin, CategoricalMethodsMixi
         variables: Union[None, int, str, List[Union[str, int]]] = None,
         ignore_format: bool = False,
         errors: str = "ignore",
+        threshold: Union[int, float, None] = None,
     ) -> None:
 
         if encoding_method not in ["count", "frequency"]:
             raise ValueError(
                 "encoding_method takes only values 'count' and 'frequency'"
             )
-        super().__init__(variables, ignore_format, errors)
 
+        if (
+            encoding_method == "count"
+            and not (isinstance(threshold, int))
+            and threshold is not None
+        ):
+            raise ValueError(
+                "threshold expects int for encoding=count but got {0}".format(
+                    type(threshold)
+                )
+            )
+
+        if (
+            encoding_method == "frequency"
+            and not (isinstance(threshold, float))
+            and threshold is not None
+        ):
+            raise ValueError(
+                "threshold expects float for encoding=frequency but got {0}".format(
+                    type(threshold)
+                )
+            )
+
+        super().__init__(variables, ignore_format, errors)
         self.encoding_method = encoding_method
+        self.threshold = threshold
 
     def fit(
         self,
         X: pd.DataFrame,
         y: Optional[pd.Series] = None,
-        threshold: Optional[List[Optional[float]]] = None,
     ):
         """
         Learn the counts or frequencies which will be used to replace the categories.
@@ -158,40 +185,72 @@ class CountFrequencyEncoder(CategoricalInitExpandedMixin, CategoricalMethodsMixi
 
         self.encoder_dict_ = {}
 
-        if not isinstance(threshold, list):
-            threshold = [None] * len(self.variables_)
-
-        # learn encoding maps
-        for var, threshold in zip(self.variables_, threshold):
-
-            if (self.encoding_method == "count") and threshold:
+        if self.threshold is not None and self.encoding_method == "count":
+            # learn encoding maps
+            for var in self.variables_:
                 temp = {"backoff_bin": 0}
                 for k, v in X[var].value_counts().to_dict().items():
-                    if v <= threshold:
+                    if v <= self.threshold:
                         temp["backoff_bin"] = temp["backoff_bin"] + v
                     else:
                         temp[k] = v
-
                 self.encoder_dict_[var] = temp
 
-            elif (self.encoding_method == "frequency") and threshold:
+        elif self.threshold is not None and self.encoding_method == "frequency":
+            for var in self.variables_:
                 temp = {"backoff_bin": 0}
                 n_obs = float(len(X))
                 for k, v in (X[var].value_counts() / n_obs).to_dict().items():
-                    if v <= threshold:
+                    if v <= self.threshold:
                         temp["backoff_bin"] = temp["backoff_bin"] + v
                     else:
                         temp[k] = v
-
                 self.encoder_dict_[var] = temp
 
-            elif self.encoding_method == "count":
-                self.encoder_dict_[var] = X[var].value_counts().to_dict()
+        else:
+
+            if self.encoding_method == "count":
+                for var in self.variables_:
+                    self.encoder_dict_[var] = X[var].value_counts().to_dict()
 
             elif self.encoding_method == "frequency":
-                n_obs = float(len(X))
-                self.encoder_dict_[var] = (X[var].value_counts() / n_obs).to_dict()
+                for var in self.variables_:
+                    n_obs = float(len(X))
+                    self.encoder_dict_[var] = (X[var].value_counts() / n_obs).to_dict()
 
         self._check_encoding_dictionary()
 
         return self
+
+    def transform(self, X: pd.DataFrame) -> pd.DataFrame:
+        """
+        Replace the categories with counts/frequencies.Replace infrequent categories by
+        the count/frequencies of the 'backoff_bin' based on the threshold provided.
+        In case, no threshold was provided,categories are replaced with
+        counts/frequencies according to encoding_method.
+
+        Parameters
+        ----------
+        X: pandas dataframe of shape = [n_samples, n_features]
+            The input samples.
+
+        Returns
+        -------
+        X: pandas dataframe of shape = [n_samples, n_features]
+            The dataframe where categories are encoded.
+        """
+        X = self._check_transform_input_and_state(X)
+
+        if self.threshold is None:
+            return super().transform(X)
+        else:
+            for feature in self.variables_:
+                keys_before_transform = set(list(X[feature].value_counts().keys()))
+                keys_from_encoder_dict = set(self.encoder_dict_[feature].keys())
+                reminder_keys = keys_before_transform - keys_from_encoder_dict
+                X[feature] = (
+                    X[feature]
+                    .replace(reminder_keys, "backoff_bin")
+                    .map(self.encoder_dict_[feature])
+                )
+            return X
--- a/feature_engine/encoding/count_frequency.py
+++ b/feature_engine/encoding/count_frequency.py
@@ -148,6 +148,9 @@ class CountFrequencyEncoder(CategoricalInitExpandedMixin, CategoricalMethodsMixi
 
         y: pandas Series, default = None
             y is not needed in this encoder. You can pass y or None.
+
+        threshold : list of floats
+            Thresholds for each column needed for binning rare categories
         """
         X = check_X(X)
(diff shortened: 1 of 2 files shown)


feature_engine/encoding/count_frequency.py
```diff
@@ -131,7 +131,12 @@ def __init__(
 
         self.encoding_method = encoding_method
 
-    def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
+    def fit(
+        self,
+        X: pd.DataFrame,
+        y: Optional[pd.Series] = None,
+        threshold: Optional[List[Optional[float]]] = None,
```
solegalli, line 138: we can't add parameters to the fit method. By convention, fit takes only X and y.

threshold should be in the init.

feature_engine/encoding/count_frequency.py
```diff
@@ -143,16 +148,44 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
...
 
         self.encoder_dict_ = {}
 
+        if not isinstance(threshold, list):
+            threshold = [None] * len(self.variables_)
+
         # learn encoding maps
-        for var in self.variables_:
-            if self.encoding_method == "count":
+        for var, threshold in zip(self.variables_, threshold):
```
solegalli, line 165: I would let the user pass 1 threshold, at this would be applied to all variables. If they want different thresholds for different variable groups, then they can use 2 different transformers.

Otherwise, the class becomes too complicated.

feature_engine/encoding/count_frequency.py
```diff
@@ -143,16 +148,44 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
 
         y: pandas Series, default = None
             y is not needed in this encoder. You can pass y or None.
+
+        threshold : list of floats
```
solegalli, line 152: I would make threshold take 1 value, either int if encoding method is count, or a float if encoding method is frequency. And we would apply the same threshold to all variables.

If the user wants different thresholds for different variable groups, then they can instantiate the transformer twice or more times with the different thresholds and variable groups.

Otherwise the class becomes too complicated.

This docstring should be moved to the init. And in the init, we need to add a catch, if int and count ok, otherwise error, if float and frequency ok, otherwise error. 


the default parameter of threshold is None, which means no categories would be grouped.

solegalli (commented):
Hi @SangamSwadiK 

Sorry for the belated response and thank you so much for kickstarting this PR.

I would simplify the code a bit by allowing the user to pass only 1 threshold value that would be applied to all variables. More details in my comments below.

WIth that, and moving the threshold to the init parameter (this is scikit-learn convention), I think we are on the right track, and you could add the tests.

Thanks a lot and apologies again for the delay in the review!

feature_engine/encoding/count_frequency.py
SangamSwadiK: Got it, Ill do the changes and resubmit.

david-cortes:
Isn't this what `RareLabelEncoder` is for? If I understand it correctly, this would have the same effect as putting both of them in scikit-learn pipeline - or am I missing some detail?

solegalli:
> Isn't this what `RareLabelEncoder` is for? If I understand it correctly, this would have the same effect as putting both of them in scikit-learn pipeline - or am I missing some detail?

yep, it is.

SangamSwadiK:
@solegalli I've implemented the changes, I've added transformer and inverse transformer(no functionality) and tests.
 Please let me know for any changes.

Thanks.

feature_engine/encoding/count_frequency.py
```diff
@@ -78,6 +78,10 @@ class CountFrequencyEncoder(CategoricalInitExpandedMixin, CategoricalMethodsMixi
 
     {errors}
 
+    threshold: int,float or None, default=None
+        With threshold you have the option to perform backoff binning for all
```
solegalli, line 82: backoff binning is how the author calls this in the book, right? 

I don't think people will understand this. We need to be more specific of what threshold is doing, for the lay audience. 

Something like: if a category is present in less than threshold number / fraction of observations, it will be grouped with other labels, and then replaced by the number/fraction of observations seen in all this rare labels combined. If encoding_method is 'count' then threshold takes an integer. If encoding_method is frequency, then threshold takes a float. If None, no grouping will be performed

I would also add a See Also section in the docstrings pointing to the rarelabelencoder for reference:

```
See Also
---------
feature_engine.encoding.RareLabelEncoder
```

feature_engine/encoding/count_frequency.py
```diff
@@ -143,22 +175,87 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
...
+                    n_obs = float(len(X))
+                    self.encoder_dict_[var] = (X[var].value_counts() / n_obs).to_dict()
 
         self._check_encoding_dictionary()
 
         return self
+
+    def transform(self, X: pd.DataFrame) -> pd.DataFrame:
+        """
+        Replace the categories with counts/frequencies.Replace infrequent categories by
```
solegalli, line 228: I would not modify this docstring. Can we revert to the original? all main info goes in the init docstrings at the top of the class.

More conventions :/

If you think it is important, then move to the top. we can also add an example in the user_guide: https://github.com/feature-engine/feature_engine/blob/main/docs/user_guide/encoding/CountFrequencyEncoder.rst

feature_engine/encoding/count_frequency.py
```diff
@@ -143,22 +175,87 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
...
+        Returns
+        -------
+        X: pandas dataframe of shape = [n_samples, n_features]
+            The dataframe where categories are encoded.
+        """
+        X = self._check_transform_input_and_state(X)
+
+        if self.threshold is None:
+            return super().transform(X)
+        else:
```
solegalli, line 247: Ideally, we should not modify transform

feature_engine/encoding/count_frequency.py
```diff
@@ -143,22 +175,87 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
...
+                keys_from_encoder_dict = set(self.encoder_dict_[feature].keys())
+                reminder_keys = keys_before_transform - keys_from_encoder_dict
+                X[feature] = (
+                    X[feature]
+                    .replace(reminder_keys, "backoff_bin")
+                    .map(self.encoder_dict_[feature])
+                )
+            return X
+
+    def inverse_transform(self, X: pd.DataFrame):
```
solegalli, line 259: why not?

my idea of the implementation is the following:

category A is present in 10 observations, B in 20 and C in 2 and D in 1.

- if threshold is none, then the encoding_dict = {A:10, B:20, C:2, D:1}
- if threshold is 5, then encoding_dict = {A:10, B:20, C:3, D:3}

Is this what you had in mind? Or what was your idea?

solegalli (commented):
Hi @SangamSwadiK 

Thanks a lot for the code changes.

I have a couple of suggestions and then down in the inverse_transform method, I added a comment to try to understand if we both have the same idea of what this changes should achieve.

Would you be able to have a look?

Thank you

feature_engine/encoding/count_frequency.py
```diff
@@ -143,22 +175,87 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
...
+                for k, v in (X[var].value_counts() / n_obs).to_dict().items():
+                    if v <= self.threshold:
+                        temp["backoff_bin"] = temp["backoff_bin"] + v
+                    else:
+                        temp[k] = v
+                self.encoder_dict_[var] = temp
+
+        else:
+
             if self.encoding_method == "count":
```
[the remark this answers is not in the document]
SangamSwadiK: Got it. I'll  try to come up with a better way.

feature_engine/encoding/count_frequency.py
SangamSwadiK: Hi,
My idea was, if threshold was None, then return super.transform() as it is as no category swapping is needed, base_encoder's transform works fine.

If some threshold was given, then we have to swap the categories whose count/frequency < threshold with "backoff_bin" category. Since that swapping was missing in the base_encoder transform and introducing NaN's in the transformed DataFrame. I did it this way.

Initially I thought of using base_encoder's transform and replace all the NaN's with backoff bin values, but that would give a warning. Hence did it this way.

feature_engine/encoding/count_frequency.py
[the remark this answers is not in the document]
SangamSwadiK: @solegalli  
Hi, this is what I came up with using comprehensions. I've provided the code below.

```
import copy 


enc_dict  = {
    'var_A': {'A':25 ,'B': 3,'C': 4},
    'var_B': {'Q': 9,'R': 21,'Q': 8}
            }

threshold = 5

enc_dct_cpy = copy.deepcopy(enc_dict.copy())
for k, v in enc_dct_cpy.items():
    enc_dict[k].update({"backoff_bin":sum([enc_dict[k].pop(key) for key in [key for key,value in v.items() if value <= threshold ]])})
    
print(f"Original dict: {enc_dct_cpy}") 
print()
print(f"Updated dict based on threshold : {enc_dict}")
```

**When formatted with **Black** the **fit()** looks like this** 
```
if self.encoding_method == "count":
            for var in self.variables_:
                self.encoder_dict_[var] = X[var].value_counts().to_dict()

        elif self.encoding_method == "frequency":
            for var in self.variables_:
                n_obs = float(len(X))
                self.encoder_dict_[var] = (X[var].value_counts() / n_obs).to_dict()

        if self.threshold is not None:
            enc_dct_cpy = copy.deepcopy(self.encoder_dict_.copy())
            for k, v in enc_dct_cpy.items():
                self.encoder_dict_[k].update(
                    {
                        "backoff_bin": sum(
                            [
                                self.encoder_dict_[k].pop(key)
                                for key in [
                                    key
                                    for key, value in v.items()
                                    if value <= self.threshold
                                ]
                            ]
                        )
                    }
                )

```
The above is a dict().update() and a nested list comprehension within the **dict update which removes keys that have values less then threshold and simultaneously updates the dict with "backoff bin"**

Will this be okay ?

feature_engine/encoding/count_frequency.py
```diff
@@ -121,17 +125,45 @@ def __init__(
...
+                )
+            )
+
+        super().__init__(variables, ignore_format, errors)
         self.encoding_method = encoding_method
+        self.threshold = threshold
 
-    def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
+    def fit(
+        self,
```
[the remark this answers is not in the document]
SangamSwadiK: > 

Hi, preserving all in one causes linting issues from flake8, hence I reformatted it to pass Flake8

Should I do them in the original layout format ?
```
count_frequency.py:137:89: E501 line too long (102 > 88 characters)
count_frequency.py:144:89: E501 line too long (108 > 88 characters)
```

feature_engine/encoding/count_frequency.py
SangamSwadiK: > 
Yes, adding about threshold again in transform would be repetitive since its already present in init, I'll revert back to original.

SangamSwadiK:
Hi @solegalli , please find the checklist of suggestions implemented.
- [x] Reverted back transform docstring to base_encoder's transform docstring.
- [x] Remove threshold doc from fit().
- [x] Redo fit() with original logic as per suggestion, reduced duplication of code.
- [x] Redo transform(). 
- [x] Preserve original layout for fit(), but not **raising value errors for threshold and combination of encoding method** .
- [x] Add alternative  explanation for backoff binning to make it easily understandable.

solegalli:
Hi @SangamSwadiK 

@david-cortes is right that the logic we are trying to implement here is already done by another transformer, the RareLabelEncoder. 

So I am having second thoughts as to whether it is a good idea to make the class more complicated by adding additional logic that is already taken care of elsewhere.

Would you mind if we put this PR on hold and see what the community thinks about this?

Sorry, I know it is frustrating to work on something that does not go through eventually :_(

PR #488, opened 2022-07-20
Information value for feature selection using weight of evidence (WoE)

### LINKED ISSUE #280: add selection transformer based on the Information Value (IV) for nominal variables
Presently, there are no packages in python to calculate Information Value using WOE for nominal/categorical variables. As WOE Encoder is already available in Feature Engine, hence I am raising the request for a feature to obtain information value  for nominal variables

Morgan-Sell:
Information value is a variable-selection method that is used with binary classifiers. Information value summarizes how much knowing `var_A` helps in predicting the dependent variable.

feature-engine includes a WoE Encoder. The InformationValue class will leverage the WoEEncoder class.

Closes #280 

Notes from # 280:

https://www.listendata.com/2015/03/weight-of-evidence-woe-and-information.html

https://www.listendata.com/2019/08/WOE-IV-Continuous-Dependent.html

http://ucanalytics.com/blogs/information-value-and-weight-of-evidencebanking-case/


Morgan-Sell:
hola @solegalli,

I have a couple of question about syntax related to `self` and the use of internal methods:

1. If a method returns a variable that will be assigned to a class attribute should syntax be kept as such? Or, is it better to create the attribute in the method and end the method with `return self`? See the comparison below.

```
# Option A: _method() is created within the class
self.attribute = self._method(data)

# Option B:
def _method(self, data):
   # code
   # code
   
   self.attribute = temp_variable

   return self

```

Which is the preferred syntax? Option A or B?



2. Should `self` always precede a method when the method iscalled within another method? 
    - Imagine we have a method called `_validation_method()` that accepts variable `X`. `_validation_method()` is called in the `fit()` method. 
    - Should the `_validation_method()` always be called using the following syntax `self._validation_method(X)`? Or, can we implement the method using `_validation_method(X)`, i.e., sans `self`?
    - Does using `self` give the method access to the class's attributes?

Morgan-Sell:
hi @solegalli,

I believe this class is now ready for your virtuoso programming skills ;)

I'm confused as to why tests for CountryFrequencyEncoder and OrdnalEncoder are failing. Once you're settled, it would be great to dive into the details of some of these master/meta unit tests.

feature_engine/_docstrings/class_inputs.py
```diff
@@ -13,3 +13,8 @@
         contain missing values. If 'ignore', missing data will be ignored when learning
         parameters or performing the transformation.
         """.rstrip()
+
+_variables_categorical_docstring = """variables: list, default=None
```
solegalli, line 17: this doctring already exists within the encoding folder. I would remove this and import the existing docstring.

feature_engine/selection/information_value.py
```diff
@@ -0,0 +1,328 @@
...
+
+    IV will allow you to assess each variable's independent contribution to
+    the target variable and rank the variables in terms of their univariate
+    predictive strength.
+
+    Parameters
+    ----------
+    {variables}
+
+    sort_values: bool, default=False
```
solegalli, line 66: do we need this?

feature_engine/selection/information_value.py
```diff
@@ -0,0 +1,328 @@
...
+
+    Attributes
+    ----------
+    {variables_}
+
+    {feature_names_in}
+
+    {n_features_in}
+
+    class_diff_encoder_dict_:
```
solegalli, line 82: we do not need to expose this

feature_engine/selection/information_value.py
```diff
@@ -0,0 +1,328 @@
...
+         The difference between the binomial distributions of positive and
+         negative classes for each unique value of the selected categorical
+         variables.
+
+    woe_encoder_dict_:
```
solegalli, line 87: I would probably not expose this

feature_engine/selection/information_value.py
```diff
@@ -0,0 +1,328 @@
...
+        Contains the weights of evidence (WoE) for all the unique values of
+        the selected categorical variables.
+
+    information_values_:
```
solegalli, line 91: I think this is the only dictionary that we want to show the user.

feature_engine/selection/information_value.py
```diff
@@ -0,0 +1,328 @@
...
+        X: pandas dataframe of shape = [n_samples, n_features]
+            The training input samples.
+            Can be the entire dataframe, not just the categorical variables.
+
+        y: pandas series of shape = [n_samples, ]
+            Target, must be binary.
+
+        """
+        # check input dataframe
+        X, y = check_X_y(X, y)
```
solegalli, line 147: The class WoEEncoder and this class, both need to calculate the WoE.

So we need to extract that calculation into a function, that both WoE and this class can use to return pos and neg. This way, we do not repeat the code /logic

feature_engine/selection/information_value.py
```diff
@@ -0,0 +1,328 @@
...
+            X, y
+        )
+
+        # get WoE values for unique values of selected categorical variables
+        self.woe_encoder_dict_ = self._calc_woe_encoder_dict(X, y)
+
+        # get information values for unique values of selected categorical variables
+        self.information_values_ = self._calc_information_values()
+
+        self._get_feature_names_in(X)
```
solegalli, line 170: as part of fit() we also need to identify the features that will be dropped in an attribute called features_to_drop_

see for example the class SelectByShuffling.

feature_engine/selection/information_value.py
```diff
@@ -0,0 +1,328 @@
...
+
+        return self
+
+    def transform(self, X: pd.DataFrame) -> pd.DataFrame:
```
solegalli, line 174: this method should be inherited from the base selection class, that knows that it should drop the features in the attribute features to drop. If you have a look at any of the selector classes (except DropFeatures), you'll understand what I mean.

solegalli (commented):
Hi @Morgan-Sell 

Thank you for the amount of code committed!

I think we need to create an abstraction (function) that can be used both by WoEEncoder and this class to calculate the WoE, and then expand to the IV.

This new function should live in the encoding module.

This means that you would need to modify slightly the WoE encoder to use this function instead of performing the calculations there.

Also, I simplified the logic of the WoE in this PR: #500 so maybe wait until that is merged before moving on with this changes, hopefully in the next few days.

Thanks a lot!

solegalli:
Hi @Morgan-Sell 

PYI: https://github.com/Morgan-Sell/feature_engine/pull/12

Morgan-Sell:
Amazing, @solegalli! How do you see these improvements?

I'll work on the unit tests.

One question, how is `self.performance_drifts_` derived? It's not from any of the 3 parent classes. Is it from a grandparent class?

solegalli:
I don't think we need performance_drifts_ here. Do we?

solegalli:
I think this is still pending.

ToDo:

- [ ] remove this [file](feature_engine/selection/drop_psi_features.py) from pr
- [ ] tests
- [ ] docs: index, api, user_guide

Morgan-Sell:
Hi @solegalli,

I'll work on the steps outlined above.

Questions/responses to the above comments:

1. I don't think this class requires performance_drifts_ as this attribute is used with shuffled features that's why I was a bit confused. If so, do we need to set a default value for `threshold`?

2. Is the drop_psi_features.py the file that needs to be removed from the PR?

solegalli:
1) yes, if i understand correctly we select features with iv beyond a threshold
2) yes.

solegalli:
why are we deleting PSI selector?

Morgan-Sell:
hola @solegalli, I've returned from Black Rock City ;)

> why are we deleting PSI selector?

Two messages prior, I asked, "Is the drop_psi_features.py the file that needs to be removed from the PR?". You responded "Yes"; therefore, I deleted the file.

I'm guessing that I misunderstood something. What would you like to be done?

Morgan-Sell:
hola @solegalli,

Are there any additional unit tests that we should create for this class?

Morgan-Sell:
Hi @solegalli,

Most material that explains information value (IV) has a table that states if IV is greater than 0.5 then the variable's predictive power is "suspicious, too good to be true".

When a variable has an IV score > 0.5, does a data scientist drop that variable? If so, should this transformer **two** threshold init params - a floor and ceiling?

solegalli:
> hola @solegalli, I've returned from Black Rock City ;)
> 
> > why are we deleting PSI selector?
> 
> Two messages prior, I asked, "Is the drop_psi_features.py the file that needs to be removed from the PR?". You responded "Yes"; therefore, I deleted the file.
> 
> I'm guessing that I misunderstood something. What would you like to be done?
> 
> I have another question regarding the unit tests. I'm predominantly using **df_enc** for the unit tests. Given that it's a made-up dataframe, each variable's information value is significantly low. Therefore, I use unrealistic thresholds, e.g. -1, to test the transformer's functionality.
> 
> Given that the unit tests assess the transformer's algorithmic functionality, is it OK for these transformers' init params to be impracticable?

I see. By "this file should be removed from the PR" I meant, we should not include changes to this file in this PR. Because here we are working on a completely different class. The way to "remove" a file from a PR is to checkout the version of the file in main. If you delete it, it will be deleted from main when we merge.

Regarding testing, the tests aim to test the logic. And the logic needs to make "sense". Now, we may be ok to add "impracticable" thresholds, but if later on we choose to restrict threshold parameters to what "makes sense", we would have to re-do all the tests. All of these is me thinking out loud, sorry. So I guess, it would make more sense to have tests that test not only the logic but also how how the transformer would be used in real life.

solegalli:
> e read that IV "loo

Hi @Morgan-Sell 

Q1) I think one threshold is enough. If IV > 0.5 it is for the user to know what to do. My view at least
Q2) Yes, this method is intended for linear models.

Thank you!

docs/index.rst
```diff
@@ -178,6 +178,7 @@ Feature Selection:
 - :doc:`api_doc/selection/SelectByTargetMeanPerformance`: selects features based on target mean encoding performance
 - :doc:`api_doc/selection/RecursiveFeatureElimination`: selects features recursively, by evaluating model performance
 - :doc:`api_doc/selection/RecursiveFeatureAddition`: selects features recursively, by evaluating model performance
+- :doc:`api_doc/selection/SelectByInformationValue`: selects features that have an information value greater than a set threshold
```
solegalli, line 181: I would rephrase:

"selects features based on the information value"

docs/user_guide/selection/SelectByInformationValue.rst
```diff
@@ -0,0 +1,153 @@
...
++----------------------+---------------------------------+
+| > 0.5                | Suspicious, too good to be true |
++----------------------+---------------------------------+
+
+
+Important
+---------
+:class:`SelectByInformationValue()` automatically identifies categorical variables, i.e., variable types that
+are object or categorical. If any dataset's categorical variables uses numeric values for its categories or bins
+the init parameter :code:`ignore_format` should be set to :code:`False`.
```
solegalli, line 64: True instead of false

feature_engine/selection/__init__.py
```diff
@@ -5,7 +5,7 @@
 from .drop_correlated_features import DropCorrelatedFeatures
 from .drop_duplicate_features import DropDuplicateFeatures
 from .drop_features import DropFeatures
-from .drop_psi_features import DropHighPSIFeatures
+from .information_value import SelectByInformationValue
```
solegalli, line 8: I probable was not clear when I asked to remove this file. We need to checkout the version in master, so this file completely disappears from this PR.

tests/test_selection/test_information_value.py
```diff
@@ -0,0 +1,74 @@
...
+    with pytest.raises(ValueError):
+        SelectByInformationValue(
+            variables=None,
+            threshold=_threshold,
+            ignore_format=False,
+            confirm_variables=False,
+        )
+
+
+def test_error_when_more_than_two_classes(df_enc_numeric):
```
solegalli, line 19: what is this testing?

tests/test_selection/test_information_value.py
```diff
@@ -0,0 +1,74 @@
...
+        ignore_format=False,
+        confirm_variables=False
+    )
+    with pytest.raises(ValueError):
+        transformer.fit(
+            df_enc_numeric[["var_A", "target"]], df_enc_numeric["var_B"]
+        )
+
+
+def test_error_when_no_categorical_values(df_test):
```
solegalli, line 32: values? variables?

tests/test_selection/test_information_value.py
```diff
@@ -0,0 +1,74 @@
...
+    X, y = df_test
+
+    transformer = SelectByInformationValue(
+        variables=None,
+        threshold=0.2,
+        ignore_format=False,
+        confirm_variables=False
+    )
+    with pytest.raises(ValueError):
+        transformer.fit_transform(X, y)
```
solegalli, line 42: what should raise the error? fit? or transform? if transform, then I would take the fit out of the pytest raises. If fit, then I would not use transform. If both, then we need to expand this test for both methods individually.

solegalli (commented):
Hi @Morgan-Sell 

This is looking really good.

We need to add tests for the method _calculate_iv. 

And also test that the information_values returns the expected dictionary.

And finally add the class to the test check estimators, were we test all the classes together for the common tests, like selecting variables features_names_in_ etc. These would be done automatically, when the class is added to the list.

Also, if the user guide is "borrowing" material from some blog /article, we need to credit them.

Finally, it's been a while, so I forgot, this method is intended only for categorical variables? what do the blogs say? 

Because I am thinking that we could discretize numerical first, then add woe, to calculate the iv, and select from all features, and not just categorical.

Thank you very much!

Morgan-Sell:
Thank you for the thorough feedback, @solegalli!

As always, I have follow-up questions/comments.

**Q1:**
> Q1) I think one threshold is enough. If IV > 0.5 it is for the user to know what to do. My view at least 

This makes sense. In practice, however, the user will always have to execute `sel.information_values_` to check the IV. This doesn't seem seamless. How would the user implement the transformer when using a scikit-learn Pipeline? How would the transformer be used in a deployed model?

**Q2:**
> I see. By "this file should be removed from the PR" I meant, we should not include changes to this file in this PR. Because here we are working on a completely different class. The way to "remove" a file from a PR is to checkout the version of the file in main. If you delete it, it will be deleted from main when we merge.

Ok got it. I thought it was strange to delete the class. I apologize. I should've asked before doing so. What's the proper approach in recreating the PSIDropFeatures class? I'm guessing there's something more elegant than creating a new .py file and copying/pasting the existing code.


**Q3:**
> Also, if the user guide is "borrowing" material from some blog /article, we need to credit them.

I was going to ask which citation methodology we use. Where can I find an example?

**Q4:**
> Finally, it's been a while, so I forgot, this method is intended only for categorical variables? what do the blogs say?
> 
> Because I am thinking that we could discretize numerical first, then add woe, to calculate the iv, and select from all features, and not just categorical.

Good thought! In the user guide, I say categorical variables have "categories or bins". I view a discretized variable as a categorical variable where the bins are categories. Is this the generally-accept view or la vista del boludo?

Are you suggesting that we say, "Numerical variables can be used but they must first be discretized. And, if the bins are assigned integers as labels, then `ignore_format` must be set to `False`."?

tests/test_selection/test_information_value.py
Morgan-Sell: I understand the IV can only be used with binary classification. The test checks whether an error is raised when the dependent variable is multiclass.

tests/test_selection/test_information_value.py
Morgan-Sell: Yes, I change it to `test_error_when_no_categorical_variables()`.

tests/test_selection/test_information_value.py
Morgan-Sell: `fit()` is when the transformer checks for categorical variables. Therefore, I revised the code for `transformer.fit(X, y)` to be under the `pytest.raises`.

feature_engine/selection/__init__.py
Morgan-Sell: I asked in Q2 in my comment below what is the appropriate way to resolve this issue. If I checkout the commit hash that is one prior to the commit that entailed deleting DropHighPSIFeatures and its dependencies, then I would lose all of the work that I've already completed, right?

As an FYI, the commit has is `7f687d467ab0a83c91d02ce519430992638c39aa`

Sorry for the mistake!

Morgan-Sell:
hi @solegalli,

I mentioned earlier that I'm seeing unrealistic information values. I've tried a few different OpenML datasets and the IVs are all over the place. I think there may be an error in the variables being returned by `self._calculate_woe()` and being passed to `self._calculate_iv()`. Before I make crazy changes, I think it's best for you to check my logic. I added comments to the code base.

I used the table from this [Medium article](https://sundarstyles89.medium.com/weight-of-evidence-and-information-value-using-python-6f05072e83eb) as the foundation of my logic.

Morgan-Sell:
hi @solegalli, I know you had a couple of deadlines. At the same time, I thought I'd check in to see if you had any thoughts to my comments/questions above. It seems like we want to increase feature-engine's feature selection capabilities ;)

solegalli:
This PR is the next one on my list. Thanks for the patience :)

Morgan-Sell:
Hola @solegalli, any thoughts? ;)

solegalli:
Hi @Morgan-Sell 

I am continuing the conversation here: https://github.com/feature-engine/feature_engine/pull/542

Issue #489, opened 2022-07-21
Polynomial Feaatures + SklearnWrapper weird behavior


datacubeR:
**Describe the bug**

When using PolynomialFeaturs + SklearnWrappers the base features are duplicated, when trying to dedup using DropDuplicateFeatures the values are repeated again!!

Using the simple Titanic Dataset you can run something like this:

```python
df = pd.read_csv('https://www.openml.org/data/get_csv/16826755/phpMYEkMl')
X = df[['pclass','sex','age','fare','embarked']]
y = df.survived

X_train, X_val, y_train, y_val = train_test_split(X, y, test_size=0.2, random_state=42)
pipe = Pipeline(steps = [
    ('ci', CategoricalImputer(imputation_method='frequent')),
    ('mmi', MeanMedianImputer(imputation_method='mean')),
    ('od', OrdinalEncoder(encoding_method='arbitrary')),
    ('pl', SklearnTransformerWrapper(PolynomialFeatures(degree = 2, interaction_only = True, include_bias=False), variables=['pclass','sex'])),
    #('drop', DropDuplicateFeatures()),
    #('sc', SklearnTransformerWrapper(StandardScaler(), variables=['Age','Fare'])),
    #('lr', LogisticRegression(random_state=42))

])
pipe.fit_transform(X_train)
```

This returns the first issue:  

![image](https://user-images.githubusercontent.com/48638337/180299902-4f0bf6c2-e4ea-419a-9768-245c45d5df29.png)

I'm getting pclass and sex duplicated, I'm expecting to get back only interactions. This is expected from [Sklearn Docs](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.PolynomialFeatures.html) but why would I want duplicated features?

Looking into Feature Engine Docs I found `DropDuplicateFeatures()`, but if applying into the Pipeline I get this:

```python
df = pd.read_csv('https://www.openml.org/data/get_csv/16826755/phpMYEkMl')
X = df[['pclass','sex','age','fare','embarked']]
y = df.survived

X_train, X_val, y_train, y_val = train_test_split(X, y, test_size=0.2, random_state=42)
pipe = Pipeline(steps = [
    ('ci', CategoricalImputer(imputation_method='frequent')),
    ('mmi', MeanMedianImputer(imputation_method='mean')),
    ('od', OrdinalEncoder(encoding_method='arbitrary')),
[...]

PR #491, opened 2022-08-04
fix Sklearnwrapper output for OneHotEncoder and PolynomialFeatures to avoid duplicated features

datacubeR:
Hi @solegalli, 
This is my shot for fixing #489.
After checking the code in detail I think the issue affects not only `PolynomialFeatures` but also Sklearn `OneHotEncoder`.
When using `SklearnTransformerWrapper` + `OneHotEncoder` I get this:

image

 Which is totally unexpected, since the idea is to replace Categorical Features for OneHotEncoded ones.

After the fix I'm proposing I get this for `OneHotEncoder`
image

Which I think is the expected behavior. 

For `PolynomialFeatures` I get the following:
image

Please note that an `OrdinalEncoder` was applied to Categorical Variables before applying `PolynomialFeatures`, otherwise it throws an error since they are not numerical features.

I think these cases could be added as test cases if you think it's OK. But before I would love to get your feedback.

Best,

Alfonso

### DIFF
--- a/docs/user_guide/wrappers/Wrapper.rst
+++ b/docs/user_guide/wrappers/Wrapper.rst
@@ -116,7 +116,11 @@ to select only a subset of the variables.
     X_train_t = selector.transform(X_train.fillna(0))
     X_test_t = selector.transform(X_test.fillna(0))
 
-Even though Feature Engine has its own implementation of OneHotEncoder, you may want to use Scikit-Learn Implementation in order to access different options, such as drop first Category.
+Even though Feature-engine has its own implementation of OneHotEncoder, you may want 
+to use Scikit-Learn's transformer in order to access different options, 
+such as drop first Category. 
+In the following example, we show you how to apply Scikit-learn's OneHotEncoder to a 
+subset of categories using the :class:SklearnTransformerWrapper().
 
 .. code:: python
 
@@ -126,19 +130,39 @@ Even though Feature Engine has its own implementation of OneHotEncoder, you may 
     from sklearn.preprocessing import OneHotEncoder
 
     df = pd.read_csv('https://www.openml.org/data/get_csv/16826755/phpMYEkMl')
-    X = df[['pclass']]
+    X = df
     y = df.survived
     X_train, X_test, y_train, y_test= train_test_split(X, y, test_size=0.2, random_state=42)
 
-    ohe = SklearnTransformerWrapper(OneHotEncoder(sparse=False, drop='first'))
+    ohe = SklearnTransformerWrapper(OneHotEncoder(sparse=False, drop='first'), variables = ['pclass','sex'])
 
     ohe.fit(X_train)
 
     X_train_transformed = ohe.transform(X_train)
     X_test_transformed = ohe.transform(X_test)
 
+    print(X_train_transformed.head())
+          age   fare     embarked  pclass_2  pclass_3  sex_male
+    772   17   7.8958        S       0.0       1.0       1.0
+    543   36     10.5        S       1.0       0.0       1.0
+    289   18    79.65        S       0.0       0.0       0.0
+    10    47  227.525        C       0.0       0.0       1.0
+    147  NaN     42.4        S       0.0       0.0       1.0
 
-Let's say you want to use :class:`SklearnTransformerWrapper()` in a more complex context. As you may note there are `?` signs to denote unknown values. Due to the complexity of the transformations needed we'll use a Pipeline to impute missing values, encode categorical features and create interactions for specific variables using Scikit-Learn PolynomialFeatures.
+    print(X_test_transformed.head())
+          age   fare      embarked  pclass_2  pclass_3  sex_male
+    1148   35    7.125        S       0.0       1.0       1.0
+    1049   20  15.7417        C       0.0       1.0       1.0
+    982   NaN   7.8958        S       0.0       1.0       1.0
+    808   NaN     8.05        S       0.0       1.0       1.0
+    1195  NaN     7.75        Q       0.0       1.0       1.0
+
+
+Let's say you want to use :class:`SklearnTransformerWrapper()` in a more complex 
+context. As you may note there are `?` signs to denote unknown values. Due to the 
+complexity of the transformations needed we'll use a Pipeline to impute missing values, 
+encode categorical features and create interactions for specific variables using 
+Scikit-Learn's PolynomialFeatures.
 
 .. code:: python
     
@@ -166,6 +190,22 @@ Let's say you want to use :class:`SklearnTransformerWrapper()` in a more complex
     pipeline.fit(X_train)
     X_train_transformed = pipeline.transform(X_train)
     X_test_transformed = pipeline.transform(X_test)
+
+    print(X_train_transformed.head())
+            age        fare      embarked  pclass  sex  pclass sex
+    772  17.000000    7.8958         0     3.0     0.0      0.0
+    543  36.000000   10.5000         0     2.0     0.0      0.0
+    289  18.000000   79.6500         0     1.0     1.0      1.0
+    10   47.000000  227.5250         1     1.0     0.0      0.0
+    147  29.532738   42.4000         0     1.0     0.0      0.0
+
+    print(X_test_transformed.head())
+            age        fare      embarked  pclass  sex  pclass sex
+    1148  35.000000   7.1250         0     3.0     0.0      0.0
+    1049  20.000000  15.7417         1     3.0     0.0      0.0
+    982   29.532738   7.8958         0     3.0     0.0      0.0
+    808   29.532738   8.0500         0     3.0     0.0      0.0
+    1195  29.532738   7.7500         2     3.0     0.0      0.0
 
 More details
 ^^^^^^^^^^^^
--- a/docs/user_guide/wrappers/Wrapper.rst
+++ b/docs/user_guide/wrappers/Wrapper.rst
@@ -116,6 +116,56 @@ to select only a subset of the variables.
     X_train_t = selector.transform(X_train.fillna(0))
     X_test_t = selector.transform(X_test.fillna(0))
 
+Even though Feature Engine has its own implementation of OneHotEncoder, you may want to use Scikit-Learn Implementation in order to access different options, such as drop first Category.
+
+.. code:: python
+
+    import pandas as pd
+    import numpy as np
+    from sklearn.model_selection import train_test_split
+    from sklearn.preprocessing import OneHotEncoder
+
+    df = pd.read_csv('https://www.openml.org/data/get_csv/16826755/phpMYEkMl')
+    X = df[['pclass']]
+    y = df.survived
+    X_train, X_test, y_train, y_test= train_test_split(X, y, test_size=0.2, random_state=42)
+
+    ohe = SklearnTransformerWrapper(OneHotEncoder(sparse=False, drop='first'))
+
+    ohe.fit(X_train)
+
+    X_train_transformed = ohe.transform(X_train)
+    X_test_transformed = ohe.transform(X_test)
+
+
+Let's say you want to use :class:`SklearnTransformerWrapper()` in a more complex context. As you may note there are `?` signs to denote unknown values. Due to the complexity of the transformations needed we'll use a Pipeline to impute missing values, encode categorical features and create interactions for specific variables using Scikit-Learn PolynomialFeatures.
+
+.. code:: python
+    
+    import pandas as pd
+    import numpy as np
+    from sklearn.model_selection import train_test_split
+    from sklearn.preprocessing import PolynomialFeatures
+    from sklearn.pipeline import Pipeline
+    from feature_engine.imputation import CategoricalImputer, MeanMedianImputer
+    from feature_engine.encoding import OrdinalEncoder
+    from feature_engine.wrappers import SklearnTransformerWrapper
+
+    df = pd.read_csv('https://www.openml.org/data/get_csv/16826755/phpMYEkMl')
+    X = df[['pclass','sex','age','fare','embarked']].replace('?',np.nan)
+    X[['age', 'fare']] = X[['age', 'fare']].astype('float64')
+    y = df.survived
+
+    X_train, X_test, y_train, y_test= train_test_split(X, y, test_size=0.2, random_state=42)
+    pipeline = Pipeline(steps = [
+        ('ci', CategoricalImputer(imputation_method='frequent')),
+        ('mmi', MeanMedianImputer(imputation_method='mean')),
+        ('od', OrdinalEncoder(encoding_method='arbitrary')),
+        ('pl', SklearnTransformerWrapper(PolynomialFeatures(interaction_only = True, include_bias=False), variables=['pclass','sex']))
+    ])
+    pipeline.fit(X_train)
+    X_train_transformed = pipeline.transform(X_train)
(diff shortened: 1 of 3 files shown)


solegalli:
Hi @datacubeR 

Thanks a lot for the changes. 

I do agree that the idea is to replace the original categorical variables by the one hot encoded ones. So I guess, it is OK to have them removed from the dataset as you did.

Originally I thought, if that is the intended functionality, then why not use the OHE from feature engine instead of the one from sklearn? that is why I decided not to drop them. But I think it makes sense to go ahead with your suggestion.

I think it would be great to have those code snippets as tests, and also maybe in the user guide: https://feature-engine.readthedocs.io/en/latest/user_guide/wrappers/Wrapper.html#sklearn-wrapper (I mean examples of wrapping the OHE and the PolynomialFeatures)

would you be able to do that?

thanks a lot!

solegalli:
PS: I fixed the failing tests in the main branch last week. It is something related to the boxcox with the latest version of scipy. If you sync your main branch and then rebase it to this feature branch, it should get those sorted @datacubeR 

thank you!

datacubeR:
@solegalli I think my fork is up to date, but the failing tests is because they considering adding Features rather than replacing the existing feature for the transformed ones. I will fix those and I'll get back to you.

datacubeR:
@solegalli do I need to create a separate PR for the examples in the User Guide or I just include those in this one?

solegalli:
> parate PR for the exampl

Here would be good :) 

Thank you!

tests/test_wrappers/test_sklearn_wrapper.py
```diff
@@ -263,6 +264,7 @@ def test_sklearn_imputer_allfeatures_with_constant(df_na):
 # One Hot Encoder
 def test_sklearn_ohe_object_one_feature(df_vartypes):
     variables_to_encode = ["Name"]
+    print("This:\n", df_vartypes)
```
solegalli, line 267: is this print statement meant to be here?

solegalli:
It's looking good.

I think we need to add a test to corroborate that the method get_feature_names_out returns the original features, without the ones used in the transformation, plus the new ones, when it is called without passing the input_feature parameters.

so `tr_wrap.get_feature_names_out()` should return = feature_names_in - self.variables_ + new features

Could you add a test for that please? Or is it already there? I mean just for the Poly and OHE

tests/test_wrappers/test_sklearn_wrapper.py
datacubeR: Nop, I wasn't able to find where df_vartypes was defined. I will get rid of that.

solegalli: https://github.com/feature-engine/feature_engine/blob/65fedf100bc85f364b26838866a264dbf7320841/tests/conftest.py#L7-L18

tests/test_wrappers/test_sklearn_wrapper.py
```diff
@@ -176,17 +175,32 @@ def test_wrap_polynomial_features():
...
+    X = fetch_california_housing(as_frame=True).frame
+
+    tr = PolynomialFeatures()
+    tr_wrap = SklearnTransformerWrapper(transformer=PolynomialFeatures())
+    varlist = ["MedInc", "HouseAge", "AveRooms", "AveBedrms"]
+
+    tr.fit(X[varlist])
+    tr_wrap.fit(X[varlist])
+
+    assert (tr.get_feature_names_out() == tr_wrap.get_feature_names_out()).all()
```
datacubeR, line 203: This is the test for Polynomial Features

tests/test_wrappers/test_sklearn_wrapper.py
```diff
@@ -414,6 +418,15 @@ def test_sklearn_ohe_with_crossvalidation():
     assert not any([np.isnan(i) for i in results])
 
 
+def test_wrap_one_hot_encoder_get_features_name_out(df_vartypes):
+    ohe = OneHotEncoder()
+    ohe_wrap = SklearnTransformerWrapper(transformer=OneHotEncoder(sparse=False))
+    ohe.fit(df_vartypes)
+    ohe_wrap.fit(df_vartypes)
+
+    assert (ohe.get_feature_names_out() == ohe_wrap.get_feature_names_out()).all()
```
datacubeR, line 427: This is the test for OneHotEncoder

tests/test_wrappers/test_sklearn_wrapper.py
```diff
@@ -571,7 +584,7 @@ def test_function_transformer_works_with_numericals():
     X_expected = pd.DataFrame({"col1": [2, 3, 4], "col2": ["a", "b", "c"]})
 
     transformer = SklearnTransformerWrapper(
-        FunctionTransformer(lambda x: x+1), variables=["col1"]
+        FunctionTransformer(lambda x: x + 1), variables=["col1"]
```
datacubeR, line 587: I think this is a change caused by running Black

datacubeR (commented):
Would you please review the wording of the new examples added?

tests/test_wrappers/test_sklearn_wrapper.py
solegalli: the test that is missing is something like this:

```
def test_wrap_polynomial_features_get_features_name_out():
    X = fetch_california_housing(as_frame=True).frame

    varlist = ["MedInc", "HouseAge", "AveRooms", "AveBedrms"]
    tr_wrap = SklearnTransformerWrapper(transformer=PolynomialFeatures(), variables=varlist)
   
    tr_wrap.fit(X)

    assert tr_wrap.get_feature_names_out() == The expected list (all features in the original + new features)
    assert tr_wrap.get_feature_names_out(varlist) == All the new polynomial features
    assert tr_wrap.get_feature_names_out(["Medinc") == the features expected from Medinc
```

The thing is, given that the logic in get_feature_names_out was changed slightly, I would like to make sure that it still returns what it is expected to return given the input to input_features

Would you be able to change it?

tests/test_wrappers/test_sklearn_wrapper.py
solegalli: Same as per previous comment, the idea is to just test the functionality of the method, and not in comparison to that of sklearn's.

docs/user_guide/wrappers/Wrapper.rst
```diff
@@ -116,6 +116,56 @@ to select only a subset of the variables.
     X_train_t = selector.transform(X_train.fillna(0))
     X_test_t = selector.transform(X_test.fillna(0))
 
+Even though Feature Engine has its own implementation of OneHotEncoder, you may want to use Scikit-Learn Implementation in order to access different options, such as drop first Category.
```
solegalli, line 119: Could you replace this sentence by the below:

Even though Feature-engine has its own implementation of OneHotEncoder, you may want to use Scikit-Learn'stransformer in order to access different options, such as drop first Category. In the following example, we show you how to apply Scikit-learn's OneHotEncoder to a subset of categories using the SklearnTransformerWrapper.

docs/user_guide/wrappers/Wrapper.rst
```diff
@@ -116,6 +116,56 @@ to select only a subset of the variables.
...
+    import numpy as np
+    from sklearn.model_selection import train_test_split
+    from sklearn.preprocessing import OneHotEncoder
+
+    df = pd.read_csv('https://www.openml.org/data/get_csv/16826755/phpMYEkMl')
+    X = df[['pclass']]
+    y = df.survived
+    X_train, X_test, y_train, y_test= train_test_split(X, y, test_size=0.2, random_state=42)
+
+    ohe = SklearnTransformerWrapper(OneHotEncoder(sparse=False, drop='first'))
```
solegalli, line 133: The point of the sklearntransformerwrapper is to apply sklearn transformer only to a subset of variables. In this case, the df has only one variable, so you could just apply the OHE directly.

I think a more relevant example would be to use the entire titanic data, and apply the ohe to just a subset for example pclass and sex.

docs/user_guide/wrappers/Wrapper.rst
```diff
@@ -116,6 +116,56 @@ to select only a subset of the variables.
...
+
+    ohe.fit(X_train)
+
+    X_train_transformed = ohe.transform(X_train)
+    X_test_transformed = ohe.transform(X_test)
+
```
solegalli, line 139: It would be good to add an image or a printout of the result, for example execute X_train_transform.head() and after that show an image of the final df, or instead print(X_train_transform.head()) and then copy the code output in a code block.

docs/user_guide/wrappers/Wrapper.rst
```diff
@@ -116,6 +116,56 @@ to select only a subset of the variables.
...
+
+Let's say you want to use :class:`SklearnTransformerWrapper()` in a more complex context. As you may note there are `?` signs to denote unknown values. Due to the complexity of the transformations needed we'll use a Pipeline to impute missing values, encode categorical features and create interactions for specific variables using Scikit-Learn PolynomialFeatures.
```
solegalli, line 141: It would be great if you could shorten the length of the sentence to 88 char (that is, break the sentence over 2 lines). 

It reads well, the last 2 words should be: Scikit-Learn's PolynomialFeatures().

docs/user_guide/wrappers/Wrapper.rst
```diff
@@ -116,6 +116,56 @@ to select only a subset of the variables.
...
+    X_train, X_test, y_train, y_test= train_test_split(X, y, test_size=0.2, random_state=42)
+    pipeline = Pipeline(steps = [
+        ('ci', CategoricalImputer(imputation_method='frequent')),
+        ('mmi', MeanMedianImputer(imputation_method='mean')),
+        ('od', OrdinalEncoder(encoding_method='arbitrary')),
+        ('pl', SklearnTransformerWrapper(PolynomialFeatures(interaction_only = True, include_bias=False), variables=['pclass','sex']))
+    ])
+    pipeline.fit(X_train)
+    X_train_transformed = pipeline.transform(X_train)
+    X_test_transformed = pipeline.transform(X_test)
```
solegalli, line 168: great example, thank you!!

Could you add a 5 row display of the resulting dataframe? like in my previous comment.

solegalli:
Hey @datacubeR 

This is almost ready. Some small changes here and there. Would you be able to have a look?

Thank you!

tests/test_wrappers/test_sklearn_wrapper.py
datacubeR: Sure. I think the assertion I have covers your first assertion. I can add the remaining two. Or do you want me to hard code the original features + new features in the first one?

docs/user_guide/wrappers/Wrapper.rst
datacubeR: Ok!!

docs/user_guide/wrappers/Wrapper.rst
datacubeR: Sure

tests/test_wrappers/test_sklearn_wrapper.py
datacubeR: @solegalli, I was checking this one, and just to make sure I didn't touch `get_feature_names_out`. The thing is I'm not sure what is the expected behavior for this one: `assert tr_wrap.get_feature_names_out(["Medinc") == the features expected from Medinc`.
This one should work even if it was trained on a different pool of features? And the expected output is: Medinc, Medinc^2 and all the interactions that includes Medinc?

solegalli: Assuming that the transformer was trained as in my example:

```
    varlist = ["MedInc", "HouseAge", "AveRooms", "AveBedrms"]
    tr_wrap = SklearnTransformerWrapper(transformer=PolynomialFeatures(), variables=varlist)
```

Then the output of tr_wrap.get_feature_names_out(["Medinc") would be as you say the polynomial combinations that involve MedInc. And this, should come out of the box from the Polynomial Features.

If the transformer was trained like this instead:

```
    varlist = ["HouseAge", "AveRooms", "AveBedrms"]
    tr_wrap = SklearnTransformerWrapper(transformer=PolynomialFeatures(), variables=varlist)
```

where MedInc was not part of variables, then the outcome of tr_wrap.get_feature_names_out(["Medinc") should be an error, and I think this should be handled by our class method get_feature_names_out.

datacubeR: @solegalli, This issue was a bit more complicated than I thought. Actually I noticed the following things:

* First, the current implementation throws an error because as per Sklearn Documentation `input_features` needs to be equals to `name_features_in` when using `get_feature_names_out` in `PolynomialFeatures` and `OneHotEncoder`. See this:

image
 * So adding something like `tr_wrap.get_feature_names_out(["Medinc"])` is not valid when trained with more features.

* This means current implementation does not work as expected and cannot be obtained out of the box as mentioned previously. 
 
I implemented a solution when input_features is not None, but I find it not very elegant:

```python
# Get the names of all the new features
added_features = self.transformer_.get_feature_names_out(
                    self.variables_
                )
# Get all the features related to input_features... Sorry for the double for loop
feature_names = []
for feature in added_features:
    for _in in input_features:
        if _in in feature and feature not in feature_names:
            feature_names.append(feature)

# In case of PolynomialFeatures and include_bias is True I need to also retrieve the Bias called '1'
if (
    self.transformer_.__class__.__name__ == "PolynomialFeatures"
    and self.transformer_.include_bias
):
    feature_names = ["1"] + feature_names
```

The reason why I'm using this double for loop (specially for PolynomialFeatures)  is because I need to check all the created features (squared ones, interactions, etc.) related to an input features are included and not repeated. Adittionally in case of include bias, I'm adding '1' as part of the output feature names, I did this because it was the expected behavior for `test_get_feature_names_out_polynomialfeature`. After a lot of trial an error (breaking a lot of tests) I came up to this solution, but I'm totally open to make it better. Didn't find a better way to implement this with lists. I tried sets, but they don't preserve the order of features, so order of features is a bit messy and difficult to test.

The thing is, if an input_variable is not part of the training variables not error is thrown, so what error should I raise if this happens?

PS: I'm currently passing all the tests and I'm getting the results as expected in our discussion, if you accept a solution like this I can push my changes.

Thanks!!

solegalli: Hi @datacubeR 

Thank you so much for so much detail.

I wonder what the point was in offering input_features as parameter in the PolynomialFeatures , if then you can only pass feature_names_in :/

In this case, I think we do not need to modify the code further or add the additional test.

Sorry, that was my bad. For some reason I thought that you could pass one variable and obtain the derived features.

solegalli:
Hi @datacubeR 

Thank you for looking into the get_feature_names_out() issue. I think we can ignore that request then.

Let me know when this is good to go. 

Thanks a lot

datacubeR:
Hi @solegalli, sorry for the delay. I updated my OS and had to reinstall lot of things. I just pushed all the last observations for your review. Please let me know if everything looks ok.

Best,

Alfonso

datacubeR:
Hi @solegalli, sorry for the delay. I updated my OS and had to reinstall lot of things. I just pushed all the last observations for your review. Please let me know if everything looks ok.

Best,

Alfonso

solegalli:
Thank you @datacubeR !!

Great fix.

PR #497, opened 2022-08-13
Ordinal encoder outputs -1 for unknown Categories

datacubeR:
This is my shot for #428. 
I noticed `OrdinalEncoder` inherits the `transform` method from `CategoricalMethodsMixin`. So I added an additional condition to output -1 only for `OrdinalEncoder`.

Additionally I had to fix `test_error_if_input_df_contains_categories_not_present_in_training_df` in order for raise an error in case `errors='raise'` and check correctness if `errors='ignore'`.

### DIFF
--- a/feature_engine/encoding/base_encoder.py
+++ b/feature_engine/encoding/base_encoder.py
@@ -231,7 +231,11 @@ class CategoricalMethodsMixin(BaseEstimator, TransformerMixin):
             else:
                 nan_columns_str = nan_columns[0]
 
-            if self.errors == "ignore":
+            if self.__class__.__name__ == "OrdinalEncoder" and self.errors == "ignore":
+                features = list(self.encoder_dict_.keys())
+                X[features] = X[features].fillna(-1).astype("int")
+
+            elif self.errors == "ignore":
                 warnings.warn(
                     "During the encoding, NaN values were introduced in the feature(s) "
                     f"{nan_columns_str}."
@@ -239,8 +243,7 @@ class CategoricalMethodsMixin(BaseEstimator, TransformerMixin):
             elif self.errors == "raise":
                 raise ValueError(
                     "During the encoding, NaN values were introduced in the feature(s) "
-                    f"{nan_columns_str}."
-                )
+                    f"{nan_columns_str}.")
 
         return X
 
--- a/tests/test_encoding/test_ordinal_encoder.py
+++ b/tests/test_encoding/test_ordinal_encoder.py
@@ -64,22 +64,23 @@ def test_error_if_ordinal_encoding_and_no_y_passed(df_enc):
 def test_error_if_input_df_contains_categories_not_present_in_training_df(
     df_enc, df_enc_rare
 ):
-    # test case 4: when dataset to be transformed contains categories not present
-    # in training dataset
     msg = "During the encoding, NaN values were introduced in the feature(s) var_A."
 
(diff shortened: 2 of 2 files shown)


feature_engine/transformation/boxcox.py
```diff
@@ -139,6 +139,10 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
         # check input dataframe and if class was fitted
         X = super().transform(X)
 
+        # check contains zero or negative values
```
solegalli, line 142: I think I modified this in the main branch already, did I not? it was making the tests fail. 

In any case, I don't think we should change the boxcox in this PR. Could you please remove?

tests/test_transformation/test_boxcox_transformer.py
```diff
@@ -36,16 +36,16 @@ def test_automatically_finds_variables(df_vartypes):
 
 def test_fit_raises_error_if_df_contains_na(df_na):
     # test case 2: when dataset contains na, fit method
+    transformer = BoxCoxTransformer()
```
solegalli, line 39: In fact, I think these are the changes I made in the PR I mention previously. So there is probably a problem with rebase.

Could you try syncing your main branch with the repo's main branch, and then rebasing to this branch so these files disappear?

solegalli (commented):
Hi @datacubeR 

Thanks for the PR.

We've already over-written the inherited parameter in the countEncoder, maybe we can mirror that logic?

See for example here:
https://github.com/feature-engine/feature_engine/blob/65fedf100bc85f364b26838866a264dbf7320841/feature_engine/encoding/count_frequency.py#L139

and here:
https://github.com/feature-engine/feature_engine/blob/65fedf100bc85f364b26838866a264dbf7320841/feature_engine/encoding/count_frequency.py#L162

Would you be able to give it a go?

The idea here is that OrdinalEncoder returns -1 for unseen categories. What we did for the CountEncoder is to return 0 for unseen categories.

datacubeR:
I'm not sure what happened with this branch. My fork is up to date but for some reason it accounts changes in the previous merged commit. I will check what happened here...

feature_engine/transformation/boxcox.py
datacubeR: removed it with rebase

tests/test_transformation/test_boxcox_transformer.py
datacubeR: fixed with rebase

solegalli:
FYI #502

datacubeR:
Hey @solegalli, I don't understand very well if this is already implemented or are there something to be implemented here? Any way, I would like to keep contributing to Feature Engine so if I can help in this PR or any other I would be happy to help.

solegalli:
Hi @datacubeR 

In this PR, we want the OrdinalEncoder to output -1 for unseen categories. It is not implemented yet.

We did something similar for the CountFrequencyEncoder, where we made it output 0 for unseen categories.

My suggestion was that you used the implementation in CountFrequencyEncoder as template for this PR.

Also, since this PR was first made, we made a few structural changes to the main source code, so that imports changed slightly.  So it needs rebasing.

It would be super useful if you could finish this PR first, which should include very few code changes. And then I would be more than happy to tag you in a new PR :)

Thanks a lot!

solegalli:
functionality added in #539

Issue #494, opened 2022-08-06
Adding auto threshold to DropHighPSIFeatures


GLevV:
**Is your feature request related to a problem? Please describe.**
Staple thresholds 0.1 and 0.25 are empirical, but there are alternatives to calculate threshold based on data and parameters of the transformation

**Describe the solution you'd like**
There are two formulas [described in this dissertation](https://scholarworks.wmich.edu/cgi/viewcontent.cgi?article=4249&context=dissertations) to calculate psi threshold from number of bins and number of datapoints in base and test datasets. From my experiments Chi2(0.999, bins) works best. We could keep default parameter as 0.25, but also provide 'auto' option.

solegalli:
Thanks @GLevV for the suggestion.

@gverbock what do you think about this?

gverbock:
I like the idea to use the number of points to set a common threshold for all features. On the other hand the number of bins is arbitrary so it there will be an arbitrary factor anyway (still need to read the dissertation on that topic). @GLevV  what is your view on this?

GLevV:
Yes, threshold will depend on size of dataset (and on `split_frac`)  and number of bins, which kinda makes sense. In practice it gives lower thresholds than 0.1 or 0.25, but we could keep the defaults as is.

Should be pretty straightforward to implement, I think

solegalli:
@GLevV would you like to give it a go?

GLevV:
@solegalli ye, why not

GLevV:
closed with #498

PR #498, opened 2022-08-13
add function to calculate PSI threshold automatically 

GLevV:
PR for #494

### DIFF
--- a/docs/user_guide/selection/DropHighPSIFeatures.rst
+++ b/docs/user_guide/selection/DropHighPSIFeatures.rst
@@ -77,7 +77,9 @@ For 'auto' formula we are using chi2 approximation, proposed by B. Yurdakul:
     threshold = \chi^2_{(q,B-1)} . (\frac{1}{N} + \frac{1}{M})
 
 where p - percentile (p-value), B - number of bins, N - size of basis dataset, N - size of test dataset.
+
 In our implementation we are using 99.9th percentile.
+
 As mentioned above number of bins has an impact on PSI value, because with higher number of bins it is easier to find divergence in data and vice versa.
 Same could be said about dataset size - the more data we have, the harder it is to find the difference (if the shift is not drastic).
 This formula tries to catch these relationships and adjust threshold to correctly detect feature drift.
--- a/docs/user_guide/selection/DropHighPSIFeatures.rst
+++ b/docs/user_guide/selection/DropHighPSIFeatures.rst
@@ -68,6 +68,19 @@ to the PSI value. The most commonly used thresholds are:
 - Below 10%, the variable has not experienced a significant shift.
 - Above 25%, the variable has experienced a major shift.
 - Between those two values, the shift is intermediate.
+- 'auto' threshold will be calculated based on size of dataset and number of bins.
+
+For 'auto' formula we are using chi2 approximation, proposed by B. Yurdakul:
+
+.. math::
+
+    threshold = \chi^2_{(q,B-1)} . (\frac{1}{N} + \frac{1}{M})
+
+where p - percentile (p-value), B - number of bins, N - size of basis dataset, N - size of test dataset.
+In our implementation we are using 99.9th percentile.
+As mentioned above number of bins has an impact on PSI value, because with higher number of bins it is easier to find divergence in data and vice versa.
+Same could be said about dataset size - the more data we have, the harder it is to find the difference (if the shift is not drastic).
+This formula tries to catch these relationships and adjust threshold to correctly detect feature drift.
 
 
 Procedure
--- a/docs/user_guide/selection/DropHighPSIFeatures.rst
+++ b/docs/user_guide/selection/DropHighPSIFeatures.rst
@@ -68,20 +68,20 @@ to the PSI value. The most commonly used thresholds are:
 - Below 10%, the variable has not experienced a significant shift.
 - Above 25%, the variable has experienced a major shift.
 - Between those two values, the shift is intermediate.
-- 'auto' threshold will be calculated based on size of dataset and number of bins.
+- 'auto': the threshold will be calculated based on the size of the base and target datasets and the number of bins.
 
-For 'auto' formula we are using chi2 approximation, proposed by B. Yurdakul:
+When 'auto', the threshold is calculated using the chi2 approximation, proposed by B. Yurdakul:
 
 .. math::
 
     threshold = \chi^2_{(q,B-1)} . (\frac{1}{N} + \frac{1}{M})
 
-where p - percentile (p-value), B - number of bins, N - size of basis dataset, N - size of test dataset.
+where q is the percentile, B is the number of bins, N is the size of basis dataset, N is the size of test dataset.
 
-In our implementation we are using 99.9th percentile.
+In our implementation, we are using the 99.9th percentile.
 
-As mentioned above number of bins has an impact on PSI value, because with higher number of bins it is easier to find divergence in data and vice versa.
-Same could be said about dataset size - the more data we have, the harder it is to find the difference (if the shift is not drastic).
+As mentioned above, the number of bins has an impact on PSI value, because with a higher number of bins it is easier to find divergence in data 
+and vice versa. The same could be said about dataset size - the more data we have, the harder it is to find the difference (if the shift is not drastic).
 This formula tries to catch these relationships and adjust threshold to correctly detect feature drift.
 
 
--- a/feature_engine/selection/drop_psi_features.py
+++ b/feature_engine/selection/drop_psi_features.py
@@ -86,10 +86,6 @@ class DropHighPSIFeatures(BaseSelector):
 
     More details in the :ref:`User Guide <psi_selection>`.
 
-    References
-    ----------
-    https://scholarworks.wmich.edu/cgi/viewcontent.cgi?article=4249&context=dissertations
-
     Parameters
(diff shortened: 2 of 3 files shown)


feature_engine/selection/drop_psi_features.py
```diff
@@ -135,10 +136,12 @@ class DropHighPSIFeatures(BaseSelector):
         test) will be switched. This is important because the PSI is not symmetric,
         i.e., PSI(a, b) != PSI(b, a)).
 
-    threshold: float, default = 0.25.
+    threshold: float, str, default = 0.25.
         The threshold to drop a feature. If the PSI for a feature is >= threshold, the
         feature will be dropped. The most common threshold values are 0.25 (large shift)
         and 0.10 (medium shift).
+        If 'auto' than threshold will be calculated based on size of basis and test 
+        dataframes and number of bins with formula proposed by B. Yurdakul.
```
solegalli, line 144: I think this 2 lines need a bit of editing for typo and grammar. Could you take a look? Also, I would add a note "See references below", so that people know where to find that formula.

solegalli (commented):
Hi @GLevV 

This is looking good.

Could we add a test to ensure that auto threshold is working as expected?

I think, many users would not know what this formula is about. It would be great to add a small explanation, and even a small demo of what this new threshold is doing in the user guide, somewhere in this file: https://github.com/feature-engine/feature_engine/blob/65fedf100bc85f364b26838866a264dbf7320841/docs/user_guide/selection/DropHighPSIFeatures.rst.

Also, the style checks are failing, we are using black and flake8 to fix and test.

feature_engine/selection/drop_psi_features.py
GLevV: Ok, will do

feature_engine/selection/drop_psi_features.py
```diff
@@ -342,6 +352,15 @@ def fit(self, X: pd.DataFrame, y: pd.Series = None):
                 f"and {test_df.shape[0]} samples in the test set. "
                 "Please adjust the value of the cut_off or split_frac."
             )
+        if self.threshold == 'auto':
```
solegalli, line 355: To follow sklearn conventions, we **cannot** re-assign new values to attributes that are defined by the user when initializing the transformer. In other words, the init attributes can't be modified. Whatever the user set, needs to prevail. 

Also, I think there is a small issue if defining the threshold at this point: below we have the `switch` parameter, that changes the order of basis and test, which, if I understand correctly, would also affect the value of the threshold, because it depends on the sizes of these datasets. 

I suggest, redefining the threshold when we need it, a bit below, when selecting the features, maybe here:
https://github.com/feature-engine/feature_engine/blob/85c82aad1dda3d3d9282f95d26d80c79261a74d5/feature_engine/selection/drop_psi_features.py#L378

something like:

```
if self.threshoold == 'auto':
   threshold == your calculation
else:
   threshold = self.threshold
```
And then here:

https://github.com/feature-engine/feature_engine/blob/85c82aad1dda3d3d9282f95d26d80c79261a74d5/feature_engine/selection/drop_psi_features.py#L395

change `self.threshold` by just `threshold`

docs/user_guide/selection/DropHighPSIFeatures.rst
```diff
@@ -68,6 +68,21 @@ to the PSI value. The most commonly used thresholds are:
 - Below 10%, the variable has not experienced a significant shift.
 - Above 25%, the variable has experienced a major shift.
 - Between those two values, the shift is intermediate.
+- 'auto' threshold will be calculated based on size of dataset and number of bins.
```
solegalli, line 71: Could we modify slightly to

'auto': the threshold will be calculated based on the size of the base and target datasets and the number of bins.

docs/user_guide/selection/DropHighPSIFeatures.rst
```diff
@@ -68,6 +68,21 @@ to the PSI value. The most commonly used thresholds are:
...
+
+For 'auto' formula we are using chi2 approximation, proposed by B. Yurdakul:
```
solegalli, line 73: Could we modify slightly to:

When 'auto', the threshold is calculated using the chi2 approximation, proposed by B. Yurdakul:

docs/user_guide/selection/DropHighPSIFeatures.rst
```diff
@@ -68,6 +68,21 @@ to the PSI value. The most commonly used thresholds are:
...
+
+.. math::
+
+    threshold = \chi^2_{(q,B-1)} . (\frac{1}{N} + \frac{1}{M})
+
+where p - percentile (p-value), B - number of bins, N - size of basis dataset, N - size of test dataset.
```
solegalli, line 79: There is no p in the formula. Did you mean q?

Also, could we replace by a full sentence with verbs? something like this:

where q is the percentile, B is the number of bins, N is the... and M is ...

docs/user_guide/selection/DropHighPSIFeatures.rst
```diff
@@ -68,6 +68,21 @@ to the PSI value. The most commonly used thresholds are:
...
+
+In our implementation we are using 99.9th percentile.
+
+As mentioned above number of bins has an impact on PSI value, because with higher number of bins it is easier to find divergence in data and vice versa.
```
solegalli, line 83: This paragraph has a few typos. Could you have a look?

I tend to use a tool that corrects typos and fixes sentences automatically: https://quillbot.com/grammar-check

docs/user_guide/selection/DropHighPSIFeatures.rst
```diff
@@ -68,6 +68,21 @@ to the PSI value. The most commonly used thresholds are:
...
+Same could be said about dataset size - the more data we have, the harder it is to find the difference (if the shift is not drastic).
```
solegalli, line 84: Could you please adjust this paragraph as well?

feature_engine/selection/drop_psi_features.py
```diff
@@ -135,10 +132,12 @@ class DropHighPSIFeatures(BaseSelector):
         test) will be switched. This is important because the PSI is not symmetric,
         i.e., PSI(a, b) != PSI(b, a)).
 
-    threshold: float, default = 0.25.
+    threshold: float, str, default = 0.25.
         The threshold to drop a feature. If the PSI for a feature is >= threshold, the
         feature will be dropped. The most common threshold values are 0.25 (large shift)
         and 0.10 (medium shift).
+        If threshold='auto', it will be calculated based on size of dataset and
```
solegalli, line 139: Small tweak please:

If 'auto', the threshold will be calculated based on the size of the base and target dataset and the number of bins.

feature_engine/selection/drop_psi_features.py
```diff
@@ -247,8 +252,13 @@ def __init__(
...
             raise ValueError(f"switch must be a boolean. Got {switch} instead.")
 
-        if not isinstance(threshold, (float, int)) or threshold < 0:
-            raise ValueError(f"threshold must be >= 0. Got {threshold} instead.")
+        if (
+            (isinstance(threshold, str) and (threshold != 'auto')) or
+            (isinstance(threshold, (float, int)) and threshold < 0)
+        ):
+            raise ValueError(
+                f"threshold must be positive float or 'auto'. Got {threshold} instead."
```
solegalli, line 260: small tweak:

threshold must be **a** positive float ...

or maybe better:

threshold must be greater than 0 or ...

because it could also be an int as per our definition

feature_engine/selection/drop_psi_features.py
solegalli: Also, thinking it further... in order to be able to test that this threshold is returning the values that we expect, would it not be better to capture the calculation in a hidden method? 

this way, we could add a test to calculate that this method returns the expected threshold values.

self._calculate_auto_threshold(self):
   return your calculation

and then we could test this method with different values of N, M, and bins.

What do you think? 

I think it would also leave us in a better shape if later we want to allow the user to modify the value of the percentile. We already have a test to ensure that nothing will be broken

solegalli (commented):
Hi @GLevV 

Thank you so much for the implementation. It looks good.

I have small requests here and there. The main one, is that we can't redefine init attributes if we want to adhere to sklearn's API conventions. So we need to calculate an intermediate threshold somewhere later in the code.

The second is a suggestion: should we not make this calculation a hidden method? this way we could add a test to corroborate that we get the expected values.

All other requests are minor tweaks in the sentences, so that we can leave it ready for publication :)  

Please let me know what you think!

Thank you so much!

tests/test_selection/test_drop_high_psi_features.py
```diff
@@ -90,6 +91,54 @@ def test_fit_attributes(df):
...
+        "var_0": 0.043828484052281,
+        "var_1": 0.040929870747665395,
+        "var_2": 0.04330418495156895,
+        "var_3": 0.03773286532548153,
+        "var_4": 0.05047388515663041,
+        "var_5": 0.014717735595712466,
+        "drift_1": 8.283089355027482,
+        "drift_2": 8.283089355027482,
+    }
+    assert math.isclose(
```
solegalli, line 121: Hi @GLevV 

This is super peaky, but could we take this assert into a separate test? i.e., def test_auto_threshold():

We could pass a combination of integers with parametrize, I don't think we need a dataframe for the test.

PR #500, opened 2022-08-17
replace concat by group by target in encoders

### LINKED ISSUE #427: replace concat in encoders by y_train.groupby(X_train["variable"]).mean().sort_values()
Applies to the following encoders:

- OrdinalEncoder
- MeanEncoder
- PRatioEncoder
- WoEEncoder

Currently we concatenate the target to the predictors (X_train) to determine the mappings. This is not necessary. We can determined them automatically with syntax like this:

y_train.groupby(X_train["A7"]).mean().sort_values()

therefore removing one unnecessary step in the calculation, the pd.concat step

solegalli:
more elegant,  simpler and, above all, faster implementation.

closes #427


solegalli:
@GLevV 

Could you have a look at this PR, it is small, and let me know if you think there is a better or more performant implementation to capture these encodings?

I will greatly appreciate that :)

feature_engine/encoding/woe.py
```diff
@@ -151,40 +151,36 @@ def fit(self, X: pd.DataFrame, y: pd.Series):
...
+        self._get_feature_names_in(X)
 
         self.encoder_dict_ = {}
 
-        total_pos = temp["target"].sum()
-        total_neg = len(temp) - total_pos
-        temp["non_target"] = np.where(temp["target"] == 1, 0, 1)
+        total_pos = y.sum()
+        total_neg = len(y) - total_pos
+        inverse_y = y.map({1: 0, 0:1}).copy()
```
GLevV: arithmetical operations would be faster than mapping, we can do it like this:
`total_pos = y.sum()`
`inverse_y = (y - 1).abs().copy()`
`total_neg = inverse_y.sum()`
Should be faster

solegalli: Thank you @GLevV 

I also replaced the previous mapping by scaling :)

PR #503, opened 2022-08-17
add deprecation warning to errors/unseen

solegalli:


solegalli:
too difficult to add the deprecation warning. It just creates errors everywhere.

PR #504, opened 2022-08-18
Add code coverage

### LINKED ISSUE #323: include code coverage
At the moment, we have no visibility of test coverage. I would like to include this in the package.

Here some useful links I gathered:

https://gist.github.com/dnozay/5b9b818ff0dc857d1358

https://pytest-cov.readthedocs.io/en/latest/tox.html

https://coverage.readthedocs.io/en/6.1.1/config.html

Also check imbalanced-learn package, they have some code to run coverage.

Morgan-Sell:
Code coverage will allow us to measure how much of the feature-engine source code is tested.

Closes #323.

Notes from #323:

At the moment, we have no visibility of test coverage. I would like to include this in the package.

Here some useful links I gathered:

https://gist.github.com/dnozay/5b9b818ff0dc857d1358

https://pytest-cov.readthedocs.io/en/latest/tox.html

https://coverage.readthedocs.io/en/6.1.1/config.html

Also check imbalanced-learn package, they have some code to run coverage.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,8 @@
+# configuration for coverage.py
+
+[run]
+branch = True
+source = feature_engine
+omit =
+    */tags.py
+    */_docstrings/*
--- a/.coveragerc
+++ b/.coveragerc
@@ -6,3 +6,12 @@ source = feature_engine
 omit =
     */tags.py
     */_docstrings/*
+    */_datetime_constants.py
+    */_docstings.py
+    */_selection_constants.py
+
+[report]
+exclude_lines =
+    pragma: no cover
+    def _more_tags
+show_missing = True
--- a/docs/about/authors.rst
+++ b/docs/about/authors.rst
@@ -18,4 +18,7 @@
             <a href='https://github.com/nicogalli'><img src='https://avatars.githubusercontent.com/nicogalli?v=4' class='avatar'width="120" height="120"/></a> <br />
             <p>Nicolas Galli</p>
         </div>
+        <div class="author">
+            <a href='https://github.com/morgan-sell'><img src='https://avatars.githubusercontent.com/morgan-sell?v=4' class='avatar' width="120" height="120" /></a> <br />
+            <p>Morgan Sell</p>
     </div>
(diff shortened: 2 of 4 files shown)


Morgan-Sell:
hola @solegalli,

I created the `.coveragerc` file.

Two questions:

1. What am I missing from `omit` and `excluded_lines`? I suspect there are other files in which unit tests do not apply.
2. Is there anywhere in the source code where we should add `# pragma: no cover` so the code is excluded from the code-coverage calculations?

solegalli:
hi @Morgan-Sell 

to be honest I have not used code coverage before, so I need to do some research to answer those questions. In short, I don't know :_(

Morgan-Sell:
Hi @solegalli,

For our current needs, I believe the two main sections are `omit` and `excluded_lines`. These are the definitions and examples from coverage documentation:

1. `omit` (multi-string): a list of file name patterns, the files to leave out of measurement or reporting.
```
[run]
omit =
    # omit anything in a .local directory anywhere
    */.local/*
    # omit everything in /usr
    /usr/*
    # omit this single file
    utils/tirefire.py
```

2. `exclude_lines` (multi-string): a list of regular expressions. Any line of your source code containing a match for one of these regexes is excluded from being reported as missing... If you use this option, you are replacing all the exclude regexes, so you’ll need to also supply the “pragma: no cover” regex if you still want to use it.

You can exclude lines introducing blocks, and the entire block is excluded. If you exclude a def line or decorator line, the entire function is excluded.

Be careful when writing this setting: the values are regular expressions that only have to match a portion of the line. For example, if you write ..., you’ll exclude any line with three or more of any character. If you write pass, you’ll also exclude the line my_pass="foo", and so on.

```
[report]
exclude_lines =
    pragma: no cover
    def __repr__
    if self.debug:
    if settings.DEBUG
    raise AssertionError
    raise NotImplementedError
    if 0:
    if __name__ == .__main__.:
    class .*\bProtocol\):
    @(abc\.)?abstractmethod
```

Here's [imbalanced-learn .coveragerc](https://github.com/scikit-learn-contrib/imbalanced-learn/blob/master/.coveragerc). With imbalanced-learn, I think using `source` and `include` is redundant.

Issue #513, opened 2022-08-24
replace loop by broadcasting in cyclical features


solegalli:
The class creates features in a loop:

https://github.com/feature-engine/feature_engine/blob/9d4b67e34973fb0dad6a1df6ba88a3305a2c5bb7/feature_engine/creation/cyclical_features.py#L151-L154

I think we could replace that by broadcasting and make it faster

solegalli:
Hi @GLevV

You are really good at programming,  so I thought I would pick your brain once more :)

Would it be possible to replace this loop by array / pandas operations? and if yes, would you be able to give it a go?

I think you should be able to obtain the variable names from the get_feature_names_out() method passing the input variables

GLevV:
Will look into it

GLevV:
In this case loop is faster, see PR

solegalli:
Thank you @GLevV 

this was really helpful. I'll close the issue then !

PR #515, opened 2022-08-26
[WIP] Cyclic speedup

GLevV:
improvement from #513


GLevV:
Tests failing because of the order of columns. Do I change testing function?

GLevV:
after some testing with 2 columns and 1mil rows

> new - 895 ms ± 21 ms per loop (mean ± std. dev. of 10 runs, 100 loops each)
> old  - 147 ms ± 281 µs per loop (mean ± std. dev. of 10 runs, 100 loops each)

so loop is faster

solegalli:
Thank you @GLevV

Issue #315, opened 2021-10-15
Smoothing on Mean Encoder


hectorpatino:
**Is your feature request related to a problem? Please describe.**
Hi Sole, currently Im doing the mini course of [feature engineering](https://www.kaggle.com/ryanholbrook/target-encoding) on kaggle. The last seccion is about Mean Encoding and how it can create risks of overfitting specially with unknown or rare categories. They mention that smooting is a way to deal with this. I checked the [code ](https://github.com/feature-engine/feature_engine/blob/main/feature_engine/encoding/mean_encoding.py#L12) but I don't see anything like that. So ¿should MeanEncoder has a smoothing parameter?

solegalli:
Yes, some users also requested to introduce this transformer or smoothing function in this issue: #160 

I am still wondering if it is worth to bring in to Feature-engine the functionality that already exists in category_encoders. Category encoders also allows you to select columns, so not sure how much value it would add.

What are your thoughts?

hectorpatino:
As someone mention in the issue there is no need to reinvent the wheel. Maybe an update on docstring suggesting `category_encoders` for smoothing?

solegalli:
Great! In that case, please make a PR to the branch "pydata_template" as the doc files have been reorganized quite a bit compared to master. 

And you probably want to modify [this file](https://github.com/feature-engine/feature_engine/blob/pydata_template/docs/user_guide/encoding/MeanEncoder.rst) and / or [this file](https://github.com/feature-engine/feature_engine/blob/pydata_template/feature_engine/encoding/mean_encoding.py).

solegalli:
Already added links to category encoders in the class api docs. We still need to add some details in the user guide.

hectorpatino:
Ive been extremely busy sole but I Will try to do it this weekend

solegalli:
added links to docs in #314

solegalli:
Reopening as I think it might be worth bringing in this functionality to our class as well. It is not so complicated.

Morgan-Sell:
hola @solegalli,

I can work on this task. Although, I'm always wary when you write, "It is not so complicated." ;)

solegalli:
Go for it. 

We need to implement the transformation described in the article cited in [this documentation](https://contrib.scikit-learn.org/category_encoders/targetencoder.html) and I guess we can borrow the logic to calculate the prior and posterior probability from category encoders as well.

[...]

PR #517, opened 2022-09-11
Add smoothing to mean encoder

GLevV:
Will solve #315 

- [x] add tests
- [x] add info to docs

### DIFF
--- a/feature_engine/encoding/mean_encoding.py
+++ b/feature_engine/encoding/mean_encoding.py
@@ -147,7 +147,7 @@ class MeanEncoder(CategoricalInitExpandedMixin, CategoricalMethodsMixin):
         variables: Union[None, int, str, List[Union[str, int]]] = None,
         ignore_format: bool = False,
         unseen: str = "ignore",
-        smoothing: Union[int, float, str] = 0.0, 
+        smoothing: Union[int, float, str] = 0.0,
     ) -> None:
         super().__init__(variables, ignore_format, unseen)
         if (
--- a/feature_engine/encoding/mean_encoding.py
+++ b/feature_engine/encoding/mean_encoding.py
@@ -144,10 +144,10 @@ class MeanEncoder(CategoricalInitExpandedMixin, CategoricalMethodsMixin):
 
     def __init__(
         self,
-        smoothing: Union[int, float, str] = 0.0,
         variables: Union[None, int, str, List[Union[str, int]]] = None,
         ignore_format: bool = False,
         unseen: str = "ignore",
+        smoothing: Union[int, float, str] = 0.0, 
     ) -> None:
         super().__init__(variables, ignore_format, unseen)
         if (
--- a/feature_engine/encoding/mean_encoding.py
+++ b/feature_engine/encoding/mean_encoding.py
@@ -67,7 +67,7 @@ class MeanEncoder(CategoricalInitExpandedMixin, CategoricalMethodsMixin):
     smoothing: int, float, str, default=0.0
         Smoothing factor, should be >= 0. If 0 then no smoothing is applied,
         higher values lead to stronger smoothing (higher weight of prior mean).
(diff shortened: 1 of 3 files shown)


solegalli:
Hi @GLevV 

Thank you for the PR. I missed that you marked it ready for review for a few days because it still said WIP, and I didnt look all the way through the history :/ . And then I was a bit busy with a deadline that I had for yesterday.

I will look at this tomorrow.
Apologies for the delay.

feature_engine/encoding/mean_encoding.py
```diff
@@ -143,8 +156,20 @@ def fit(self, X: pd.DataFrame, y: pd.Series):
 
         self.encoder_dict_ = {}
 
+        y_mean = y.mean()
```
solegalli, line 159: this would be the prior (talking to myself)

feature_engine/encoding/mean_encoding.py
```diff
@@ -143,8 +156,20 @@ def fit(self, X: pd.DataFrame, y: pd.Series):
...
+        if self.a == 'auto':
+            y_var = y.var(ddof=0)
         for var in self.variables_:
-            self.encoder_dict_[var] = y.groupby(X[var]).mean().to_dict()
+            if self.a == 'auto':
+                damping = y.groupby(X[var]).var(ddof=0) / y_var
```
solegalli, line 164: Is this described in the article of micci barrera?

feature_engine/encoding/mean_encoding.py
```diff
@@ -143,8 +156,20 @@ def fit(self, X: pd.DataFrame, y: pd.Series):
...
+            else:
+                damping = self.a
+            counts = X[var].value_counts()
+            smoothing = counts / (counts + damping)
```
solegalli, line 168: In category encoders, the smoothing parameter is weighted by the number of observations per category:
https://github.com/scikit-learn-contrib/category_encoders/blob/80288a0aa3d9bc4406f3d6df13a49113bbc3fedf/category_encoders/target_encoder.py#L229-L231

feature_engine/encoding/mean_encoding.py
GLevV: Yes, 3.1 last paragraphs

feature_engine/encoding/mean_encoding.py
GLevV: I did not use sigmoid formula, I used (x / (x+smoothing)) formula, which does not have this parameter

solegalli: Oh, I see, so `damping = y.groupby(X[var]).var(ddof=0) / y_var` is equation 6 in the article and `x / (x+smoothing)` is equation 5, am I right? Both in section 3.1

Any particular reason you chose this implementation?

I think it might be useful to offer some guidance regarding what are good values of a, if they exist. Or is it just by cross-validation and hyperparamter search?

GLevV: I choose this formula for two reasons:
1. Almost any implementation of target encoding uses sigmoid formula (category_encoders, dirty_cat etc), and it's meaningless to just reimplement it for 100th time.
2. This formula has an interesting auto option that I've never encountered in target encoding estimators (or maybe I searched not good enough, idk), so it seemed interesting.

As for meaningful smoothing values, I think in general it's goes like this "the more noise you have in target, the bigger value of smoothing should be", but there is no advice on concrete values. Most of the time its optimized by cv, sometimes its just set as 1. We will have an auto option for that.

solegalli: Makes sense. Thank you!

solegalli:
FYI @GLevV 

https://github.com/GLevV/feature_engine/pull/1

Other than that, this is good to go. Thank you so much!

Issue #426, opened 2022-04-25
one hot encoder of frequent categories should capture top categories with similar number of observations


solegalli:
At the moment the OHE will create binary variables for the top k categories with most observations. However, if 2 categories should have similar number of observations, only 1 of them will be encoded and the other one ignored. We probably want to encode both?

BromleyJames:
Hi @solegalli , I want to attempt this as my first issue, but I'm a little confused as to what is needed here.

Are you talking about the case where k=2 categories, in which case we have only 1 binary output?

e.g. current state

| Input | Encoding_1_green |
|------| ----------|
| green | 1 |
| green | 1 |
| green | 1 |
| green | 1 |
| green | 1 |
| blue | 0 |
| blue | 0 |
| blue | 0 |
| red | 0 |

e.g. desired state

| Input | Encoding_1_green | Encoding_2_blue |
|------| ----------| ----------|
| green | 1 | 0 |
| green | 1 | 0 |
| green | 1 | 0 |
| green | 1 | 0 |
| green | 1 | 0 |
| blue | 0 | 1 |
| blue | 0 | 1 |
| blue | 0 | 1 |
| red | 0 | 0 |

solegalli:
The OHEEncoder has the parameter top_categories, that if set to 3 for example, the encoder will count the number of observations per category, order them from those with most to those with less categories, and then select the top 3.

So if the variable City has the following categories: London, Berlin, Manchester, Amsterdam, Buenos Aires, it will count the number of observations in each city, and select the top 3.

Now say the following happens:

- London 10 observations
- Berlin 9 observations
- Manchester 8 observations
- Amsterdam 7 observations
- Buenos Aires 6 observations

then the encoder will encode only London, Berlin and Manchester, and that should be fine. This is the current functionality.

Now if the following happens:

- London 10 observations
- Berlin 9 observations
- Manchester 8 observations
- Amsterdam 8 observations
- Buenos Aires 6 observations

The encoder will only encode London, Berlin and Manchester, even though Amsterdam also has 8 observations, just like Manchester, So the selection of the 2 cities here is kind of random. This is also the current functionality, and what I think needs to be fixed in this PR.

The correct functionality should be that the transformer picks London, Berlin, Manchester and Amsterdam for a situation like this.

Does this make sense?

solegalli:
Btw, welcome to the project @BromleyJames 

Where did I leave my manners?

BromleyJames:
[...]

PR #537, opened 2022-09-26
add new StringSimilarityEncoder transformer

### LINKED ISSUE #530: String Similarity Encoder
Recently dirty_cat dropped support for edit distances and left only ngram similarity, so I think it's time to add encoder like this.
The whole concept is pretty simple: instead of encoding categories if there is an exact match like OneHotEncoder, we encode categories based on their string similarity. It's very useful if dataset has 'dirty' (poorly-defined) categories (for example, professions). Since only very few libraries have something like that, it could be beneficial to add it here.

Proposition:
Add StringSimilarityEncoder with only basic functionality for now, i.e. support one similarity measure (gestalt pattern matching will do, it's from standard library so it will be fast and no additional dependencies required) and only 'most_frequent' categories, i.e. just like OneHotEncoder we encode all or top-k categories in the variable, no kmeans method for finding prototypes for now.

[dirty_cat similarity encoder](https://dirty-cat.github.io/stable/generated/dirty_cat.SimilarityEncoder.html), [twds article](https://towardsdatascience.com/similarity-encoding-for-dirty-categories-using-dirty-cat-d9f0b581a552)

GLevV:
closes #530


GLevV:
Example notebook https://github.com/feature-engine/feature-engine-examples/pull/11

docs/index.rst
```diff
@@ -130,6 +130,7 @@ Categorical Encoders: Encoders
 - :doc:`api_doc/encoding/PRatioEncoder`: replaces categories by a ratio of probabilities
 - :doc:`api_doc/encoding/DecisionTreeEncoder`: replaces categories by predictions of a decision tree
 - :doc:`api_doc/encoding/RareLabelEncoder`: groups infrequent categories
+- :doc:`api_doc/encoding/StringSimilarityEncoder`: performs string similarity encoding for poorly definde categories
```
solegalli, line 133: typo: **defined**

feature_engine/encoding/similarity_encoder.py
```diff
@@ -0,0 +1,298 @@
...
+class StringSimilarityEncoder(CategoricalInitMixin, CategoricalMethodsMixin):
+    """
+    The StringSimilarityEncoder() replaces categorical variables by a set of
+    float variables representing similarity between unique categories in the variable.
+    This new variables will have values in range between 0 and 1, where 0 is the least
+    similar and 1 is the exact match.
+    This encoding is an alternative to OneHotEncoder in the case of poorly
+    defined categorical variables.
+
+    The encoder will create k variables, where k is the number of unique categories.
```
solegalli, line 50: I am trying to understand the logic without reading the article.

If a variable was say 10 well defined categories, like Data Analyst, Product Manager etc, you probably want to encode them into 10 float variables, each one would represent 1 category, and the value would be the distance between the value of the observation to the category in question.

If the variable has dirty categories, were the professions are mispelled, instead the user would encode the top categories, hoping that the top ones are spelled correctly. 

Do I get this right?

feature_engine/encoding/similarity_encoder.py
```diff
@@ -0,0 +1,298 @@
...
+
+    The encoder has the option to specify the behaviour when NaN is present in the
+    variable, see parameter `handle_missing`.
+
+    The encoder will encode only categorical variables by default (type 'object' or
+    'categorical'). You can pass a list of variables to encode. Alternatively, the
+    encoder will find and encode all categorical variables (type 'object' or
+    'categorical').
+
+    With `ignore_format=True` you have the option to encode numerical variables as well.
```
solegalli, line 65: This parameter exists for the other encoders because sometimes, numerical variables with numbers like 1,2,3 could be categorical, for example if they represent months, or days, or some sort of internal code.

For this transformer, would it make sense to calculate distances between numbers? When would this be useful?

feature_engine/encoding/similarity_encoder.py
```diff
@@ -0,0 +1,298 @@
...
+
+    Parameters
+    ----------
+    top_categories: int, default=None
+        If None, dummy variables will be created for each unique category of the
+        variable. Alternatively, we can indicate in the number of most frequent
+        categories to encode. In this case, similarity variables will be created
+        only for those popular categories and the rest will be ignored.
+
+    handle_missing : str, default='impute'
```
solegalli, line 81: is impute as default the safer option for the user?

feature_engine/encoding/similarity_encoder.py
```diff
@@ -0,0 +1,298 @@
...
+        Action to perform when NaN is seen.
+            'error' - raise an error;
+            'impute' - impute NaN with an empty string;
+            'ignore' - ignore NaN and leave them in resulting columns.
+
+    {variables}
+
+    {ignore_format}
+
```
solegalli, line 90: what would this encoder do for unseen categories? should we add parameter unseen and some logic?

feature_engine/encoding/similarity_encoder.py
```diff
@@ -0,0 +1,298 @@
...
+    """
+
+    def __init__(
+        self,
+        top_categories: Union[None, int] = None,
+        handle_missing: str = 'impute',
+        variables: Union[None, int, str, List[Union[str, int]]] = None,
+        ignore_format: bool = False,
+    ):
+        if top_categories and not isinstance(top_categories, int):
```
solegalli, line 143: For the error messages could we add f"Got {param_value} instead."

We are rolling that out for all new classes, and slowly updating the old ones.

feature_engine/encoding/similarity_encoder.py
```diff
@@ -0,0 +1,298 @@
...
+            raise ValueError(
+                "handle_missing should be one of 'error', 'impute' or 'ignore'"
+            )
+        super().__init__(variables, ignore_format)
+        self.top_categories = top_categories
+        self.handle_missing = handle_missing
+
+    def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
+        """
+        Learns the unique categories per variable. If top_categories is indicated,
```
solegalli, line 157: what happens if the user selects 5 top categories, and then categories 5 and 6 are equally popular? 

At the moment, our OneHotEncoder would select category 5th and ignore 6th, but we have an issue to try and fix this behaviour so that it selects both. Should we incorporate this here as well?

feature_engine/encoding/similarity_encoder.py
```diff
@@ -0,0 +1,298 @@
...
+            new_features = np.asarray(
+                [f'{var}_{i}' if i else f'{var}_None' for i in self.encoder_dict_[var]]
+            )
+            X.loc[:, new_features] = encoded
+            if self.handle_missing == 'ignore':
+                X.loc[X[var].isna(), new_features] = np.nan
+
+        return X.drop(self.variables_, axis=1)
+
+    def get_feature_names_out(self, input_features: Optional[List] = None) -> List:
```
solegalli, line 254: We've got a Mixin for this method that handles all the checks and makes it compatible with the sklearn pipeline.
https://github.com/feature-engine/feature_engine/blob/main/feature_engine/_base_transformers/mixins.py

We only need to add 2 functions, 1 to generate the new feature names, and the second one that would be used by the mixin to add the new features, see for example the OneHotEncoder

https://github.com/feature-engine/feature_engine/blob/93cb37b6650f65281c44ec6d8555d3fb0e55a82c/feature_engine/encoding/one_hot.py#L267-L281

Also, if we have the function to create the new feature names, we don't have to repeat the logic in this method and then also in transform. We can call this function from both. Would you be able to adjust this logic?

tests/test_encoding/test_similarity_encoder.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+def test_error_if_handle_missing_invalid():
+    with pytest.raises(ValueError):
+        StringSimilarityEncoder(handle_missing='propagate')
+
+
+def test_nan_behaviour(df_enc_big, df_enc_big_na):
+    with pytest.raises(ValueError):
+        encoder = StringSimilarityEncoder(handle_missing='error')
```
solegalli, line 82: could we take this line out of pytest.raises? this line should not fail

tests/test_encoding/test_similarity_encoder.py
```diff
@@ -0,0 +1,176 @@
...
+        encoder.fit(df_enc_big_na)
+
+    with pytest.raises(ValueError):
+        encoder = StringSimilarityEncoder(handle_missing='error')
+        encoder.fit(df_enc_big)
```
solegalli, line 87: could we take these 2 lines out of pytest.raises? they should not fail

tests/test_encoding/test_similarity_encoder.py
```diff
@@ -0,0 +1,176 @@
...
+    with pytest.raises(ValueError):
+        StringSimilarityEncoder(top_categories=0.5)
+
+
+def test_error_if_handle_missing_invalid():
+    with pytest.raises(ValueError):
+        StringSimilarityEncoder(handle_missing='propagate')
+
+
+def test_nan_behaviour(df_enc_big, df_enc_big_na):
```
solegalli, line 80: we should probably break this test in 3, one for each param option. That's what I am always told by my developer colleagues anyways.

tests/test_encoding/test_similarity_encoder.py
```diff
@@ -0,0 +1,176 @@
+import pandas as pd
+import pytest
+from feature_engine.encoding import StringSimilarityEncoder
+
+
+def test_encode_top_categories():
```
solegalli, line 6: at the moment, we are testing with letters. Do I understand correctly that the distance will be 1 for each of them?

Should we not add a test with a more realistic scenario, to test that we are indeed using this similarity function from difflib? If someone changes this function for another similarity measure, this test will probably still pass?

feature_engine/encoding/similarity_encoder.py
```diff
@@ -0,0 +1,298 @@
...
+    n_features_in_=_n_features_in_docstring,
+    fit_transform=_fit_transform_docstring,
+)
+class StringSimilarityEncoder(CategoricalInitMixin, CategoricalMethodsMixin):
+    """
+    The StringSimilarityEncoder() replaces categorical variables by a set of
+    float variables representing similarity between unique categories in the variable.
+    This new variables will have values in range between 0 and 1, where 0 is the least
+    similar and 1 is the exact match.
+    This encoding is an alternative to OneHotEncoder in the case of poorly
```
solegalli, line 47: Should we add information regarding how the similarity is calculated? I'd be curios if I was a user. Is it too long to explain?

tests/test_encoding/test_similarity_encoder.py
```diff
@@ -0,0 +1,176 @@
...
+    encoder = StringSimilarityEncoder(handle_missing='impute')
+    X = encoder.fit_transform(df_enc_big_na)
+    assert (X.isna().sum() == 0).all(axis=None)
+
+    encoder = StringSimilarityEncoder(handle_missing='ignore')
+    X = encoder.fit_transform(df_enc_big_na)
+    assert (X.isna().any(1) == df_enc_big_na.isna().any(1)).all()
+
+
+def test_get_feature_names_out(df_enc_big, df_enc_big_na):
```
solegalli, line 99: we should probably also break down this test in smaller tests.

solegalli (commented):
Hi @GLevV 

The code looks great! Thank you so much for such an amazing and complete contribution.

I have a couple of questions, some of them because I am not 100% familiar with the methodology, and some of them regarding the functionality of the class. Could you have a look and let me know what you think?

Also, in the user guide and in the docstrings, there are a few typos and small grammar things that need to be fixed. I normally use this tool: https://quillbot.com/grammar-check since I am not a native speaker.

Would you be able to run the paragraphs with the tool to have them fixed? If not I could do it next week.

Thanks a lot!

feature_engine/encoding/similarity_encoder.py
GLevV: Yes, you are correct. 

On top of that not only misspells/typos, but also categories that are close, like in this example - Data Analyst and Product Analyst should be closer than Data Analyst and Software Engineer. In case of OHE they will be equidistant

feature_engine/encoding/similarity_encoder.py
GLevV: For example in cases where numbers could be represented as a string, for example barcode (substracting them has no sense, but calculating edit distance does). Another one will be ticket column in titanic dataset, sometimes ticket numbers will have different formatting and typos, and this encoding could mitigate that.

But I guess this option could be removed, no hard opinion.

feature_engine/encoding/similarity_encoder.py
GLevV: It makes more sense to impute NaNs as empty strings. It's also a default option for similarity encoder in dirty_cat

feature_engine/encoding/similarity_encoder.py
GLevV: I don't know about this issue, can you send me a link to it?

tests/test_encoding/test_similarity_encoder.py
GLevV: I don't think that we should test functions from standard library, that would be excessive

feature_engine/encoding/similarity_encoder.py
GLevV: Ye, it's not very straightforward, but we can sen user to difflib.SequenceMatcher class user doc

GLevV:
Should there be a copyright header?

feature_engine/encoding/similarity_encoder.py
solegalli: Allowing numerical columns also gives us the chance to test the transformer with the check_estimator from sklearn. Maybe we leave it then. 

It would be good to add this example in the user guide.

solegalli:
> hould there be a copyright head

What do you mean by copyright header?

feature_engine/encoding/similarity_encoder.py
solegalli: I think we can borrow these 3 sentences from the difflib library:

Return a measure of the sequences’ similarity as a float in the range [0, 1].

Where T is the total number of elements in both sequences, and M is the number of matches, this is 2.0*M / T. Note that this is 1.0 if the sequences are identical, and 0.0 if they have nothing in common.

And mention that this transformer uses difflib.SequenceMatcher under the hood, for the most experienced user.

feature_engine/encoding/similarity_encoder.py
```diff
@@ -0,0 +1,276 @@
...
+    def __init__(
+        self,
+        top_categories: Union[None, int] = None,
+        handle_missing: str = 'impute',
+        variables: Union[None, int, str, List[Union[str, int]]] = None,
+        ignore_format: bool = False,
+    ):
+        if top_categories and not isinstance(top_categories, int):
+            raise ValueError(
+                f"top_categories takes only integer numbers, 1, 2, 3, etc."
```
solegalli, line 147: we need a space between the full stop and the " otherwise the bottom line comes right next to the top one in the error message.

feature_engine/encoding/similarity_encoder.py
solegalli: #426

solegalli (commented):
Hi @GLevV 

Thank you for taking the time to answer my questions and making the code changes. 

One thing I forgot to mention is that we need to add the class here:
https://github.com/feature-engine/feature_engine/blob/main/tests/test_encoding/test_check_estimator_encoders.py

To check compatibility with sklearn and to test some defo params from Feature engine.

I am aware that the docs on Feature-engine's tests is non-existing, so I can help if the tests don't pass.

feature_engine/encoding/similarity_encoder.py
GLevV: I think it should be in the notes

feature_engine/encoding/similarity_encoder.py
GLevV: It's already has whitespace before 'Got', next line.

GLevV:
> > hould there be a copyright head
> 
> What do you mean by copyright header?

For example this

https://github.com/feature-engine/feature_engine/blob/93cb37b6650f65281c44ec6d8555d3fb0e55a82c/feature_engine/encoding/one_hot.py#L1-L2

feature_engine/encoding/similarity_encoder.py
GLevV: I don't think this is an issue. The one category that will be taken among tied will be first in alphabetical order, which is consistent and this is all that matters.

When we define top_k we want to get k features out, not k-1 or k+1 or k+n. This parameter exist to define number of features to be outputted. If you implement this fix of adding all tied categories you will cause more questions and problems to the end user.

solegalli:
Hi @GLevV 

Thank you very much for the contributions and for taking the time to go through my questions.

FYI: I made this PR to your repo:
https://github.com/GLevV/feature_engine/pull/2

There a a few things that we need to look at:

1)

As it is, the encoder is failing one of feature-engine's general tests, because you can't add "" in fillna to a categorical variable (encoded as categorical, instead of object). The value "" needs to be added as a category first. See for example how we handle that here:
https://github.com/feature-engine/feature_engine/blob/93cb37b6650f65281c44ec6d8555d3fb0e55a82c/feature_engine/imputation/categorical.py#L210-L216

2)

The test `test_nan_behaviour_impute` tests that the nan are encoded. But it does not test the fact that nan are replaced by "". 
At first, this may seem trivial, but if for example, another contributor decides that we should change "" by "_", while the output of the transformer may still make sense, this could break backward compatibility. If we have a test, we are making sure that this change is made intentionally, and not accidentally. Could we find a way to be able to test this?

3) 

This test: `test_nan_behaviour_ignore`, if I understand correctly tests that we have the same number of nan in input and output. Ideally, we should test that the same observation was nan as input and nan as output. If transformer replaces this nan by something else, and introduces another nan, this test could still pass I think.

I do agree that this test might be a bit over the top. But would it be too hard to implement?

solegalli:
1) I updated pandas and the local error I was getting is now solved.
2) I tried to think how to go about this one, but I could not come up with a good solution, so we can let this one go.
3) I think this one is not too important.

I am not sure what your thoughts are, but I think this PR could be good to go after you merge.

Let me know what you think

Cheers

GLevV:
> 1. I updated pandas and the local error I was getting is now solved.
> 
>     2. I tried to think how to go about this one, but I could not come up with a good solution, so we can let this one go.
> 
>     3. I think this one is not too important.
> 
> 
> I am not sure what your thoughts are, but I think this PR could be good to go after you merge.
> 
> Let me know what you think
> 
> Cheers

I think for the first one there is an easy solution for SimilarityEncoder - cast column to str before value_counts and fillna. I was doing it in previous iterations but decided to go with other approach. I can revert it, if it's needed.

PR #539, opened 2022-10-12
fix inverse_transform in encoders

### LINKED ISSUE #428: `OrdinalEncoder` could output -1 for unseen categories

### LINKED ISSUE #538: Inverse transform functionality of encoders
While looking at base encoder class I saw inverse_transform function. It won't work with the most of the encoders (or maybe even all of them), and will only cause problems.

Maybe it should be dropped completely or replaced by dummy function raising NotImplementedError?

solegalli:
In addition:

- add functionality to encode unseen in ordinal encoder
- change logic of encoding unseen categories
- increase code coverage for encoders
- remove CategoricalMixinExpanded
- remove dictionary check in fit
- add deprecation warning in PRatioEncoder

closes #428 
closes #538


solegalli:
FYI @GLevV

PR #544, opened 2022-10-20
dataset module

datacubeR:
Hi @solegalli.
This is my first try on the `load_titanic` function. I just created a function trying to emulate sklearn datasets API.

* I created a `return_X_y_frame`: Scikit-learn allows to return Dataframes or Numpy arrays, but I don´t think that works for feature-engine. I think it should always return a DataFrame/Series object, but one whole Dataset if `False` or X and y if `True`.
* I created a `raw=False`: raw means as is... With nulls as `?`. If True it returns proper `np.nan`.
* `predictors_only=False`: I'm not quite convinced about the name of this parameter. If `False` it returns all the features. If `True` it removes features that normally do not help for predictions such as: `name`,`ticket`, `home.dest`, `boat`, `body` and `cabin`. Normally these features have a high number of nulls, except for `name`.

Also the name of the python file is data.py. Not sure if it's ok, or probably another name should be given.

I created also initial set of tests to check these functionalities. Please provide some feedback especially in the naming convention.

Looking forward to your response,

Alfonso

### DIFF
--- a/feature_engine/datasets/data.py
+++ b/feature_engine/datasets/data.py
@@ -1,0 +1,22 @@
+import pandas as pd
+import numpy as np
+
+
+def load_titanic(return_X_y_frame=False, predictors_only=False, raw=False):
+    df = pd.read_csv("https://www.openml.org/data/get_csv/16826755/phpMYEkMl")
+
+    if predictors_only:
+        df.drop(
+            columns=["name", "ticket", "home.dest", "boat", "body", "cabin"],
+            inplace=True,
+        )
+
+    if not raw:
+        df = df.replace("?", np.nan)
+
+    if return_X_y_frame:
+        X = df.drop(columns="survived")
+        y = df["survived"]
+        return X, y
+    else:
+        return df
--- a/feature_engine/datasets/data.py
+++ b/feature_engine/datasets/data.py
@@ -1,9 +1,12 @@
 import pandas as pd
 import numpy as np
+from sklearn.pipeline import Pipeline
+from feature_engine.imputation import CategoricalImputer, MeanMedianImputer
 
 
-def load_titanic(return_X_y_frame=False, predictors_only=False, raw=False):
+def load_titanic(return_X_y_frame=False, predictors_only=False, handle_missing=False):
     df = pd.read_csv("https://www.openml.org/data/get_csv/16826755/phpMYEkMl")
+    df = df.replace("?", np.nan)
 
     if predictors_only:
         df.drop(
@@ -11,8 +14,18 @@ def load_titanic(return_X_y_frame=False, predictors_only=False, raw=False):
             inplace=True,
         )
 
-    if not raw:
-        df = df.replace("?", np.nan)
+    if handle_missing:
+        pipeline = Pipeline(
+            steps=[
+                (
+                    "categorical_imputer",
+                    CategoricalImputer(imputation_method="missing"),
+                ),
+                ("mean_median_imputer", MeanMedianImputer(imputation_method="mean")),
+            ]
+        )
+
+        df = pipeline.fit_transform(df)
 
     if return_X_y_frame:
         X = df.drop(columns="survived")
--- a/docs/api_doc/datasets/index.rst
+++ b/docs/api_doc/datasets/index.rst
@@ -1,0 +1,15 @@
+.. -*- mode: rst -*-
+
+Datasets
+========
+
+We are starting to build a library of functions that allow you and us to quickly load
+datasets to demonstrate and test the functionality of Feature-engine (and, why not,
+other Python libraries).
+
+At the moment, we support the following functions:
+
+.. toctree::
+   :maxdepth: 1
+
+   titanic
--- a/docs/api_doc/datasets/titanic.rst
+++ b/docs/api_doc/datasets/titanic.rst
@@ -1,0 +1,6 @@
+﻿load__titanic
+=============
+
+.. currentmodule:: feature_engine.datasets
+
+.. autofunction:: load_titanic
--- a/docs/api_doc/index.rst
+++ b/docs/api_doc/index.rst
@@ -42,7 +42,6 @@ Time series
 
    timeseries/index
 
-
 Other
 -----
 .. toctree::
@@ -51,6 +50,12 @@ Other
    preprocessing/index
    wrappers/index
 
+Datasets
+--------
+.. toctree::
+   :maxdepth: 1
+
+   datasets/index
 
 Tools
 -----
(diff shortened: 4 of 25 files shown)


feature_engine/datasets/data.py
```diff
@@ -0,0 +1,22 @@
+import pandas as pd
+import numpy as np
+
+
+def load_titanic(return_X_y_frame=False, predictors_only=False, raw=False):
+    df = pd.read_csv("https://www.openml.org/data/get_csv/16826755/phpMYEkMl")
+
+    if predictors_only:
+        df.drop(
+            columns=["name", "ticket", "home.dest", "boat", "body", "cabin"],
```
solegalli, line 10: This looks really good. 

I'd remove cabin from this list, because cabin is actually a good predictor, and try to add the functionality to extract the first cabin and (perhaps optional) extract the first letter from the cabin, because this is the one that indicates class, which correlates with survival.

feature_engine/datasets/data.py
```diff
@@ -0,0 +1,22 @@
+import pandas as pd
+import numpy as np
+
+
+def load_titanic(return_X_y_frame=False, predictors_only=False, raw=False):
+    df = pd.read_csv("https://www.openml.org/data/get_csv/16826755/phpMYEkMl")
+
+    if predictors_only:
```
solegalli, line 8: I read somewhere, that it is better to specifically say if predictors_only is True, instead of if it just exists.

feature_engine/datasets/data.py
```diff
@@ -0,0 +1,22 @@
...
+        df.drop(
+            columns=["name", "ticket", "home.dest", "boat", "body", "cabin"],
+            inplace=True,
+        )
+
+    if not raw:
```
solegalli, line 14: maybe we could call this param `handle_missing`, and if True, impute them (by mean in numerical and mode in  categorical), and if False, return them as np.nan. I don't think we need the "?" at all, or do we?

solegalli (commented):
Hi @datacubeR 

Great implementation, thank you!

I added a few suggestions for minor modifications. Let me know if you agree with them.

Thank you!

feature_engine/datasets/data.py
datacubeR: @solegalli 
Well I inspired in Sklearn's code convention. For instance this is `fit_intercept` for `LogisticRegression` and they do it this way:

https://github.com/scikit-learn/scikit-learn/blob/36958fb240fbe435673a9e3c52e769f01f36bec0/sklearn/linear_model/_logistic.py#L473

In case you think it needs to be changed,  it should go as `if predictors_only == True:` or `if predictors_only == True:`

feature_engine/datasets/data.py
datacubeR: I like it...

feature_engine/datasets/data.py
datacubeR: Actually I removed `cabin` because 1014 out of 1309 are null values. So in case of implementing what you say... how should I impute this variable with this amount of missing values? Probably a missing indicator with CategoricalImputer, because mode doesn't seem to be a good idea...

feature_engine/datasets/data.py
datacubeR: @solegalli 
In order to implement imputation here, I used `CategoricalImputer` with `encoding_method="frequent"` to impute mode. But when doing this I get the error that `body` has multiple frequent values. So my question is, this is an expected behavior for `CategoricalImputer`? How are the ties addressed in this Imputation?

datacubeR: @solegalli I did a different thing. Since imputation can be applied to `predictors_only` True or False, I thought it could be a better idea to impute categories with a missing indicator. The reason of this, is since not predictor variables have a high number of missing values and imputing mode it could be not quite correct. What do you think?

feature_engine/datasets/data.py
solegalli: makes sense, thank you!

feature_engine/datasets/data.py
solegalli: I think you did a great job here. I wouldn't change it. Thank you for picking this up.

feature_engine/datasets/data.py
solegalli: makes total sense. thank you

solegalli:
Hi @datacubeR 

This looks good to go to me. I'd change the file name from `data.py` to `titanic.py`, because we'll probably add more data loading files.

Then, the only bit remaining would be to go over all the user_guide files, check which ones use the titanic dataset, and change the loading function. Would you be up for it?

Thank you!

datacubeR:
@solegalli I made the requested changes. On fixing the User guide, they are a lot of changes. After checking some examples like this:

image

There are several references to features like `cabin`, that we ended up removing from predictors. Is it OK to fix the examples in this same PR? or should I raise different ones for every module, after merging this one?

solegalli:
Hi @datacubeR 

Great thank you! I think it would be best to make all changes related to the titanic dataset in this PR :)

datacubeR:
Hi @solegalli,

I'll be adding updates for this PR during this week. So far I just modified Categorical encoders in the user guide. As soon as I have more updates I'll let you know...

solegalli:
Awesome, thank you!

solegalli:
Hey @datacubeR 

How are you doing? I've read you are doing well over Linkedin. I am happy to hear that.

Could you give me a shout when this PR is ready? So that I don't miss it like I did with the selection module?

Any chance we could have this ready by the end of the week? I'd like to release a new version of feature-engine, and want to check if we should wait for this PR or we leave it for the next release.

Thank you!

datacubeR:
Hi @solegalli 
Just wanted to let you know that it seems I almost finished the PR.

I have a little issue with the test_docs tests. Not sure why Python code is failing. I have checked the same line the test is pointing as failing and it works. Would you help me with that?

Regarding the dataset module. I added one more test and made some modifications to ensure dtypes for some variables. Working on the examples I noticed sometimes incorrect data types were inferred so I preferred to force it to the correct one. 

Please tell me how it looks to you and let me know if I need to make any modifications, specially for the failing test.

Kind Regards,

Alfonso

solegalli:
Hey @datacubeR 

Thank you so much for the incredible amount of work on this module.

I made a PR to your repo with further changes: https://github.com/datacubeR/feature_engine/pull/2

Could you have a look and merge so we can close this PR?

Thanks a lot!

datacubeR:
@solegalli 
Really nice changes.
I should have come up with something like this seeing a lot of the cabin letter was being used.

Thanks for the feedback and hope this module could be useful for the release.

Thanks, Alfonso

PR #549, opened 2022-10-26
Add fill_value param to forecasting classes to handle NA as per pandas

### LINKED ISSUE #536: add parameter to handle NA as per pandas functionality in forecasting classes
pandas shift and pandas rolling have a parameter to handle NaN introduced during the operations, we should allow users to tap into that parameter in our transformers.

@KishManani one for you? :p

Should be an easy fix.

Morgan-Sell:
Closes #536 

Selected Notes from #536:

pandas shift and pandas rolling have a parameter to handle NaN introduced during the operations, we should allow users to tap into that parameter in our transformers.

We need to add a parameter in the init of the LagFeatures, WindowFeatures and ExpandedWindowFeatures that then would be passed onto the fill_value parameter of pandas.shift (https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.shift.html) to replace the nan introduced by that value.

### DIFF
--- a/feature_engine/timeseries/forecasting/base_forecast_transformers.py
+++ b/feature_engine/timeseries/forecasting/base_forecast_transformers.py
@@ -14,6 +14,7 @@ from feature_engine._docstrings.init_parameters import (
     _drop_original_docstring,
     _missing_values_docstring,
     _fill_value_docstring,
+
 )
 from feature_engine._docstrings.methods import _fit_not_learn_docstring
 from feature_engine._docstrings.substitute import Substitution
--- a/feature_engine/timeseries/forecasting/base_forecast_transformers.py
+++ b/feature_engine/timeseries/forecasting/base_forecast_transformers.py
@@ -50,6 +50,7 @@ class BaseForecastTransformer(BaseEstimator, TransformerMixin, GetFeatureNamesOu
 
     {drop_original}
 
+
     Attributes
     ----------
     {feature_names_in_}
@@ -63,6 +64,7 @@ class BaseForecastTransformer(BaseEstimator, TransformerMixin, GetFeatureNamesOu
         variables: Union[None, int, str, List[Union[str, int]]] = None,
         missing_values: str = "raise",
         drop_original: bool = False,
+        fill_value: Union[float, int, str] = None,
     ) -> None:
 
         if missing_values not in ["raise", "ignore"]:
@@ -77,9 +79,16 @@ class BaseForecastTransformer(BaseEstimator, TransformerMixin, GetFeatureNamesOu
                 f"Got {drop_original} instead."
             )
 
+        if not isinstance(fill_value, (float, int, str)):
+            raise ValueError(
+                "fill_value takes only float, integer, or string. "
+                f"Got {fill_value} instead."
+            )
+
         self.variables = _check_init_parameter_variables(variables)
         self.missing_values = missing_values
         self.drop_original = drop_original
+        self.fill_value = fill_value
 
     def _check_index(self, X: pd.DataFrame):
         """
--- a/feature_engine/timeseries/forecasting/base_forecast_transformers.py
+++ b/feature_engine/timeseries/forecasting/base_forecast_transformers.py
@@ -68,10 +68,16 @@ class BaseForecastTransformer(BaseEstimator, TransformerMixin, GetFeatureNamesOu
     def __init__(
         self,
         variables: Union[None, int, str, List[Union[str, int]]] = None,
-        fill_value: Hashable = lib.no_default,
+        fill_value: Union[None, int, float, str] = None,
         missing_values: str = "raise",
         drop_original: bool = False,
     ) -> None:
+
+        if fill_value is not None and not isinstance(fill_value, (int, float, str)):
+            raise ValueError(
+                "fill_value must be None type, integer, float or string. "
+                f"Got {fill_value} instead."
+            )
 
         if missing_values not in ["raise", "ignore"]:
             raise ValueError(
--- a/feature_engine/timeseries/forecasting/base_forecast_transformers.py
+++ b/feature_engine/timeseries/forecasting/base_forecast_transformers.py
@@ -1,6 +1,7 @@
-from typing import List, Optional, Union
+from typing import Hashable, List, Optional, Union
 
 import pandas as pd
+from pandas._libs import lib
 from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.utils.validation import check_is_fitted
 
@@ -66,16 +67,10 @@ class BaseForecastTransformer(BaseEstimator, TransformerMixin, GetFeatureNamesOu
     def __init__(
         self,
         variables: Union[None, int, str, List[Union[str, int]]] = None,
-        fill_value: Union[float, int, str] = None,
+        fill_value: Hashable = lib.no_default,
         missing_values: str = "raise",
         drop_original: bool = False,
     ) -> None:
-
-        if not isinstance(fill_value, (float, int, str)):
-            raise ValueError(
-                "fill_value takes only float, integer, or string. "
-                f"Got {fill_value} instead."
-            )
 
         if missing_values not in ["raise", "ignore"]:
             raise ValueError(
--- a/feature_engine/timeseries/forecasting/base_forecast_transformers.py
+++ b/feature_engine/timeseries/forecasting/base_forecast_transformers.py
@@ -12,6 +12,7 @@ from feature_engine._docstrings.fit_attributes import (
 from feature_engine._docstrings.init_parameters import (
     _drop_original_docstring,
     _missing_values_docstring,
+    _fill_values_docstring,
 )
 from feature_engine._docstrings.methods import _fit_not_learn_docstring
 from feature_engine._docstrings.substitute import Substitution
@@ -31,6 +32,7 @@ from feature_engine.tags import _return_tags
 
 
 @Substitution(
+    fill_value=_fill_values_docstring,
     missing_values=_missing_values_docstring,
     drop_original=_drop_original_docstring,
     feature_names_in_=_feature_names_in_docstring,
@@ -45,6 +47,8 @@ class BaseForecastTransformer(BaseEstimator, TransformerMixin, GetFeatureNamesOu
     ----------
     variables: str, int, or list of strings or integers, default=None.
         The variables to use to create the new features.
+
+    {fill_value}
 
     {missing_values}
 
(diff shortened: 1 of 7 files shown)


Morgan-Sell:
Hola @solegalli,

How should we handle the `fill_value` default value? The pandas shift performs some nifty autodetection that would be nice to include. From pandas docs:

_The scalar value to use for newly introduced missing values. the default depends on the dtype of self. For numeric data, np.nan is used. For datetime, timedelta, or period data, etc. NaT is used. For extension dtypes, self.dtype.na_value is used._

I coded the ExpandWindowFeatures in such a fashion to keep the advanced functionality. What do you think?

feature_engine/timeseries/forecasting/base_forecast_transformers.py
```diff
@@ -61,10 +66,17 @@ class BaseForecastTransformer(BaseEstimator, TransformerMixin, GetFeatureNamesOu
     def __init__(
         self,
         variables: Union[None, int, str, List[Union[str, int]]] = None,
+        fill_value: Union[float, int, str] = None,
```
solegalli, line 69: I think the default value should be that of pandas shift:

https://github.com/pandas-dev/pandas/blob/91111fd99898d9dcaa6bf6bedb662db4108da6e6/pandas/core/frame.py#L5766

feature_engine/timeseries/forecasting/base_forecast_transformers.py
```diff
@@ -61,10 +66,17 @@ class BaseForecastTransformer(BaseEstimator, TransformerMixin, GetFeatureNamesOu
...
         missing_values: str = "raise",
         drop_original: bool = False,
     ) -> None:
 
+        if not isinstance(fill_value, (float, int, str)):
```
solegalli, line 74: This is a param for pandas shift, so I would leave the checks to pandas. I would remove this from here. Not sure how many options this parameter can take, and also they may expand it or reduce it in the future. So no checks makes maintaining easier.

feature_engine/timeseries/forecasting/expanding_window_features.py
```diff
@@ -170,12 +179,27 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
         X = super().transform(X)
 
-        tmp = (
-            X[self.variables_]
-            .expanding(min_periods=self.min_periods)
-            .agg(self.functions)
-            .shift(periods=self.periods, freq=self.freq)
-        )
+        # use pandas shift default autodetect
+        if self.fill_value is None:
```
solegalli, line 183: I would just pass fill_value to shift without the if clause. Pandas shift has their own if clauses to work with whatever the user passes to this param

solegalli (commented):
Hi @Morgan-Sell 

thanks for the start on this PR.

I think we should do whatever pandas shift does with this param. And also "outsource" to them all the checks. 

I added a link in my comments below with the defaiut value of this param.

I think the changes here are really small. The strings as you did already, add the param to the init which you did already, and then, just pass self.fill_value to pandas shift within the transformers.

And a couple of tests passing integers and floats to fill_value to ensure that the nan are filled.

How does this sound?

Morgan-Sell:
@solegalli,

Thanks for the explanation! I didn't know about `lib.no_default` that's why I was overcomplicating things.

I've implemented the new code into the 4 forecasting classes. The next step is to create the unit tests.

solegalli (commented):
Hi @Morgan-Sell 

the code so far looks good to me, but there must be a small bug somewhere, because the classes can not be imported. 

Would you be able to sort this out?

Cheers
Sole

Morgan-Sell:
@solegalli,

I think the programming gods have tricked us again! Editing the WindowFeature and ExpandingWindowFeature classes are more challenging than we envisioned.

We both thought that by adding the `fill_value` param to the `shift()` that all `NaN` would be replaced. This is not the case for WindowFeatures and ExpandingWindowFeatures ;)

Some of the `NaN` is generated due to an insufficient quantity of data. E.g., if the user passes `5` for the `window` param then there will be `nan` when there are less than 5 time periods. I'm guessing that we also want these `nan` to be filled.

Will you please confirm? 

In the meantime, I'll work on the LagFeatures.

Morgan-Sell:
@solegalli,

While working on the unit test for LagFeatures(), I had a thought. Should we limit LagFeatures to numerical variables? In the end, LagFeatures only shifts the data. Is there a scenario in which the user would apply this action on a categorical variable?

The following code in the BaseForecastTransformer class is preventing the shift of categorical variables:

```
 # find or check for numerical variables
```

solegalli:
> 

Good catch @Morgan-Sell! I forgot about this.

I wonder if we should just use pandas.fillna(fill_value) to replace the nan introduced by pandas rolling as well.

@KishManani what is your view on this?

The idea was to expose the parameter fill_value from pandas.shift(). Which is fairly straightforward.

But pandas.rolling() will also introduce nan (in min_periods is bigger than 1), and this will remain nan. 

The question is, shall we leave the nan? shall we replace the nan with the same fill_value we pass to pandas.shift()?

solegalli:
> @solegalli,
> 
> While working on the unit test for LagFeatures(), I had a thought. Should we limit LagFeatures to numerical variables? In the end, LagFeatures only shifts the data. Is there a scenario in which the user would apply this action on a categorical variable?
> 
> The following code in the BaseForecastTransformer class is preventing the shift of categorical variables:
> 
> ```
>  # find or check for numerical variables
>  self.variables_ = _find_or_check_numerical_variables(X, self.variables)
> ```

Atm we  lag only numerical variables. In forecasting it would be rare to lag categorical variables as well. Let's keep it as it is, and we see what users say.

Morgan-Sell:
@solegalli @KishManani ,

I like the idea of using `fillna()` after the dataframe is created and ignoring the `fill_value` param in pandas `shift()`. There will only be one action that fills all of the missing values.

Morgan-Sell:
@solegalli,

Do you have any thoughts? I think creating a _fill_na() method for the parent forecast class is the best approach. I would love to knock this PR out. I could use the constructive distraction ;)

solegalli:
Hi @Morgan-Sell 

I'd say, for the LagFeatures class, we use the imputation method offered in pandas shift.

And for the Window classes, we use fillna.

Go ahead with the changes :)

If you find it easier, we can fix LagFeatures in one PR, and Window and Expanding in a different one, since we will be doing different things.

Morgan-Sell:
@solegalli,

If a user assumes a value for the `freq` param for any of the 3 forecasting, then the pandas shift imputation will not handle the NaN.

What are your thoughts on having a fillna after the following code in the transform() method?

`X = X.merge(tmp, left_index=True, right_index=True, how="left")`

The above code is likely to generate NaN. Also, the merge is performed after `pd.shift()`.

solegalli:
Hard to say without seeing it. Could we have some starter "functional" code to base the discussions on?

Maybe let's just focus on getting lagFeatures to work with NA first? At the moment, all tests are failing, so hard to say what is not working as intended.

Morgan-Sell:
@solegalli,

Above is the code from the LagFeatures class. LagFeatures is passing the unit tests on test_lag_features.py.

I suggest we implement this code for the other 2 classes. What do you think?

solegalli:
> @solegalli,
> 
> Above is the code from the LagFeatures class. LagFeatures is passing the unit tests on test_lag_features.py.
> 
> I suggest we implement this code for the other 2 classes. What do you think?

I think this might generate unexpected behaviour. Why not not just add the fill_value in pandas shift?

Morgan-Sell:
@solegalli,

pandas `fill_value` will not fill the NaN that may be generated when merging dataframes.  See the relevant code below.

`X = X.merge(tmp, left_index=True, right_index=True, how="left")`

At first, I used `fill_value` in pandas shift. However, there were will dataframe cells that had NaN.

Here's an example.

- Dataframe `X` has an index that's 1-Jan-22, 2-Jan-22, 3-Jan-22, 4-Jan-22, and 5-Jan-22. 
- Shift `X` by 2 days using `X.shift(freq="2d", axis=0)`to create `X_tr`
- `X_tr` has an index that's 3-Jan-22, 4-Jan-22, 5-Jan-22, 6-Jan-22, and 7-Jan-22.
- Merging `X` and `X_tr` by their indices will generate `NaN` values for `X_tr` on 1-Jan-22 and 2-Jan-22.

If we would like to fill these null values, I suggest that the program uses `fillna()` after the merge.

Hopefully, I explained myself somewhat coherently. I'm still working on that skill ;)

solegalli:
Crystal clear @Morgan-Sell 

Thanks a lot for taking the time. 

Ok then, we go forth with your solution!

solegalli:
In which case, we won't need the hashable thing from pandas, right? We can get rid of all of that, lol. 

Sorry @Morgan-Sell I know it was my suggestion :s

Morgan-Sell:
No pasa nada, @solegalli!

What's the "hashable thing"? Does pandas use a hash table? I'm lost ;)

Morgan-Sell:
hi @solegalli,

Could you provide a simple example of using the WindowsFeatures() with the `periods` and `freq`?

I'm starting to think that the shift performed by periods is `n - 1` instead of `n`. And, using the energy generation data from our unit tests, the WindowFeatures() is 15 minutes ahead of when it should be.

Sorry to be causing problems ;)

solegalli:
Hi @Morgan-Sell 

I am thinking again about the way we are filling na in this transformers. If we just use fillna() from pandas, all na will be replaced by the fillvalue, the ones introduced by the transformers, and all other nan already existing in the dataset. And this is not what we want. We just want to fill the nan introduced by the creation of the new features. 

So I am having second thoughts about this procedure.

I need to sleep on it a bit longer. What's your view?

Morgan-Sell:
@solegalli,

If the goal is to only fill the new features, then your approach makes sense. Will you provide a scenario in which we will only filly the NaNs of the new features? How come we don't fill the NaN of the existing features, if NaNs are present?

solegalli:
If I remember correctly, we did not enforce the transformers to work on dataframes without nan.

We left it to the user to choose/decide if they want to impute data before creating the lags or if they want to create the lags even with missing data. 

I think there is so little best practice around creating features for forecasting, that I would not be able to say which approach is better, if enforcing imputation or not.

If you have any experience on it, I will appreciate some links/tips. Alternatively, I think I'd have to make more forecasts, to become more experienced and then be able to suggest one way or another :/

Morgan-Sell:
@solegalli,

I don't have enough experience to opine one way or the other. Should we put this PR on hold? I'll find another task ;)

solegalli:
I think so :_(

PR #547, opened 2022-10-25
Adding examples for Encoding Module 

datacubeR:
Hi @solegalli 

I just added examples for the encoding module. 
* I would like your feedback on this especially because I'm not sure how to test this examples are ok. Is there any way to render a dev documentation to see it?
* Second, scikit-learn shows some kind of glimpse of the answer, but they use `np.arrays` so it's a bit more concise than showing a glimpse of a dataframe, what do you think about that?

Please let me know if we are OK with this before start working on other modules. 

Finally, do you think it could be a good a idea to have some notebooks for every module with this examples?

Thanks, 

Alfonso

PR #550, opened 2022-10-28
[ENH] Add keywords parameter to StringSimilarityEncoder

glevv:
Added keywords support for StringSimilarityEncoder.

Keywords allow user to introduce some domain knowledge (if any) into encoder


glevv:
I added a bunch of tests, seems good, but maybe it needs more

feature_engine/encoding/similarity_encoder.py
```diff
@@ -102,14 +102,22 @@ class StringSimilarityEncoder(CategoricalInitMixin, CategoricalMethodsMixin):
...
         replace missing values with an empty string, '', and then return the similarity
         measures.
 
+    keywords: dict, default=None
+        User defined dictionary of keywords, dict(feature: [keyword1, keyword2, ...]).
+        Instead of finding top_k categories in features, encoder will use this keywords
+        to create similarity variables. Useful when someone has domain knowledge of the
+        problem. Could be defined only for several features; in this case for specified
+        features keywords will be used and most common categories will be used for
+        unspecified.
```
solegalli, line 119: I don't understand the last sentence of this description. Would you be able to re-phrase it?

feature_engine/encoding/similarity_encoder.py
```diff
@@ -198,20 +217,19 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
...
-                self.encoder_dict_[var] = (
-                    X[var]
-                    .astype(str)
-                    .value_counts()
-                    .head(self.top_categories)
-                    .index.tolist()
-                )
-        elif self.missing_values == "impute":
-            for var in self.variables_:
+        for var in self.variables_:
```
solegalli, line 229: are we removing the functionality to handle missing data? I guess, not because the tests pass. 

But regarding this code:
https://github.com/feature-engine/feature_engine/blob/a2c315db76cd6cd1327b1c7247cd797a63cec768/feature_engine/encoding/similarity_encoder.py#L236

Are we not making nan part of the categories to appear as dummies?

tests/test_encoding/test_similarity_encoder.py
```diff
@@ -213,3 +213,162 @@ def test_get_feature_names_out_na(df_enc_big_na):
     }
     assert tr.get_feature_names_out(input_features=None) == out
     assert tr.get_feature_names_out(input_features=input_features) == out
+
+
+def test_keywords_bad_type():
```
solegalli, line 218: could we test more than just a string by using parametrize? maybe tuples, integers?

tests/test_encoding/test_similarity_encoder.py
```diff
@@ -213,3 +213,162 @@ def test_get_feature_names_out_na(df_enc_big_na):
...
+    with pytest.raises(ValueError):
+        StringSimilarityEncoder(keywords="hola")
+
+
+def test_keywords_bad_items():
```
solegalli, line 223: same here, could we use parametrize and test more options?

tests/test_encoding/test_similarity_encoder.py
```diff
@@ -213,3 +213,162 @@ def test_get_feature_names_out_na(df_enc_big_na):
...
+    with pytest.raises(ValueError):
+        StringSimilarityEncoder(keywords={"column": "hola"})
+
+
+def test_keywords_dont_match(df_enc_big):
```
solegalli, line 228: Could we either add columns that are in the data to the dictionary, or even better, test a second dictionary with a combination of columns that are in the data, and a column that is not?

Also, here we are testing that the variables/columns don't match, correct? can we adjust the name of the test?

tests/test_encoding/test_similarity_encoder.py
```diff
@@ -213,3 +213,162 @@ def test_get_feature_names_out_na(df_enc_big_na):
...
+        "var_C": ["C", "D"],
+    }
+    # test transform output
+    for col in transf.keys():
+        assert X[col].sum() == transf[col]
+    assert "var_B" not in X.columns
+    assert "var_B_F" not in X.columns
+
+
+def test_encode_full_keywords():
```
solegalli, line 290: what do you mean by "full_keywords"?

solegalli (commented):
Hi @glevv 

Thank you for the enhancement. It looks good to me. I have mostly some minor comments.

And one question regarding how Nan in the train set should be handled. My understanding is:

- "impute", replaces nan by "". So when creating the dummies, would "" be one of them?
- "raise"; raises an error
- "ignore": nan remain nan. But the current logic, won't make "" part of the dummies?

Thanks a lot for the contribution!

solegalli:
Hi @glevv 

This is looking really good. Thank you so much for the quick turnaround.

I've got a couple of questions about the logic in fit:

**fit logic:**

- `self.encoder_dict_` is a dictionary of variables as keys and a list of dummies (or keywords in this case) as values.
- `self.keywords` is the same thing, if I understand correctly: variables as keys and list of keywords as dummies.

So if the user defined all categorical variables, or the variables in the parameter `variables` in the `self.keywords` dict, in essence, `self.encoder_dict_` and `self.keyword` dict are the same. In which case, we could just copy one to the other? instead of looping?
Would that not be faster?

At the moment, we've got the keywords dictionary after all the logic in fit():
https://github.com/feature-engine/feature_engine/blob/05ae18c5a6d01601de209116fe908acf1ee88635/feature_engine/encoding/similarity_encoder.py#L256-L258

This means that for the variables in the dictionary, we are fitting twice.

I wonder if there is a way in which we could avoid running the logic twice for those variables that are in the keywords dictionary?

Something like, if self.keyword exists and all the keys are in variables_ then just copy, alternatively, define the variables not in self.keywords.

**imputation logic**

Could we add a few tests to corroborate that:

- when `handle_missing="impute"` and the fit() data has nan,, then "" is part of the dummies? is this the expected behaviour?
- when `handle_missing="ignore"` and the fit() data has nan,, then no dummy is generated at the back of it? 

In short, we need a test to differentiate the behaviour between handle_missing is impute or ignore, and the train set has nan values.

Finally, we've got another PR coming up that modifies the strings of this class: #547 

Thanks a lot for your time!

glevv:
Done

feature_engine/encoding/similarity_encoder.py
```diff
@@ -174,9 +182,19 @@ def __init__(
                 "missing_values should be one of 'raise', 'impute' or 'ignore'."
                 f" Got {missing_values!r} instead."
             )
+        if keywords and not isinstance(keywords, dict):
+            raise ValueError(
+                f"keywords should be dict or None. Got {keywords!r} instead."
+            )
+        if keywords and not all(isinstance(item, list) for item in keywords.values()):
```
solegalli, line 189: This may be a bit picky, but, should we not test that all items in the lists are strings as well?

feature_engine/encoding/similarity_encoder.py
```diff
@@ -226,9 +249,12 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
                     X[var]
                     .astype(str)
                     .value_counts(dropna=True)
+                    .drop("nan", errors="ignore")
                     .head(self.top_categories)
                     .index.tolist()
                 )
+        if self.keywords:
```
solegalli, line 256: If we update the dictionary here, and the dictionary keys contain all the variables in `variables_`, we are running all of the logic between 226 and this line, for no reason. Because those keywords will not be used. 

Would it not be better to check, at around line 223 that:

- if, all of `variables_` are in `keywords.keys()`, then just copy dict. 
- If some of `variables_` are in keywords.keys(), then apply the logic to find keywords in the remaining variables
- If keywords is None, then apply the logic in all `variables_`

What do you think about this?

tests/test_encoding/test_similarity_encoder.py
```diff
@@ -213,3 +223,165 @@ def test_get_feature_names_out_na(df_enc_big_na):
     }
     assert tr.get_feature_names_out(input_features=None) == out
     assert tr.get_feature_names_out(input_features=input_features) == out
+
+
+@pytest.mark.parametrize("keywords", ["hello", 0.5, [1]])
```
solegalli, line 228: If keywords is a dictionary where the list contains items other than strings, it should also fail. Should we add this to the parametrize list?

tests/test_encoding/test_similarity_encoder.py
```diff
@@ -213,3 +223,165 @@ def test_get_feature_names_out_na(df_enc_big_na):
...
+def test_keywords_bad_type(keywords):
+    with pytest.raises(ValueError):
+        StringSimilarityEncoder(keywords=keywords)
+
+
+@pytest.mark.parametrize("item", ["hello", 0.5, 1])
```
solegalli, line 234: or maybe it fits better here?

solegalli (commented):
Hi @glevv 

Thank you for the code changes! I think this is almost there.

Would you mind rebasing to incorporate #547 and also have a look at my comment below?

Thanks a lot!

feature_engine/encoding/similarity_encoder.py
glevv: It does not matter, they will all be converted to strings later in `gpm` function

feature_engine/encoding/similarity_encoder.py
glevv: No, it will be worse. It will make logic more complicated (which in turn will mess up coverage and tests) and won't give any speed up. 

This way it is straightforward and readable.

tests/test_encoding/test_similarity_encoder.py
glevv: It won't fail, since we will convert any object to it string representation before calculating string similarity

glevv:
I implemented this logic, but tests are failing because some other transformers are not using Optional typing

see #555

solegalli:
Sorry for the delay, I just merged #555 

Logic changes look good to me!

solegalli:
Hi @glevv 

FYI: https://github.com/glevv/feature_engine/pull/4

and we are then good to go.

Thanks for the contribution!!!

PR #554, opened 2022-11-08
Update fold default value for outlier class

px39n:
**This change is not necessary, but make this function more friendly.**

Changed default value of fold to 0.05 when the capping_method='quantile', for other methods, keep the default as 3.

Past :
```
OutlierTrimmer(capping_method='quantile')
OutlierTrimmer.fit(X)
will raise error due to default fold is 3.
```
For now:
```
OutlierTrimmer(capping_method='quantile')
OutlierTrimmer.fit(X)
Will work,
```

### DIFF
--- a/feature_engine/outliers/base_outlier.py
+++ b/feature_engine/outliers/base_outlier.py
@@ -170,12 +170,13 @@ class WinsorizerBase(BaseOutlier):
         distribution. Can take 'left', 'right' or 'both'.
         """.rstrip()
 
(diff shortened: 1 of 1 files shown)


px39n:
Whats wrong with test_type :0

solegalli:
test_type errors have been fixed in #555 

If you rebase main onto your feature branch, it should work now.

I don't see the files with the changes. What happened?

PR #556, opened 2022-11-09
Update baseoutlier default fold value

px39n:
**This change is not necessary, but make this function more friendly.**

Changed default value of fold to 0.05 when the capping_method='quantile', for other methods, keep the default as 3.

Past :
```
OutlierTrimmer(capping_method='quantile')
OutlierTrimmer.fit(X)
will raise error due to default fold is 3.
```
For now:
```
OutlierTrimmer(capping_method='quantile')
OutlierTrimmer.fit(X)
Will work,
```

### DIFF
--- a/feature_engine/outliers/base_outlier.py
+++ b/feature_engine/outliers/base_outlier.py
@@ -149,9 +149,7 @@ class WinsorizerBase(BaseOutlier):
 
     If `capping_method='mad'` fold is the value to multiply the MAD.
 
-    If `capping_method='quantiles'`, fold is the percentile on each tail that should
-    be censored. For example, if fold=0.05, the limits will be the 5th and 95th
-    percentiles. If fold=0.1, the limits will be the 10th and 90th percentiles.
+    If `capping_method='quantiles'`, check unit_test
     """.rstrip()
 
     _capping_method_docstring = """capping_method: str, default='gaussian'


feature_engine/outliers/base_outlier.py
```diff
@@ -149,9 +149,7 @@ class WinsorizerBase(BaseOutlier):
 
     If `capping_method='mad'` fold is the value to multiply the MAD.
 
-    If `capping_method='quantiles'`, fold is the percentile on each tail that should
-    be censored. For example, if fold=0.05, the limits will be the 5th and 95th
-    percentiles. If fold=0.1, the limits will be the 10th and 90th percentiles.
+    If `capping_method='quantiles'`, check unit_test
```
solegalli, line 152: The documentation needs to be very specific and let the user know anything they need. We can't send them to check unit_test.

Could you please rephrase it so that it reflects the current functionality?

feature_engine/outliers/base_outlier.py
```diff
@@ -170,14 +168,15 @@ class WinsorizerBase(BaseOutlier):
         distribution. Can take 'left', 'right' or 'both'.
         """.rstrip()
 
-    _fold_docstring = """fold: int or float, default=3
+    _fold_docstring = """fold: int or float, default=3 or 0.05 (quantile)
```
solegalli, line 171: I would say as follows:

```
fold: int or float, default=0.05 if `capping_method` is `quantile`, or 3 otherwise.
```

feature_engine/outliers/base_outlier.py
```diff
@@ -170,14 +168,15 @@ class WinsorizerBase(BaseOutlier):
...
         The factor used to multiply the std, MAD or IQR to calculate
         the maximum or minimum allowed values.
         Recommended values are 2 or 3 for the gaussian approximation,
         1.5 or 3 for the IQR proximity rule and 3 or 3.5 for MAD rule.
 
-        If `capping_method='quantile'`, then `'fold'` indicates the percentile. So if
-        `fold=0.05`, the limits will be the 95th and 5th percentiles.
+        If `capping_method='quantile'`,the fold default=0.05,
+        The `'fold'` indicates the percentile.
+        eg: if fold=0.05, the limits will be the 95th and 5th percentiles.
```
solegalli, line 180: I don't think we need to rephrase lines 179 and 180. Could we leave the original docstring?

solegalli (commented):
Hi @px39n 

Thanks a lot for the suggestion and the changes in the init.

We need to add a test to corroborate the new functionality, i.e., that when the user sets the method to quantile the default is 0.05, and otherwise it remains 3.

Could you add that please?

Also, could you please check my suggestions below?

feature_engine/outliers/base_outlier.py
px39n: Sorry I added that line to check if the unit_test bug is fixed.
I forgot to change it back.

px39n:
> Hi @px39n
> 
> Thanks a lot for the suggestion and the changes in the init.
> 
> We need to add a test to corroborate the new functionality, i.e., that when the user sets the method to quantile the default is 0.05, and otherwise it remains 3.
> 
> Could you add that please?
> 
> Also, could you please check my suggestions below?

Hi. Thanks for your comment.
I have updated and added two unit_test functions to check
- when the user sets the method to quantile the default is 0.05
- and otherwise it remains 3.

px39n:
Actually, there is a small drawback.
This commit check the default fold value by simply comparing to 3.
if 3 & quantile --> default value --> change fold to 0.05
or otherwise, do nothing. 

I could add a function to determine if user really change the default value by using [Link](https://stackoverflow.com/questions/64427593/how-to-check-if-default-value-for-python-function-argument-is-set-using-inspect). But that will add the complexity.

solegalli:
> 

I'd be keen to see the implementation of this. How much more complexity would it add?

solegalli (commented):
Hey @px39n 

Thanks for the tests and the docstring changes.

I think we could be good to go with the minor tweak in the tests, which are to some degree, just cosmetic.

How much more complexity would it add to do the test for the user entering the parameter 3 manually for quantile? I don't see why a user would select quantile and 3 as fold, but in the unlikely event, the error should say that with quantile the values for fold should be between 0 and 0.2.

If it is just a few lines, then I'd say let's do it. If it adds a lot more complexity and dependencies, then skip it.

Thank you

tests/test_outliers/test_winsorizer.py
```diff
@@ -156,6 +156,20 @@ def test_quantile_capping_right_tail_with_fold_15_percent(df_normal_dist):
...
 
+def test_quantile_fold_default_value(df_normal_dist):
+    # test case 1: Test quantiles, with fold default = 0.05
+    transformer = Winsorizer(capping_method="quantiles")
+    assert transformer.fold == 0.05
+
+
+def test_other_fold_default_value(df_normal_dist):
+    # test case 2: Test gaussian, iqr, mad, with fold default = 3
+    capping_method_list = ["gaussian", "iqr", "mad"]
```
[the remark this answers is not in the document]
px39n: Yes I update the unit test

px39n:
Hi, @solegalli 

Sorry, I can't do that. I tried to write something to detect if I get an maual input 3. But unfortunately this is not feasible within a few lines.

I also tried use another way instead, (in Comment fix6):

1. Set default = None, 
2. the class initialize self.fold according to capping_method.

This works well, however, it cannot pass tests\test_outliers\test_check_estimator_outliers.py.
Because change default in init do not adheres to scikit-learn conventions, here will be a mutated argument.

> In short, `OutlierTrimmer().get_params()`  is not same as init_params type defined in class (which is None).  So this convention check self.fold instead of fold with init_params . So the error will always be raised.


To fix this, we have to change value in fit function, which break the code's legibility

PR #569, opened 2022-11-30
Reorganized badges

glevv:
- Added codestyle badge
- Added number of contributors badge
- "Normalized" badges (they all have links now and the same styling)


glevv:
@solegalli you can check how it looks on my branch, maybe something needs to be taken off or added

solegalli:
LGTM! Thank you @glevv

PR #572, opened 2022-12-07
[WIP] allow encoders to encode variables with nan

### LINKED ISSUE #481: Handling of missing values in categorical encoders
Some transformers such as the `CountFrequency` categorical encoder have options for handling new unseen categories during calls to `transform`. In some cases, unseen categories and explicit missing values might be getting mixed together however.

It would be good to have some distinction between the two and potentially different options for courses of action that would be consistent across transformers. For example, if `CountFrequency` encoder receives an explicit missing value during transform, the user might want to do one of the following:
* Propagate those missing values to the output of `transform`.
* Treat them the same as unseen categories (encoding them with zeros).
* Encode them with the frequency of missing values that were passed to the training data (treating missing values as a separate category).
* Raise an error.

In most of the cases, only the last alternative (raising an error) is supported, but would be ideal to have options and to make explicit in the documentation what happens with missing values.

### LINKED ISSUE #493: Ignore NaNs before using `OrdinalEncoder`
*Is your feature request related to a problem?*
`OrdinalEncoder` should accept nulls. Sometimes you don't want to impute directly but using Imputing Options of XGBoost, LightGBM or CatBoost. Because of the constraints of always Impute before this is currently not possible.

*Describe the solution you'd like*
I like that Feature Engine forces you to Impute first, but I will add some kind of default flag `ignore_nan=False` in case we want to use other imputation afterwards.

Hope you find this helpful.

solegalli:
closes #493 
closes #481

FYI @glevv 

I made a start

- [x] CountFrequencyEncoder
- [x] DecisionTreeEncoder: will not allow NAN
- [x] MeanEncoder
- [x] OrdinalEncoder
- [x] RareLabelEncoder
- [x] WoEEncoder: will not allow NAN
- [x] PRatioEncoder needs to be deprecated (otherwise will through an error)
- [x] StringSimilarityEncoder: needs adaptation of inheritance if init params
- [ ] OneHotEncoder: needs new transform logic
- [x] MatchCategories inherits from the base encoders, so needs update

OneHotEncoder could have 3 options:

- missing_values="raise" ==> current functionality
- missing_values="encode" ==> treat na as an additional category
- missing_values="ignore" ==> ignore, but only if encoding in k-1, otherwise they will be merged with the dropped category

**Tree encoder**:  Change logic so that instead of encoder pipeline it has an encoding dictionary. I could return the output of the tree over the unique values to create a dictionary.


solegalli:
@glevv 

I'd like to pick your brain on these ones:

- **WoEEncoder**: the weight of the evidence is suitable and designed for logistic regression. Thus, in principle, it makes little sense to use this transformation for xgboost, catboost or lightGBM. So I am tempted to leave the transformer unchanged, that is, to error when it sees NaN. That is because, sklearn logistic regression does not work with nan. What's your view?

- **DecisionTreeEncoder**: similarly, the idea of this transformer is to make the variable "monotonic" so that it can be used in linear models. I don't think it adds a lot of value to perform this encoding and then pass it to a decision tree based model. Therefore, I am also inclined to leave the transformer as is, that is, to error when it sees NAN. What's your view?

**Pros of not changing the logic**: makes my life easier. Those transformers would need quite a bit of a logic change
**Pros of changing the logic**: homogeneous API for users, even if they most likely shouldn't use these transformers with xgb and the lot

The thing is, I need to modify the API quite a bit to accommodate these transformers, whatever decision we make, so I'd like to brainstorm a bit before committing to the changes.

@david-cortes and @datacubeR I'd also value your views on these ones, since you requested in the linked issues to allow the encoders to work with NAN values. I'd be keen to know what you think or how you use the encoders.

Thank y'all!

glevv:
I'm always for making life easier, so +1 on leaving DT and WOE encoder without 'allow-nan' option

solegalli:
expanding tests:

- [ ] ohe
- [ ] ordinal
- [ ] rare label
- [ ] similarity
- [ ] woe

Add error messages to tests, fix bugs,  reorganize per method

solegalli:
Tests pass locally. There is something weird with circle ci :_(

solegalli:
FYI 
#575

david-cortes:
Thanks for looking into this.

I think it's fine if the transformers mentioned earlier are left without an option for missing values as the rest already cover the important functionalities.

In theory I _could_ envision a scenario in which someone might want to use the WoE encoder in order to pass something to a multivariate imputer and later to a logistic regression model, but that's probably a far fetched case.

solegalli:
TODO: 

- [ ] Rarelabels: test behavious when nan in fit()

solegalli:
Overall, this is done. 

I'd need to check if the current class inheritance is OK, or if it is better to try and create functions instead.

PR #573, opened 2022-12-08
Create ProbeFeatureSelector class

### LINKED ISSUE #292: feat: Random Noise Selection method
Filtering Noisy features from a set of features can be easily accomplished by adding a one or more random variables to train set, train a gradient boosting or linear model and compute the feature importance.

When a feature is  below the importance of the random noise probably that feature is not so important.

[Reference](https://www.youtube.com/watch?v=RtqtM1UJfZc&t=3150s)

A `NoiseFilterSelection` method that trains a gradient boosting model (sklearn is a required dependency already) and returns only the features with importance above the random variable.

Morgan-Sell:
Closes #292 

Selected notes from #292:

References to selection method:
- [Feature Selection beyond Feature Importance](https://www.kdnuggets.com/2019/10/feature-selection-beyond-feature-importance.html).
- [satra](https://gist.github.com/satra/c6eb113055810f19709fa7c5ebd23de8)
- [Feature Importance by example of a Random Forest](https://towardsdatascience.com/explaining-feature-importance-by-example-of-a-random-forest-d9166011959e)
- [Original paper](https://jmlr.org/papers/volume3/stoppiglia03a/stoppiglia03a.pdf)

Approach:
- Introduce 3 random variables: (a) binary variable (0 or 1); (b) Uniform between 0 and 1; and (c) random integers
- Train the model with all existing variables and new random variables
- Derive the feature importance
- Track the number of times that each variable is not more important than all 3 random variables
- Return a tuple of tuples of all the variables and the number of times that the variable was not more important than all 3 random variables


Morgan-Sell:
@solegalli,

Given that this transformer is atypical since it will not be used in a pipeline, it doesn't require some of the standard attributes, e.g., `variables_top_drop_`, and methods, e.g., `_confirm_variables()` used in other feature selection classes.

Should the parent class be BaseSelector()? Or, should ProbeFeatureSelection() inherit the functionality from BaseEstimator, TransformerMixin, and GetFeatureNamesOutMixin?

feature_engine/selection/probe_features_selector.py
```diff
@@ -0,0 +1,163 @@
...
+        X: pandas dataframe of shape = [n_samples, n_features]
+           The input dataframe
+        y: array-like of shape (n_samples)
+           Target variable. Required to train the estimator.
+        """
+        # check input dataframe
+        X, y = check_X_y(X, y)
+
+        # find numerical and categorical variables
+        self.variables = _find_categorical_and_numerical_variables(X, None)
```
solegalli, line 69: This transformer takes an sklearn estimator, so it can only work with numerical variables

feature_engine/selection/probe_features_selector.py
```diff
@@ -0,0 +1,163 @@
...
+        self.feature_importances_ = feature_importances_cv
+
+        return self
+
+    def transform(self, X: pd.DataFrame) -> pd.DataFrame:
+        """
+        Return a tuple comprised of the variables and the number of times
+        each variable was worse than all three random variables.
+        """
+        pass
```
solegalli, line 123: is this WIP?

feature_engine/selection/probe_features_selector.py
```diff
@@ -0,0 +1,163 @@
...
+        df["rndm_binary_var"] = binary_var
+        df["rndm_uniform_var"] = uniform_var
+        df["rndm_gaussian_var"] = gaussian_var
+
+        return df
+
+
+    def _count_variables_importance_less_than_random_variables(self) -> Dict:
+        """
+        Count how many times a variable is less than the maximum value
```
solegalli, line 145: My understanding is that this transformer will select the variables whose importance greater than the importance of the random features.

Now, we have 3 random features, so we would have 2 importance values. What to do?

From this logic, I understand that you are taking the greatest of the importance values, correct?

Why count? I'd done if var_importance > max_var_random_vars then select otherwise drop.

Now, continuing thinking out loud, the random variable importance, will be random, so a conservative approach could be to take the average instead of the max.

feature_engine/selection/probe_features_selector.py
```diff
@@ -0,0 +1,163 @@
...
+        return self
+
+    def transform(self, X: pd.DataFrame) -> pd.DataFrame:
+        """
+        Return a tuple comprised of the variables and the number of times
+        each variable was worse than all three random variables.
+        """
+        pass
+
+    def _generate_probe_features(self, n_obs: int) -> pd.DataFrame:
```
solegalli, line 125: great method! Thank you.

I'd like to give the user the option to decide if they want 1 feature of each, or 3 features of the same. So if this function could take the parameter "distribution" which values "all", "binary", "uniform" and "normal", to return either 1 of each, or all 3 of the same distribution, that will be great!

solegalli: This function also needs a seed so that the same features are created in case it is ran multiple times.

feature_engine/selection/probe_features_selector.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+    """
+    def __init__(
+        self,
+        estimator,
+        scoring: str = "roc_auc",
+        n_iter: int = 10,
+        seed: int = 0,
+        # TODO: Do we need confirm_variable given that this selector will not be used in a pipeline?
```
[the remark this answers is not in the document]
Morgan-Sell: In issue #292, @glevv said, "This is an ok FS method, i just don't think it should be used as step in pipeline as done in satra and should be more of a model examination tool."

I was following his suggestion.

@glevv, do you have additional comments?

feature_engine/selection/probe_features_selector.py
Morgan-Sell: See my reference to @glevv's comment above. I was assuming that would be for exploratory purposes. I'm happy to construct a complete transformer. That's Team feature-engine's superpower!

feature_engine/selection/probe_features_selector.py
Morgan-Sell: I included the count because I thought this would be an "exploratory class".

At this moment, I cannot recall precisely why, but I had the idea of comparing the variables' feature importances against multiple random variables, hence the `cv` parameter. I guess I wrote the code as such because of pure chance one random variable could be deemed more informative than variable X. We don't want to remove variable X because it was proven to be less informative than a variable on one unexpected occasion.

Therefore, the transformer performs `cv` number of iterations.

If we make this class a "standard" feature-engine transformer, which is to be used in a pipeline, we may want to count and add a `threshold` param.

The transformer will drop features that have frequencies of being less informative than the best-performing random variable above the passed `threshold` value.

Morgan-Sell:
Thanks for the feedback, @solegalli!

Once you confirm that this class should be used in a sklearn pipeline, I'll implement the changes. 

I'm not informed enough on the subject to opine whether this selection method should be used only for exploratory purposes or not. Hopefully, @glevv can elaborate on his comments.

glevv:
In this thread I was talking about this class as more of an inspector than feature selector. But since we decided to go with it, I think it's reasonable to make more suitable to feature_engine package and code it as a feature selector with pipeline support.

glevv:
I have some suggestions though. 
I think it will be more efficient and straightforward to generate 1 noise feature N times, where of the noise and n_repeats will be hyperparameters. Another hyperparameter will be cut_rate or something similar. And the algorithm will be:

1. Generate noise with distribution (we could support normal, binary, uniform for example);
2. Train model;
3. Extract feature importance or coefs and save them to df;
4. Repeat 1,2,3 `n_repeat` times;
5. In resulting dataframe (n_features, n_repeats) find features that were worse than noise and sum across all trials. Divide this sum by `n_repeats` and compare it to `cut_rate`.
6. Drop features that were higher than `cut_rate`.

feature_engine/selection/probe_features_selector.py
solegalli: Let's update as per @glevv 's comment :)

feature_engine/selection/probe_feature_selector.py
```diff
@@ -0,0 +1,233 @@
...
+        # importance for each cross-validation fold
+        feature_importances_cv = pd.DataFrame()
+
+        # Populate the feature_importances_cv dataframe with columns containing
+        # the feature importance values for each model returned by the cross
+        # validation.
+        # There are as many columns as folds.
+        for i in range(len(model["estimator"])):
+            m = model["estimator"][i]
+            feature_importances_cv[i] = get_feature_importances(m)
```
solegalli, line 139: we are missing the step where we take the average of the importance across the different cv splits:

https://github.com/feature-engine/feature_engine/blob/a0106c968b14922d5123aaf51625fa66123dfad8/feature_engine/selection/base_recursive_selector.py#L182-L190

With this average, we are ready to select features. See for example here:
https://github.com/feature-engine/feature_engine/blob/a0106c968b14922d5123aaf51625fa66123dfad8/feature_engine/selection/base_recursive_selector.py#L182-L190

feature_engine/selection/probe_feature_selector.py
```diff
@@ -0,0 +1,233 @@
...
+        self._check_variable_number()
+
+        # save input features
+        self._get_feature_names_in(X)
+
+        # create probe feature distribution
+        self.probe_feature_data_ = self._generate_probe_feature(X.shape[0])
+
+        # get probe feature name
+        self.probe_feature_ = self.probe_feature_data_.columns
```
solegalli, line 114: Let's unify the 2 last parameters into one, called `probe_features_` that will contain a dataframe or a dictionary with the probe features that were created for the assessment.

feature_engine/selection/probe_feature_selector.py
```diff
@@ -0,0 +1,233 @@
...
+            feature_importances_cv[i] = get_feature_importances(m)
+
+        # add the variables as the index to feature_importances_cv
+        feature_importances_cv.index = X_new.columns
+
+        # save feature importances as an attribute
+        self.feature_importances_ = feature_importances_cv
+
+        # count the frequencies in which the variables are less important than the probe feature
+        self.vars_freq_low_importance_ = self._count_variables_importance_less_than_probe_feature()
```
solegalli, line 148: remove

solegalli (commented):
Hey @Morgan-Sell 

I've seen you've been doing some work. The logic of this transformer is very similar to the recursive feature selectors, so I added links that you could use and copy.

The main new bit that we need to work on is on how to create the probes.

Thank you!

feature_engine/selection/probe_feature_selector.py
```diff
@@ -0,0 +1,233 @@
...
+        Returns a dataframe comprised of the probe feature using the selected distribution.
+        """
+        # create dataframe
+        df = pd.DataFrame()
+
+        # set random state
+        np.random.seed(self.random_state)
+
+        if self.distribution == "normal":
+            df["rndm_gaussian_var"] = np.random.normal(0, 3, n_obs)
```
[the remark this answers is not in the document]
Morgan-Sell: Will it be useful to know the distribution when `all` is selected for the `distribution`?

feature_engine/selection/probe_feature_selector.py
Morgan-Sell: I didn't average the feature importances because I thought the class is intended to compare the probe feature's importance to the other variables importances for multiple models.

This is the purpose of the `threshold` param. If we average the feature importances, there is only one feature importance comparison.

solegalli: I don't understand.

This is the logic:

dataset has 3 variables, A, B and C.

train say random forest, derive 3 feature importance values (the mean of the cv), so that each one would have 1 value, ie, A: 0,5, B:0.6, C:0.7

But, we added probe features, probA and probB, which were also used to train the random forest and have importance of 0.5 each. Hence, we select features B and C, because importance is higher than 0.5

If probA and probB have 0.4 and 0.5, then the threshold needs to be above 0.45. Which now that I am writing it, I realize that we do not really need the parameter threshold. If importance is greater than probe, then select, otherwise remove.

feature_engine/selection/probe_feature_selector.py
```diff
@@ -0,0 +1,233 @@
...
+
+    def _generate_probe_feature(self, n_obs: int) -> pd.DataFrame:
+        """
+        Returns a dataframe comprised of the probe feature using the selected distribution.
+        """
+        # create dataframe
+        df = pd.DataFrame()
+
+        # set random state
+        np.random.seed(self.random_state)
```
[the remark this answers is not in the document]
Morgan-Sell: If we keep more than one probe feature, do we want to compare the other features' importance values to the maximum or the average value of the probe features' importance values?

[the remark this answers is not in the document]
solegalli: let's start with the average, and then we see, I'd say

feature_engine/selection/probe_feature_selector.py
Morgan-Sell: I understand. My interpretation of @glevv's suggestion is different though. I understood that @glevv recommended testing the feature importance during every iteration, counting the results, then creating a proportion that is compared to the `threshold`.

I'll implement your approach unless advised otherwise. I may not get to it until later this week.

solegalli: I'd start simple. We just compare averages. If we want to include the uncertainty, later we can use the standard deviation. But let's get a minimum viable version first. And then we see if and when it is worth adding layers of complexity.

Morgan-Sell:
Thanks for the guidance, @solegalli! Let me know what you think.

I started creating the unit tests. Afterwards, I'll work on the documentation ;)

solegalli (commented):
Awesome job @Morgan-Sell thank you!!

The class looks great. I added a few minor, cosmetic changes. But that's all :)

Well done. And thank you so much!

I look forward to tests and docstrings then :)

Morgan-Sell:
@solegalli,

I'm hoping that I'm not as big of a boludo as I think.

Did the incomplete version of ProbeFeatureSelection() get merged with the main branch?

I see you've made a lot of updates! Amazing!

I tried to rebase the `rndm_noise_sel` branch. All of my code has disappeared. I, more or less, finished the class.

I didn't expect a conflict to occur.

Any suggestions?

solegalli:
I reorganised the docstrings, hence i modified the base classes, including the selection ones, that's why you see the conflicts.

In theory, if you check out main and then rebase onto this feature branch, it should work. in theory.

Morgan-Sell:
Hi @solegalli, when I run `git rebase main` from the `rndm_noise_sel` branch, nearly all the code for ProbeFeatureSelection() is deleted the probe_featue_selection.py exists in `main` but it's an incomplete iteration of ProbeFeatureSelection().

How do I rebase without `main` overriding the more recent ProbeFeatueSelection() code?

Morgan-Sell:
@solegalli,

when I run `mypy feature-engine`, i receive the following error:

```
note: In member "fit" of class "ProbeFeatureSelection":
feature_engine/selection/probe_feature_selection.py:219: error: No return value expected
Found 1 error in 1 file (checked 97 source files)
```

Isn't the fit() method suppose to return `self`? May I go ahead and run `sphinx-build`?

solegalli:
Hi @Morgan-Sell 

Sorry for the delayed response. The mypy issue is resolved by removing `None` from the expected output. I fixed that and made a PR onto your repo with some changes. Please have a look at the PR, before continuing with these changes :)

First of all, thank you so much for the amount of work. The description of the class reads amazing. So good, that I was hoping you could draft the user guide as well :D

Second, the tests are great. So much work went into them. Well done. I've got one thing to ask, you did test the probe feature creation from within various tests. I would like though to have specific tests to test the 2 hidden methods: the one that creates the probe features, and the one that selects those based on the importance. Could you please add those 2 additional tests?

I commented out one test that is not passing. Could you check what is going on there and try to fix it?

I think the 2 things missing is 1, the test that is failing plus adding tests for the hidden methods, and the user guide. 

In the docstrings within the class, we need to also add references and a code example for users to copy and paste. You can get a template from the other feature selection classes. This is a new thing that we are slowly rolling out

Thank you!

solegalli:
And we also need to add the class to the test_check_estimator_selectors.py so that all common tests run on it as well. Would you mind doing that as well?

solegalli:
Quick question: any chance we can have this class ready by the end of the week? I'd like to release a new version of Feature-engine, and I just wondered if we can make this transformer part of it as well?

Morgan-Sell:
hi @solegalli,

Whenever I try to rebase to main, I pull an old version of the class. It was one of the original skeletons of ProbeFeatureSelection, e.g., there is no code for the transform() method.

Also, is there a type-o on contribute_code.rst. It states the following:

Pull the changes to your fork:

`git push -f origin main`


Should the code read `git pull -f origin main`?

solegalli:
Could you maybe check out this branch: https://github.com/feature-engine/feature_engine/tree/rndm_noise_sel in the main repo, make changes there and PR to this branch? or make a new PR with the changes?

PR #591, opened 2023-01-14
Add new transformer: GeometricWidthDiscretiser

### LINKED ISSUE #445: [new transformer] exponential width discretiser
when a variable is in a logarithmic scale, it might make sense to create the intervals based on a log scale instead of linear scale.

Quote:
"
When the numbers span multiple magnitudes, it may be better to group by powers of
10 (or powers of any constant): 0–9, 10–99, 100–999, 1000–9999, etc. The bin widths
grow exponentially
"

the idea is taken from: Feature Engineering for Machine Learning" Alice Zheng, O'Reilly.

glevv:
Closes #445 

- [x] Add tests
- [x] Add examples
- [x] Bug fixes


glevv:
It is somewhat unstable with small values. I think pandas could be the perpetrator, will look into it

docs/user_guide/discretisation/IncreasingWidthDiscretiser.rst
```diff
@@ -0,0 +1,134 @@
...
+	# Load dataset
+	data = pd.read_csv('houseprice.csv')
+
+	# Separate into train and test sets
+	X_train, X_test, y_train, y_test =  train_test_split(
+		    data.drop(['Id', 'SalePrice'], axis=1),
+		    data['SalePrice'], test_size=0.3, random_state=0)
+
+
+Now we want to discretise the 2 variables indicated below into 10 intervals of equal
```
solegalli, line 47: equal width? is this correct?

docs/user_guide/discretisation/IncreasingWidthDiscretiser.rst
```diff
@@ -0,0 +1,134 @@
...
+        2212.974,
+        inf]}
+
+With increasing width discretisation, each bin does not necessarily contain the same number of observations.
+It will also "skew" distribution. It could be useful to apply it to right skewed distribution
+to get more symmetrical one.
+
+.. code:: python
+
+	fig, ax = plt.subplots(1, 2)
```
solegalli, line 104: indent

feature_engine/discretisation/increasing_width.py
```diff
@@ -0,0 +1,134 @@
...
+    n_features_in_=_n_features_in_docstring,
+    fit_transform=_fit_transform_docstring,
+)
+class IncreasingWidthDiscretiser(BaseDiscretiser):
+    """
+    The IncreasingWidthDiscretiser() divides continuous numerical variables into
+    intervals of increasing width with equal increments. Note that the
+    proportion of observations per interval may vary.
+
+    The size of the interval will follow geometric progression.
```
solegalli, line 38: I don't understand from the docstring how the discretizer is creating the intervals. Could we add the formula / procedure?

feature_engine/discretisation/increasing_width.py
```diff
@@ -0,0 +1,134 @@
...
+    {fit}
+
+    {fit_transform}
+
+    {transform}
+
+    References
+    ----------
+    .. [1]  J. Reiser, "Classification Systems",
+        https://www.slideshare.net/johnjreiser/classification-systems
```
solegalli, line 82: Would it be worth adding additional resources? I haven't heard of this method, so I had to search. I found these which could be useful:
http://wiki.gis.com/wiki/index.php/Geometric_Interval_Classification
https://en.wikipedia.org/wiki/Geometric_progression

Also, it sounds like it is used for GIS data. Is this correct? If yes, maybe you could add a note on this either here or in the user guide?

Based on this, and according to the name the person uses in the slides you shared, would it not be better to call this transformer **GeometricIntervalDiscretiser**? Instead of increasingWidth?

If we use the same nomenclature, then users find it easier to join the dots and find more documentation and examples online.

feature_engine/discretisation/increasing_width.py
```diff
@@ -0,0 +1,134 @@
...
+            raise ValueError(f"bins must be an integer. Got {bins} instead.")
+
+        super().__init__(return_object, return_boundaries)
+
+        self.bins = bins
+        self.variables = _check_init_parameter_variables(variables)
+
+    def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
+        """
+        Learn the boundaries of the equal width intervals / bins for each
```
solegalli, line 104: needs re-wording

tests/test_discretisation/test_increasing_width_discretiser.py
```diff
@@ -0,0 +1,77 @@
...
+    # test case 4: when dataset contains na, transform method
+    with pytest.raises(ValueError):
+        transformer = IncreasingWidthDiscretiser()
+        transformer.fit(df_vartypes)
+        transformer.transform(df_na[["Name", "City", "Age", "Marks", "dob"]])
+
+
+def test_non_fitted_error(df_vartypes):
+    with pytest.raises(NotFittedError):
+        transformer = IncreasingWidthDiscretiser()
```
solegalli, line 76: Could we place commands that should not fail outside pytest raises?

I am discovering lots of bugs as I am refactoring the tests :_(

tests/test_discretisation/test_increasing_width_discretiser.py
```diff
@@ -0,0 +1,77 @@
+import numpy as np
+import pytest
+from sklearn.exceptions import NotFittedError
+
+from feature_engine.discretisation import IncreasingWidthDiscretiser
+
+
+def test_automatically_find_variables_and_return_as_numeric(df_normal_dist):
```
solegalli, line 8: This test structure is inheritance from my lesser programming skills from when I started :p

I think it would be better to break down this test into smaller tests:

- Test init params work and raise error when non permitted
- test fit method
- test transform method

Would you be able to do that? 

And did you run coverage on this class? it tells you which bits of code are not being tested yet.

Thank you!

docs/index.rst
```diff
@@ -138,6 +138,7 @@ Variable Discretisation: Discretisers
 - :doc:`api_doc/discretisation/EqualFrequencyDiscretiser`: sorts variable into equal frequency intervals
 - :doc:`api_doc/discretisation/EqualWidthDiscretiser`: sorts variable into equal width intervals
 - :doc:`api_doc/discretisation/DecisionTreeDiscretiser`: uses decision trees to create finite variables
+- :doc:`api_doc/discretisation/IncreasingWidthDiscretiser`: sorts variable into increasing width intervals
```
solegalli, line 141: following the slides you linked to the class, would it be not better to say: `sorts variable into geometrical intervals`?

docs/api_doc/discretisation/index.rst
```diff
@@ -17,6 +17,7 @@ into continuous intervals.
 :class:`EqualWidthDiscretiser()`         Sorts values into intervals of equal size.
 :class:`ArbitraryDiscretiser()`          Sorts values into intervals predefined by the user.
 :class:`DecisionTreeDiscretiser()`       Replaces values by predictions of a decision tree, which are discrete
+:class:`IncreasingWidthDiscretiser()`    Sorts variable into increasing width intervals
```
solegalli, line 20: following the slides you linked to the class, would it be not better to say: `sorts variable into geometrical intervals`?

docs/user_guide/discretisation/IncreasingWidthDiscretiser.rst
```diff
@@ -0,0 +1,134 @@
+.. _increasing_width_discretiser:
+
+.. currentmodule:: feature_engine.discretisation
+
+IncreasingWidthDiscretiser
+==========================
+
+The :class:`IncreasingWidthDiscretiser()` divides continuous numerical variables into
+intervals of increasing width with equal increments. Note that the
+proportion of observations per interval may vary.
```
solegalli, line 10: An example of the interval calculation would be extremely helpful, like the table the person uses in their slides. Maybe with different numbers, or alternatively clearly stating where the table was taken from.

docs/user_guide/discretisation/IncreasingWidthDiscretiser.rst
```diff
@@ -0,0 +1,134 @@
...
+
+This discretisation technique is great when the distribution of the variable is right skewed.
+
+The :class:`IncreasingWidthDiscretiser()` works only with numerical variables. A list of
+variables to discretise can be indicated, or the discretiser will automatically select
+all numerical variables in the train set.
+
+**Note**
+
+This discretiser could be numerically unstable when the values of the variable are small (their order of magnitude is small).
```
solegalli, line 20: You are looking into this, correct? It'd be great to say when it would be unstable and when one can safely use it.

docs/user_guide/discretisation/IncreasingWidthDiscretiser.rst
```diff
@@ -0,0 +1,134 @@
...
+        362.515,
+        399.894,
+        486.27,
+        685.871,
+        1147.115,
+        2212.974,
+        inf]}
+
+With increasing width discretisation, each bin does not necessarily contain the same number of observations.
+It will also "skew" distribution. It could be useful to apply it to right skewed distribution
```
solegalli, line 99: when would it skew the distribution? Could we add a bit more detail?

solegalli (commented):
Hi @glevv 

Great discretizing suggestion. I wasn't aware of this method.

I made a few suggestions here and there. The main one, the name of the discretizer, see if you agree.

Does it work with positive and negative variables? mathematically speaking it looks like so, but in practice? would it spread out the values in both cases when the variable is right skewed? Asking out of ignorance.

Thank you!

feature_engine/discretisation/increasing_width.py
glevv: `EqualWidth`
`EqualFrequency`
`IncreasingWidth`

This seems more consistent than:

`EqualWidth`
`EqualFrequency`
`GeometricInterval`

But I don't know

docs/user_guide/discretisation/IncreasingWidthDiscretiser.rst
glevv: The graph below clearly shows how it works

docs/user_guide/discretisation/IncreasingWidthDiscretiser.rst
glevv: Yep. In my code in the project I use np.digitize and sklearn logic from KBinsDiscretizer and it works fine. So I have my suspicions about pandas

glevv: Fixed

feature_engine/discretisation/increasing_width.py
solegalli: GeomtricWidth?

docs/user_guide/discretisation/GeometricWidthDiscretiser.rst
```diff
@@ -0,0 +1,137 @@
...
+=========================
+
+The :class:`GeometricWidthDiscretiser()` divides continuous numerical variables into
+intervals of increasing width with equal increments. Note that the
+proportion of observations per interval may vary.
+
+Sizes of the intervals will follow geometric progression
+
+    .. math::
+        a_i+1 = a_i r^(n+i)
```
solegalli, line 15: This formula is not displaying correctly. I changed the () by {} so that r is at the power of n+1

However, this formula is not that of slide 48 from the presentation linked in the docs: https://www.slideshare.net/johnjreiser/classification-systems

So where is it coming from?

solegalli: I am still having trouble with the formula. r^(n+i) is the constant cw, so why n+i?

feature_engine/discretisation/arbitrary.py
```diff
@@ -147,13 +147,17 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
 
         return self
 
-    def transform(self, X: pd.DataFrame) -> pd.DataFrame:
-        """Sort the variable values into the intervals.
+    def transform(self, X: pd.DataFrame, precision: int = 3) -> pd.DataFrame:
```
solegalli, line 150: why do we need this change?

solegalli (commented):
Hi @glevv 

Thank you so much for the amazing contribution. We are almost there!

Here are the main things:

I reworded the documentation. Please see this PR: https://github.com/glevv/feature_engine/pull/6

I am not sure I want to modify the default parameters of the `transform` method. Could we think of something else? Like making the parameter part of the init?

Why do we need to modify the ArbitraryDiscretizer as part of the PR? was there a bug? the intervals are passed by the user, so not sure why would we need to change them.

Could we please expand the tests? see for example some of the encoders, like this one, for inspiration: https://github.com/feature-engine/feature_engine/blob/main/tests/test_encoding/test_encoders/test_mean_encoder.py

We need to add the discretizer to the general checks: https://github.com/feature-engine/feature_engine/blob/main/tests/test_discretisation/test_check_estimator_discretisers.py

Thank you!

docs/user_guide/discretisation/GeometricWidthDiscretiser.rst
glevv: Its here https://en.wikipedia.org/wiki/Geometric_progression

feature_engine/discretisation/arbitrary.py
glevv: We have to increase precision to correctly calculate geometric bins with `pd.cut`. Its either add parameter to method of the base class or change the value inside from 3 to 7, which will impact all discretisers

feature_engine/discretisation/arbitrary.py
```diff
@@ -148,12 +148,16 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
         return self
 
     def transform(self, X: pd.DataFrame) -> pd.DataFrame:
-        """Sort the variable values into the intervals.
```
solegalli, line 151: could we please remove this file from the PR?

feature_engine/discretisation/base_discretiser.py
```diff
@@ -14,7 +14,7 @@ class BaseDiscretiser(BaseNumericalTransformer):
     """
 
     _return_object_docstring = """return_object: bool, default=False
-        Whether the the discrete variable should be returned as numeric or as
+        Whether the discrete variable should be returned as numeric or as
         object. If you would like to proceed with the engineering of the variable as if
         it was categorical, use True. Alternatively, keep the default to False.
         """.rstrip()
```
solegalli, line 20: could we add docstrings for precision? Something like this : `The precision at which to store and display the bins labels.`

solegalli: This parameter will be inherited by all discretizers, so we need to add the docstring in all files now I think.

feature_engine/discretisation/arbitrary.py
solegalli: Actually, we are now inheriting precision, so I think we need to add the docstrings instead in the init. Could you please check and see if we need to remove or update this file?

feature_engine/discretisation/gemoetric_width.py
```diff
@@ -0,0 +1,152 @@
...
+    ----------
+    {variables}
+
+    bins: int, default=10
+        Desired number of intervals / bins.
+
+    {return_object}
+
+    {return_boundaries}
+
```
solegalli, line 76: we need to add precision here, and in all discretizers

solegalli (commented):
Hi @glevv 

I made a few comments, but now I am thinking that you were probably still working on this PR. Apologies if that was the case.

Thank you!

feature_engine/discretisation/base_discretiser.py
glevv: I understand adding it to the base discretiser, but why would we add it to other, if this parameter supposed to be private and dev-only?

feature_engine/discretisation/gemoetric_width.py
solegalli: its private. We need the _ so that the user does not see it when pressing tab

glevv: I think sklearn does not support private parameters in init

solegalli:
Why do we need the `precision`? Is it for this instability that you mentioned? 

I still don't understand why we need to make it part of the base class. I guess it is because it is taken in the `pd.cut`, which is in the transform method, which is in the base class.

But in theory, we need it only in the GeometricDiscretiser, correct?

I am just trying to understand to see if there is a better design

glevv:
Yes, it is needed to solve numerical stability problem. I think we only need it in this discretizer. It is possible to rewrite transform method only adding precision parameter to pd.cut, but it feels a bit off

solegalli:
Yes, I agree. It wouldn't hurt to have the precision in the other transformers either.

There was an issue during rebase though. In main the code runs alright. But in this branch it does not. Have you tried to fix it? Then, I can help with the precision parameter in the other transformers.

solegalli:
looks good to me. 

Is this ready to go @glevv from your side?

glevv:
I think it should be OK. I'll recheck it later

feature_engine/discretisation/geometric_width.py
```diff
@@ -0,0 +1,161 @@
...
+    n_features_in_=_n_features_in_docstring,
+    fit_transform=_fit_transform_docstring,
+)
+class GeometricWidthDiscretiser(BaseDiscretiser):
+    """
+    The GeometricWidthDiscretiser() divides continuous numerical variables into
+    intervals of increasing width with equal increments. Note that the
+    proportion of observations per interval may vary.
+
+    Sizes (a) of n intervals will follow geometric progression
```
solegalli, line 51: Why did you remove all the additional info that I added here?

feature_engine/discretisation/geometric_width.py
```diff
@@ -0,0 +1,161 @@
...
+    A list of variables can be passed as argument. Alternatively, the discretiser
+    will automatically select all numerical variables.
+
+    The GeometricWidthDiscretiser() first finds the boundaries for the intervals for
+    each variable. Then, it transforms the variables, that is, sorts the values into
+    the intervals.
+
+    Note: this discretiser needs higher precision, because bins width could be very
+    small. With smaller range of the binned values and/or higher number of bins,
+    smaller precision values could be needed.
```
solegalli, line 66: We need to re-word this paragraph a bit.

feature_engine/discretisation/geometric_width.py
```diff
@@ -0,0 +1,161 @@
...
+    def __init__(
+        self,
+        variables: Union[None, int, str, List[Union[str, int]]] = None,
+        bins: int = 10,
+        return_object: bool = False,
+        return_boundaries: bool = False,
+        precision: int = 7,
+    ):
+
+        if not isinstance(bins, int):
```
solegalli, line 123: here we probably also need a check that bins > 0? or we leave that to pandas.cut?

feature_engine/discretisation/geometric_width.py
glevv: I did not, maybe when I did rebase it got erased. Do you have a copy?

feature_engine/discretisation/geometric_width.py
glevv: We could leave all checks of precision to pandas, no need to duplicate it

feature_engine/discretisation/geometric_width.py
solegalli: [feature_engine/discretisation/gemoetric_width.py](https://github.com/feature-engine/feature_engine/blob/7ceb9827406ac2c0e73d83c80a4dd9e751d21bf8/feature_engine/discretisation/gemoetric_width.py)

In this PR: https://github.com/glevv/feature_engine/pull/6

feature_engine/discretisation/geometric_width.py
solegalli: How about this?

Note: The width of some bins might be very small. Thus, to allow this transformer to work properly, it might help to increase the precision value, that is, the number of decimal values allowed to define each bin. If the variable has a narrow range or you are sorting into several bins, allow greater precision (i.e., if precision = 3, then 0.001; if precision = 7, then 0.0001).

feature_engine/discretisation/geometric_width.py
glevv: Already change it

solegalli:
All done! Thank you @glevv !

PR #604, opened 2023-01-29
take docstrings outside discretiser base class

solegalli:


solegalli:
FYI @glevv 

Please rebase

PR #614, opened 2023-02-15
Speed up DropDuplicateFeatures()

### LINKED ISSUE #580: drop_duplicate_features with huge datasets
When running drop_duplicate_features() with a huge dataset (More that 60K features) it lasts forever (36 hours and couting), while simple code:

       # Remove duplicated features
        duplicated_feat = []
        for i in range(0, len(X_train.columns)):
            # if i % 10 == 0:  # this helps me understand how the loop is going
            #     print(i)
 
            col_1 = X_train.columns[i]

            for col_2 in X_train.columns[i + 1:]:
                if X_train[col_1].equals(X_train[col_2]):
                    duplicated_feat.append(col_2)

Took half an hour to finish. Please check.

Okroshiashvili:
This PR speeds up duplicate feature removal code. It fixes #580 

Please, review the PR @solegalli and @NicoGalli

### DIFF
--- a/feature_engine/selection/drop_duplicate_features.py
+++ b/feature_engine/selection/drop_duplicate_features.py
@@ -148,7 +148,9 @@ class DropDuplicateFeatures(BaseSelector):
 
         # create tuples of duplicated feature groups
         self.duplicated_feature_sets_ = [
-            set(duplicate) for duplicate in _features_hashmap.values() if len(duplicate) > 1
+            set(duplicate)
+            for duplicate in _features_hashmap.values()
+            if len(duplicate) > 1
         ]
 
         # set to collect features that are duplicated
--- a/feature_engine/selection/drop_duplicate_features.py
+++ b/feature_engine/selection/drop_duplicate_features.py
@@ -1,3 +1,4 @@
+from collections import defaultdict
 from typing import List, Union
 
 import pandas as pd
@@ -99,7 +100,6 @@ class DropDuplicateFeatures(BaseSelector):
         missing_values: str = "ignore",
         confirm_variables: bool = False,
     ):
-
         if missing_values not in ["raise", "ignore"]:
             raise ValueError("missing_values takes only values 'raise' or 'ignore'.")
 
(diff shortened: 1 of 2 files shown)


solegalli:
closing and re-opening to see if that triggers circle ci checks again.

solegalli:
Here we go!

solegalli:
Hi @Okroshiashvili 

This looks good to me. Thank you!

Is it possible somehow to add a (speed) benchmark test, so that we do not accidentally make it slower if we introduce some changes down the line? Asking out of ignorance. 

Otherwise, it's good to merge I'd think.

Okroshiashvili:
Hi @solegalli 

I don't think it's possible to add speed benchmark tests to test my solution vs current one on main branch. However, what I can do is to add test for the new solution. Note that, with this new test case, we increase overall time for unit tests. If that sounds okay then I will do that

solegalli:
Sounds good! Please add the new test. Thank you!

Okroshiashvili:
@solegalli 

I just realised that adding this sort of test does not make any sense. Let me explain.
By adding the test for speed we won't be able to measure how fast the new solution is. Nor other users will be able to replicate my timings.

We can do the following: maintainers can pull this branch and perform speed benchmarking. I used  the following code to generate random DataFrame with many features.

```
import pandas as pd
import numpy as np

def df_with_many_features():
    np.random.seed(425)
    col_names = [f"col_{i}" for i in range(60000)]
    df = pd.DataFrame(np.random.randint(0, 100, size=(10000, 60000)), columns=col_names)

    df["col_2"] = df["col_1"]
    df["col_3"] = df["col_1"]
    df["col_4"] = df["col_1"]
    df["col_30000"] = df["col_50"]
    df["col_43000"] = df["col_50"]
    df["col_0"] = df["col_59999"]

    return df
```
And using this DataFrame came up with the solution.

I guess this is the only logical thing we can do here.

solegalli:
Alright. Thanks!

I'll go ahead and merge this PR then.

Issue #612, opened 2023-01-31
Correlation-based feature selection is column-order dependent


dlaprins:
Correlation-based feature selection makes use of a for-loop running over the features, and checking whether or not they are collinear to another feature (see for example line 183 of [selection.drop_correlated_features.py](http://selection.drop_correlated_features.py/)). In my view, this is problematic, since it introduces a dependency on column-ordering in the outcome of the selection.
 
As an example, consider the following feature set: 
``` 
X = pd.DataFrame()
X['a'] = [1, -1, 0, 0, 0, 0]
X['b'] = [0, 0, 1, -1, 1, -1]
X['c'] = [0, 0, 1, -1, 0, 0]
X['d'] = [0, 0, 0, 0, 1, -1]
``` 
These features are constructed in such a way that, given a correlation-threshold of 0.7,
a is uncorrelated, b is collinear to c and d, and c and d are collinear only to b. 
 
If we run 
```
DropCorrelatedFeatures(threshold = 0.7).fit(X)
```
the features [‘a’, ‘b’] will be selected. On the other hand, if we run 
```
DropCorrelatedFeatures(threshold = 0.7).fit(X[‘a’, ‘c’, ‘b’, ‘d’])
```
the featuers [‘a’,’c’,’d’] will be selected instead. 
 
Similarly, SmartCorrelatedSelection() is also not invariant under such reordering of the columns of X. 

As a solution, ideally, whatever feature set is selected satisfies (at least) the following two constraints, in addition to being independent of column-order: 
1) The set of features selected contains no features collinear to one another (by definition)
2) Given such a set of features, there is no feature that can be added to the set without introducing collinearity between two features

As a first attempt, I tried to construct the full set of sets that satisfy these two constraints. Unfortunately, this is computationally intensive and quickly becomes impractical for datasets with > 100 features. 

I am not aware of any faster algorithm that is guaranteed to return a feature set that satisfies both constraints. But a minimal improvement would be to remove features based on a rank-ordering of the features rather than column order: what I have in mind for such a rank-ordering would be number of features the feature is collinear with, information value, correlation with the target, or (like in SmartCorrelatedSelection()), estimator-based.  Such a rank-ordering can be an input variable, with a certain default approach.

I would be happy to further work on this issue once methodology is clear, if this is of interest to others.

solegalli:
Hi @dlaprins 

[...]

PR #619, opened 2023-02-28
Fixed DropCorrelatedFeatures to be independent of column ordering. 

dlaprins:
#612
Order of features dropped is now determined either by variance (default), or custom ordering (optional list as a variable).


dlaprins:
The following changes were made: 
1) DropCorrelatedFeatures now uses an ordering for the features. The default ordering is by variance of the feature (from high to low), or an optional given list. The algorithm considers each feature in order: if it is not collinear with the set of all kept features, it is added to the set of kept features. Otherwise, it is dropped. 

2) The attribute correlated_feature_sets_ consisted of sets of features that were correlated (above the threshold) to another feature. However, correlation above a threshold is not transitive: For features a,b,c, if a is correlated to b and b to c, then a need not be correlated to c. 
I have replaced this attribute with a dictionary. The dictionary has features as keys, with as values all features that are correlated above the threshold to the key-feature.

glevv:
I think adding `ordering` parameter to init that will take string ('std', 'uniques', 'nans', 'alphabetical' etc) or None (use order as passed to transformer) is clearer.

solegalli:
Hey @dlaprins thank you so much for the code changes!

I agree with @glevv - We need to make that parameter part of init, and leave fit to take just X and y. 

It would also be great to offer more options to order the variables (ie what @glevv suggests). On this note, it would be useful to have the ordering functionality captured in a hidden method, so that we can test it individually in a dedicated test.

Regarding the features and the transitive property, I have a question: Say we have features a, b, c, d and e:

- a is correlated with b and c. 
- b is correlated with d
- c is correlated with e.

Will this be the resulting dictionary:
```
{
a: [b, c],
b: [d],
c: e
}
```
?

And if that is the case, which features will be retained? a, b and c? even though c is correlated with a and thus redundant? or how is this situation handled?

I have another question: is by any chance your implementation faster than the current implementation? would you know by how much?

Could we please not use \ to break lines? to be consistent with the rest of the codebase?

Thank you so much for your contribution!

dlaprins:
> I think adding `order_by` parameter to init that will take string ('std', 'unique', 'nan', 'alphabetical' etc) or None (use order as passed to transformer) is clearer.

Sounds fine by me. Note that at that point, pretty much all of SmartCorrelatedFeatures functionality is included in DropCorrelatedFeatures. Also, I would strongly be in favour of not using None as default, but to put big exclamation marks and red arrows there: naively speaking, one would expect None to not do anything special, but in this case, it severely effects functionality and should only be used in case one has a preference for feature ordering based on metrics not available with the string options.

solegalli:
We should expand the user guide with this information (ie., @dlaprins), mentioning that the order is crucial, because it will affect the output of the transformer, and that we encourage users to order the features based on blabla. 

In other words: I agree with putting big exclamation marks and red arrows, but in the user guide please :) We could even add one example of running the class with features in order a, and then order b and compare outputs. In the class docstrings, let's just mention briefly and refer the reader to the user guide.

Having it defaulted to None will not break backwards compatibility, so that is a plus.

solegalli:
Hey @dlaprins 

Just checking in to see if we could have this PR ready for the next week? so we make it part of the next release?

Thank you!

dlaprins:
Hi @solegalli, I would like to, but realistically, I don't think I'll have the time available. Please go ahead and release without waiting on my contribution.

dlaprins:
Hi @ClaudioSalvatoreArcidiacono , sorry for the late response. Yes, you are completely correct: the boolean correlation matrix is the adjacency matrix. In fact, the code I privately used was written exactly in those terms, I rewrote the feature names for the PR here just for clarity.

I'm not sure getting the connected components without knowledge of the edge-structure is useful, and if you have the edges (either as a dictionary or as the adjacency matrix), then getting the connected components seems rather trivial. But if it's kept, I would suggest that it requires either a renaming of the attribute or at the very least extensive documentation to avoid the misconception about transitivity of correlation (or in graph language: the connected component is not a complete graph). 

@solegalli I apologise for not having to temporarily drop out before, I noticed that it was picked up and taken over to ensure it could be pushed into the new release. But I see it's still open. What is the current status of this feature?

solegalli:
Hey guys, @dlaprins @ClaudioSalvatoreArcidiacono, thank you for the engagement with the issue.

@dlaprins we did open a separate issue. @glevv worked on it. But then I discovered some bugs in the correlation selector that need to be fixed before we do anything else. Or at least they need to be taken care of as part of the changes you suggest here.

The bug is the following:

If we have features a, b, c and d, and say a, c, d are correlated, the transformer works as expected: It finds acd as correlated group, a is retained, and c and d are dropped. 

But if we change the order of the features in the train set to say bcad, now I would expect c to be retained and a and d dropped, and that does not happen. So there is something weird going on that we need to look at.

Regarding nodes and graphs suggestions on how to represent features, I am not sure what to say.

The idea of this transformer is to drop correlated features. In the brute force approach, if a series of features are correlated with a first feature, say A, then they are all dropped and the first one (A) retained. It is not strictly important if one of the features correlated to A is also correlated with something else that is not in the group, because it will be dropped anyways.

Now, that is not so neat and tidy when we do a smart correlation, because from a group of features correlated with feature A, we may choose a feature in the middle from that group, different from A, and drop the rest. That selected feature may also be correlated with something else that was not correlated with A, and hence, we still have correlated variables in the final train set. 

One possible solution could be that we iterate the correlation search and elimination process, so that if correlated features remain after the first search they can be removed in the second round.

If you have a better solution, I am very keen to hear. Let me know.

And finally, super apologies for the delayed response. I am going on holidays next Thursday, and I was super busy trying to leave everything ready at work.

Cheers!

dlaprins:
Hi @solegalli, 

> The bug is the following:

If we have features a, b, c and d, and say a, c, d are correlated, the transformer works as expected: It finds acd as correlated group, a is retained, and c and d are dropped.

But if we change the order of the features in the train set to say bcad, now I would expect c to be retained and a and d dropped, and that does not happen. So there is something weird going on that we need to look at.

Could you post the specific dataset for which this happened? I'd be happy to take a look. For what it's worth, I performed some checks for such a phenomenon with the PR I made before, there it did not happen. But I understand that it might be faster to solve the bug here rather than to go back to my version of the code. 

> Now, that is not so neat and tidy when we do a smart correlation, because from a group of features correlated with feature A, we may choose a feature in the middle from that group, different from A, and drop the rest. That selected feature may also be correlated with something else that was not correlated with A, and hence, we still have correlated variables in the final train set.

Given an order of preference of features, this problem should be solved. By looping in order of preference, the situation you mention (where one investigates feature A, finds that it is correlated to feature B, notes that B is higher on the ordering and thus removes A rather than B) does not occur. 

To be honest, I think that all functionality of SmartCorrelatedSelection (with the exception of integrated model-based feature ordering) is already present in the regular DropCorrelatedFeatures, once ordering functionality is integrated. 

> One possible solution could be that we iterate the correlation search and elimination process, so that if correlated features remain after the first search they can be removed in the second round.

I do not believe that this is necessary, and I think that fixing the algorithm in such a way that such additional complexity can be avoided would be preferable. 

> And finally, super apologies for the delayed response. I am going on holidays next Thursday, and I was super busy trying to leave everything ready at work.

I'm hardly in a position to complain considering Gleb picked up my last PR :)

solegalli:
Actually, it is not correct what I said. The class is working properly. It is just that it is difficult to understand how it is making the decisions.

For example, this code:

```
import pandas as pd
from sklearn.datasets import make_classification
from feature_engine.selection import DropCorrelatedFeatures

X, y = make_classification(
    n_samples=1000,
    n_features=12,
    n_redundant=4,
    n_clusters_per_class=1,
    weights=[0.50],
    class_sep=2,
    random_state=1,
)

# transform array into pandas df
colnames = ["var_" + str(i) for i in range(12)]
X = pd.DataFrame(X, columns=colnames)

tr = DropCorrelatedFeatures()
tr.fit(X)

tr.correlated_feature_sets_
```

Results in the following correlated sets:

```
[{'var_0', 'var_8'}, {'var_4', 'var_6', 'var_7', 'var_9'}]
```

And from that, it will retain var0, var4 which are the first ones in each set, and drop the rest:

```
tr.features_to_drop_
```

As seen here:

```
{'var_6', 'var_7', 'var_8', 'var_9'}
```

So in this case, it is very easy to know which variables will be retained and which ones will be dropped.


However, if I have the variables in the reverse order and I repeat the process:

```
varnames = X.columns.to_list()
varnames.reverse()

tr = DropCorrelatedFeatures()
tr.fit(X[varnames])
tr.correlated_feature_sets_
```

Then, the sets are still correct, but from the display below, we can't infer which one will be retained:

```
[{'var_4', 'var_6', 'var_7', 'var_9'}, {'var_0', 'var_8'}]
```

It does retain variables 8 and 9 as expected from the inversed order though:

```
tr.features_to_drop_
```

```
{'var_0', 'var_4', 'var_6', 'var_7'}
```

But the user doesn't really know which variable will be retained. 

I wonder if there is a better way of displaying the sets, so that it is obvious which variable will be retained from each set.

That's one thing that was not part of the issue I guess.

@dlaprins happy for you to give it a go, either in this PR or in a new one.

Only thing to consider, whenever possible, not to break backwards compatibility.

It is not always possible though

Thank you

Issue #592, opened 2023-01-14
add profiling functionality to feature-engine


solegalli:
Using cProfile, we need to be able to test the performance of the entire library.

Okroshiashvili:
Hi @solegalli 

1. What do you think what has to be done here?
2. Will this profiling functionality be only for devs or for end users as well?
3. Where profiling functionality be run? I mean, from where we (devs) will get profile output?
4. Are we only interested in timings and not memory footprint?
5. Are we okay with third party library for profiling?

solegalli:
Hey @Okroshiashvili 

1. we need to understand the performance of each one of our classes. For example, drop duplicates was super slow. I'd like to be able to pick that up before making the class public by using profiler. Because we already have a lot of public classes, this would help us understand if they are performant, and if not, modify them
2. devs
3. I don't understand the question. I normally run pytest for tests, pytest with coverage for coverage, and here I guess we would run the profiler for the profile. Should we also make this part of the standard circleci checks? Not sure about this.
4. both speed and memory (we have a few issues regarding using a lot of memory by pd.copy()
5. yes. I thought of eprofiler

I am new to this, so I asked for help to @ChristopherGS 

I copy here the notes that I got from him:

```
def profile(output_file=None, sort_by="cumulative", strip_dirs=False):

      """A time profiler decorator.
      Inspired by and modified the profile decorator of Giampaolo Rodola:

      http://code.activestate.com/recipes/577817-profile-decorator/

      Args:
      output_file: str or None. Default is None
      Path of the output file. If only name of the file is given, it's
      saved in the current directory.
      If it's None, the name of the decorated function is used.

      sort_by: str or SortKey enum or tuple/list of str/SortKey enum
      Sorting criteria for the Stats object.
      For a list of valid string and SortKey refer to:
      https://docs.python.org/3/library/profile.html#pstats.Stats.sort_stats

      strip_dirs: bool
      Whether to remove the leading path info from file names.
      This is also useful in reducing the size of the printout
      Returns:
      Profile of the decorated function
      """

    # nosemgrep
    def inner(func):
    # nosemgrep
    @wraps(func)
    # nosemgrep
    def wrapper(*args, **kwargs): # nosemgrep
    if not ENABLE_PROFILING:
    return func(*args, **kwargs)
    _output_file = output_file or UPDATE_ME / "sim_profile.prof"
    pr = cProfile.Profile()
    pr.enable()
    retval = func(*args, **kwargs)
[...]

PR #622, opened 2023-03-01
Add small utility to profile any function

Okroshiashvili:
Addresses #592 issue

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -105,6 +105,9 @@ venv.bak/
 # mypy
 .mypy_cache/
 
+# profiling
+/profiles
+
 # Miscelaneous
 .idea
 .vscode
--- a/docs/contribute/contribute_code.rst
+++ b/docs/contribute/contribute_code.rst
@@ -449,6 +449,76 @@ you want to know why we prefer tox, this
 will tell you everything ;)
 
 
+Code Profiling
+--------------
+
+If you want to profile your code, you can use the **profiling** module in root directory. There you will find two files,
+`profiling.py` and `profiling.sh`. Both file does the same thing but in different ways. The profiling.py file is a python script
+containing a function that must be used as a decorator for the class/method we want to profile.
+The profiling.sh file is a bash/zsh script that you can run from the command line to profile whole .py script file.
+Let us see how to use them. First, start with profiling.py file.
+
+I doubt that `DropDuplicateFeatures` class should take more time than other classes as it iterates over the columns and
+checks if they are duplicated or not. So, I will profile the `DropDuplicateFeatures` class.
+
+First, I will find where this class resides and on top of the imports I will add the following line::
+
+    from profiling.profiling import profile_function
+
+Now, I will decorate the `DropDuplicateFeatures.fit` method with the `profile_function` function::
+    
+        @profile_function(output_file="profile.html")
+        def fit(self, X: pd.DataFrame, y: pd.Series = None):
+            ...
+
+The next step is to create a temporary .py file that will contain the code that we want to profile.
+
+For example, I will create a file named `temp.py` and copy the following code::
+
+    import pandas as pd
+    import numpy as np
+
+    from feature_engine.selection import DropDuplicateFeatures
+
+
+    if __name__ == "__main__":
+        rows = 10000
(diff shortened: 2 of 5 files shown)


Okroshiashvili:
@solegalli Whenever you'll find time please have a look

solegalli:
Hey @Okroshiashvili 

This looks good, thank you!

We need a few more details:

- Could we add scripts to sample every class? Maybe a script per module (imputation, encoding, etc), within profiling, with the function decorating the class so that we can run them locally and see the profile?
- Details in the contribute documentation with code examples of how we can use the prifiling module to check the function
- Details in the contribute documentation telling users how they should use this function when developing a new class, and how they can visualize the results?

In short, if you go here: https://feature-engine.trainindata.com/en/latest/contribute/contribute_code.html

You'll see that we tell users how to run and check the tests: `pytest`, or code style: `mypy feature-engine` or `flake8 feature-engine tests`, and all that functionality is already available for developers.

I've just noticed that we need info to explain how to run coverage :/

And we need to add info to explain how to run the profiler, but the user should be able to hit something like `profile feature-engine` and then the profiling for all classes, or alternatively, the desired classes, shows up.

Okroshiashvili:
Hi @solegalli 

Thanks for your input. I have crazy week 😕 I hope I will return to this issue by the end of the week.

Btw, I'm testing new library for memory profiling as well. There are some difficulties and will update you as soon as I figure out things there.

solegalli:
Thank you so much!

Okroshiashvili:
@solegalli

Here is a Python library [for memory profiling](https://github.com/bloomberg/memray). Please, have a look and if it looks good then I will work to have it for feature engine

solegalli:
Hi @Okroshiashvili 

Thanks for the files. And sorry for the delayed response.

I don't understand how I am meant to use these files though.

We have the profiling function in profiling.py

Now, say I want to profile the `MeanMedianImputer`. How do I do that?

And how should I use profiling.sh locally? and what would be the outcome of executing that file?

It would help if you could add information in the contribute guide (https://github.com/feature-engine/feature_engine/blob/main/docs/contribute/contribute_code.rst) so that a user can actually use this functionality.

Okroshiashvili:
@solegalli 

Sorry for that confusion. Let me add appropriate documentation about profiling (time as well as memory) and then decide what to do

Okroshiashvili:
Hi @solegalli 

Sorry for the latency 😔

I've updated docs. Please, check and if all good then we can go and merge.

solegalli:
Hey @Okroshiashvili 

This looks really good! Thank you for adding such clear details in the documentation. That's great. 

I am going on hols from tomorrow till Monday 15th. As soon as I am back, I'd like to play around with the code a bit, and then I merge.

Sorry for the delay on my side now ;)

Have a good weekend!

solegalli:
Hey @Okroshiashvili 

Thank you so much for the contribution and the patience with me coming back to this.

I just tested the code you implemented.

I made a PR: https://github.com/Okroshiashvili/feature_engine/pull/1 with minor rewording on the contribute file.

This is my experience from running the code as in the contribute:

When doing it via the python script, I do get the profile.html. 

The content of the profile.html file is very limited. I don't understand how that is helpful. Maybe this is me being completely ignorant about profiling, but I just get speeds? It doesn't really tell me what could be optimized, which is what I was hoping for. I don't understand how I can use that information to improve my code.

When running the file via bash, I still get the profile.html, and then a folder with another folder inside, and in it 2 files. But the app you mention in the documentation didn't manage to open any of those files. Not sure what went wrong.

Maybe I need to do some reading about profiling. Maybe I am expecting more from it that it can give. My idea for this functionality was to to have a function that provides some information about the performance of the code, and ideally how it can be improved.

One user made a PR sometime ago regarding copy() from pandas, and they said that they did profiling, and they discovered that the transformers from feature engine were slow because that command, copy() which comes from pandas is putting a lot of data in memory. This is useful actionable insight. Speeds with nothing else, I am not sure how it helps.

I am happy to admit that my knowledge on the topic is very little though. I might be totally wrong with my expectation.

solegalli:
Closed in #984

PR #625, opened 2023-03-05
refactor code according to pandas 1.5.0 and sklearn 1.2.0 releases

luismavs:
Update to fix some Warnings when running code with recent Pandas (>=1.5.0) and Scikit-Learn (>=1.2.0).

Specifically, the following types of warnings were fixed:

```
FutureWarning: In a future version of pandas all arguments of DataFrame.any and Series.any will be keyword-only.
    assert (X.isna().any(1) == df_enc_big_na.isna().any(1)).all()`

FutureWarning: The `inplace` parameter in pandas.Categorical.add_categories is deprecated and will be removed in a future version. Removing unused categories will always return a new Categorical object.
    X[variable].cat.add_categories(

DeprecationWarning: In a future version, `df.iloc[:, i] = newvals` will attempt to set the values inplace instead of always setting a new array. To retain the old behavior, use either `df[df.columns[i]] = newvals` or, if columns are non-unique, `df.isetitem(i, newvals)`
    X.fillna(value=self.imputer_dict_, inplace=True)

SettingWithCopyWarning: 
  A value is trying to be set on a copy of a slice from a DataFrame
  
  See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
    data["var_3"].iloc[15] = np.nan

DeprecationWarning: In a future version, `df.iloc[:, i] = newvals` will attempt to set the values inplace instead of always setting a new array. To retain the old behavior, use either `df[df.columns[i]] = newvals` or, if columns are non-unique, `df.isetitem(i, newvals)`
    X.loc[:, self.variables_] = np.power(X.loc[:, self.variables_], 1 / self.exp)

FutureWarning: `sparse` was renamed to `sparse_output` in version 1.2 and will be removed in 1.4. `sparse_output` is ignored unless you leave `sparse` to its default value.
    warnings.warn(
```


solegalli:
Hey @luismavs 

Awsome job, thank you!

We need a small fix here for style tests:

```
tests/test_wrappers/test_sklearn_wrapper.py:55:89: E501 line too long (94 > 88 characters)
```

solegalli:
Is this PR finished @luismavs ?

If you rebase main, the test docs should pass. I fixed them in #627 

Alternatively, I think it is safe to merge like this as well, because I know where the error comes from and it should be fixed in main.

Just waiting for your reply either way! Thank you!

test_run_cut.txt
```diff
@@ -0,0 +1,166 @@
+==================== solved =====================
+
```
solegalli, line 2: We also need to remove this file from the PR

luismavs:
> I just noticed 2 extra files that I think helped you find the warnings. If you remove those, you can probably also rebase main and then the test would pass.
> 
> Let me know whenever you think this is ready to merge.
> 
> I've seen some discussions on linkedin about this issue from pandas. Thank you so much for picking it up! :)

Hi, fixed and merged with main.

solegalli:
Hey @luismavs 

Not sure what went wrong, but something broke with the last commits. Could you have a look?

I'll really appreciate it :)

solegalli:
closing and reopening to see if tests pass

solegalli:
Hey @luismavs ignore my comment, it was some failure to load a dataframe from an url, for another unrelated test.

Merging as we speak. Thanks a lot for the contribution!

PR #633, opened 2023-03-12
DropCorrelatedFeatures and SmartCorrelatedSelection consistency changes

### LINKED ISSUE #612: Correlation-based feature selection is column-order dependent

glevv:
Fixes #612

Will also close #619

### DIFF
--- a/docs/user_guide/selection/DropCorrelatedFeatures.rst
+++ b/docs/user_guide/selection/DropCorrelatedFeatures.rst
@@ -7,21 +7,23 @@ DropCorrelatedFeatures
 
 The :class:`DropCorrelatedFeatures()` finds and removes correlated variables from a dataframe.
 Correlation is calculated with `pandas.corr()`. All correlation methods supported by `pandas.corr()`
-can be used in the selection, including Spearman, Kendall, or Spearman. You can also pass a
+can be used in the selection, including Pearson's, Kendall, or Spearman. You can also pass a
 bespoke correlation function, provided it returns a value between -1 and 1.
 
 Features are removed on first found first removed basis, without any further insight. That is,
-the first feature will be retained an all subsequent features that are correlated with this, will
+the first feature will be retained an all subsequent features that are correlated with it, will
 be removed.
 
 The transformer will examine all numerical variables automatically. Note that you could pass a
 dataframe with categorical and datetime variables, and these will be ignored automatically.
 Alternatively, you can pass a list with the variables you wish to evaluate.
 
-Note: ordering of the coulmns matters! Different order of the columns could lead to
-different results. To make estimator more consistent, please use 'order_by' parameter.
+Note: the order of the variables in the dataset matters! A different variable order can
+lead to different results. To make the transformer consistent, use the `'order_by'`
+parameter.
 
-**Example**
+Example
+-------
 
 Let's create a toy dataframe where 4 of the features are correlated:
 
@@ -75,7 +77,7 @@ The correlated feature groups are stored in the transformer's attributes:
     [{'var_0', 'var_8'}, {'var_4', 'var_6', 'var_7', 'var_9'}]
 
 
-As well as the features that will be removed from the dataset:
+The features that will be removed from the dataset are also stored in an attribute:
 
 ..  code:: python
 
@@ -108,19 +110,30 @@ have been removed.
     3  0.484649
     4 -0.186530
 
-Now let's reduce number of unique values and use order_by = 'unique' option.
+The variable order matters
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Let's now compare how the order of the variables in the training dataset may affect the
+result, that is, the selected features.
+
+We'll start by reducing the number of unique values in our dataframe and repeat the
+feature selection procedure, without altering the order of the variables in the training
+set.
 
 .. code:: python
 
+    X = X.round(3)
     tr = DropCorrelatedFeatures(variables=None, method='pearson', threshold=0.8)
-    Xt = tr.fit_transform(X.round(3))
+    Xt = tr.fit_transform(X)
     tr.correlated_feature_sets_
 
-We will get the same result
+We see in the following output that we obtain the same feature sets as previously:
 
 ..  code:: python
 
     [{'var_0', 'var_8'}, {'var_4', 'var_6', 'var_7', 'var_9'}]
+
+And the transformer will remove the same variables as before:
 
 ..  code:: python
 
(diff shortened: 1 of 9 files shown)


glevv:
Decided to keep only {None, 'nan', 'unique', 'alphabetical'}, since std could be confusing with different scales of features

feature_engine/selection/drop_correlated_features.py
```diff
@@ -73,6 +77,15 @@ class DropCorrelatedFeatures(BaseSelector):
...
         another one and removed from the dataset.
 
+    order_by: str, default=None
+        Type of sorting to use before feature selection. This option could help with
+        the consistency of the selection.
+
+        - None - preserve original order of the dataframe
+        - 'nan' - sort columns by number of missing values (ascending)
+        - 'unique' - sort columns by number of unique values (descending)
+        - 'alphabetic' - sort columns alphabetically.
```
solegalli, line 87: I am thinking the following:

- alphabetic isn't really a meaningful way of ordering. It has nothing to do with the variable's predictive value. It helps for consistency and is easy and fast, but other than that it does not add much
- nan has some value in that it will preserve the variables with less nan, but it would be useful only before imputation and when data does have a lot of nan.
- unique will take ages if all variables are continuous

Hence, I think it might be useful to add the additional option to use the std, and do tell users that the limitation is the feature scale, so to scale features when using this option.

Would you agree?

feature_engine/selection/drop_correlated_features.py
```diff
@@ -133,22 +146,35 @@ def __init__(
...
         missing_values: str = "ignore",
         confirm_variables: bool = False,
     ):
 
         if not isinstance(threshold, float) or threshold < 0 or threshold > 1:
             raise ValueError("threshold must be a float between 0 and 1")
 
-        if missing_values not in ["raise", "ignore"]:
+        if missing_values not in ("raise", "ignore"):
             raise ValueError("missing_values takes only values 'raise' or 'ignore'.")
```
solegalli, line 158: could we add the f"Got {missing_values} instead." please? since we are here :)

feature_engine/selection/drop_correlated_features.py
```diff
@@ -133,22 +146,35 @@ def __init__(
...
 
+        if order_by is not None and order_by not in ("nan", "unique", "alphabetic"):
+            raise ValueError(
+                "order_by takes only values 'nan', 'unique', 'alphabetic', None."
+            )
```
solegalli, line 163: same here, could we add the f"Got {blabla} instead." please?

feature_engine/selection/drop_correlated_features.py
```diff
@@ -186,6 +212,14 @@ def fit(self, X: pd.DataFrame, y: pd.Series = None):
         # create tuples of correlated feature groups
         self.correlated_feature_sets_ = []
 
+        # sort columns for consistency
+        if self.order_by == "nan":
+            X = X[X.isna().sum(0).sort_values().index]
+        elif self.order_by == "unique":
+            X = X[X.nunique(0).sort_values(ascending=False).index]
+        elif self.order_by == "alphabetic":
+            X = X.sort_index(axis=1)
```
solegalli, line 221: I think it'd be good to have this in a bespoke hidden method `(_sort_variables(self)`, so that we can test just the ordering functionality without having to go through the whole fit transform process to get something useful.

solegalli (commented):
Hey @glevv 

Thank you for picking this one up!

I only have minor comments / suggestions.

And, could we also add this functionality to the smart correlation class? it shares the same consistency problems when it finds the correlated groups.

Thank you!

feature_engine/selection/drop_correlated_features.py
glevv: Std would be bad if:
 - you have features with different scales;
 - you have dtypes that pandas does not count as numerical and will ignore them;
In first case it won't be useful for the transformer, in second case it will fail with not very informative message. While sort by stdev sounds good, after tests I'm not a big fan of it.

All of this options would have drawbacks, the idea is to be consistent and transparent about in what order features are passed through elimination process.

glevv: For 1mil rows and 8 features:

`297 ms ± 3.33 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`

docs/user_guide/selection/DropCorrelatedFeatures.rst
```diff
@@ -18,6 +18,9 @@ The transformer will examine all numerical variables automatically. Note that yo
 dataframe with categorical and datetime variables, and these will be ignored automatically.
 Alternatively, you can pass a list with the variables you wish to evaluate.
 
+Note: ordering of the coulmns matters! Different order of the columns could lead to
+different results. To make estimator more consistent, please use 'order_by' parameter.
```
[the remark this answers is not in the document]
glevv: Done

feature_engine/selection/drop_correlated_features.py
solegalli: Thank you. 

Yes, I got up thinking that also if you use the standard scaler, then the std should be close to 1 for all vars so dies not make a lot of sense either.

solegalli:
We still need to extend this functionality to the smart correlation selector to merge

solegalli:
See #648

PR #636, opened 2023-03-14
unnest test files of categorical encoders

### LINKED ISSUE #629: Nested directory structure of tests and feature_engine itself
What was the rationale behind nested directory structure of tests?

I think if we flatten that hierarchy newcomer developers will benefit by navigating easily. Also, for me it's easy to get lost inside that hierarchy.

Here is the screenshot of that nested hierarchy:
image

Here, the module we're testing is **feature_engine/encoding** which itself is not nested. If we flatten that hierarchy at least it will resemble other directories.

Regarding the feature_engine there is **timeseries** module with nested directory. All other modules don't have that sort of structure and there is only this one with nested structure.

Here is the screenshot:

image

What about flattening this too?

solegalli:
closes #629 

FYI @Okroshiashvili 

Would you agree with this structure?

Whenever I have a single file I have at the root of the test module. And when there are multiple test files per transformer I leave in folder.

let me know
Thank you!


Okroshiashvili:
@solegalli seems good for me

PR #643, opened 2023-03-19
Add code examples in preprocessing and wrappers modules' docstrings

datacubeR:
Added examples for the following modules (both in the same PR because they are just a few classes).

- Wrappers: `SklearnTransformerWrapper`
-  Preprocessing: `MatchCategories` and `MatchVariables`.

### DIFF
--- a/feature_engine/preprocessing/match_columns.py
+++ b/feature_engine/preprocessing/match_columns.py
@@ -127,11 +127,9 @@ class MatchVariables(BaseEstimator, TransformerMixin, GetFeatureNamesOutMixin):
 
     >>> import pandas as pd
     >>> from feature_engine.preprocessing import MatchVariables
-
     >>> X_train = pd.DataFrame(dict(x1 = ["a","b","c"],
     >>>                             x2 = [4,5,6], x3 = [1,1,1]))
     >>> X_test = pd.DataFrame(dict(x1 = ["c","b","a","d"], x2 = [5,6,4,7]))
-
     >>> mv = MatchVariables(missing_values="ignore")
     >>> mv.fit(X_train)
     >>> mv.transform(X_train)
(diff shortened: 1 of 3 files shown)


feature_engine/preprocessing/match_columns.py
```diff
@@ -100,6 +100,52 @@ class MatchVariables(BaseEstimator, TransformerMixin, GetFeatureNamesOutMixin):
...
+    The following variables are dropped from the DataFrame: ['x3']
+      x1  x2
+    0  c   5
+    1  b   6
+    2  a   4
+    3  d   7
+
+    >>> import pandas as pd
+    >>> from feature_engine.preprocessing import MatchVariables
+
```
solegalli, line 130: can we remove this space?

feature_engine/wrappers/wrappers.py
```diff
@@ -144,6 +144,45 @@ class SklearnTransformerWrapper(BaseEstimator, TransformerMixin):
...
+    >>> skw.transform(X)
+       x2  x1_a  x1_b  x1_c
+    0   4   1.0   0.0   0.0
+    1   5   0.0   1.0   0.0
+    2   6   0.0   0.0   1.0
+
+    >>> import pandas as pd
+    >>> from feature_engine.wrappers import SklearnTransformerWrapper
+    >>> from sklearn.preprocessing import PolynomialFeatures
+    >>> skw = SklearnTransformerWrapper(PolynomialFeatures(include_bias = False))
```
solegalli, line 179: the examples are meant for the user to be able to copy and paste and reproduce, so we need to regenerate X here. Also, it'd be nicer if we added 1 additional numerical variable to be able to make more combinations. But do keep the categorical one, because its the one that justifies the use of the wrapper in this example.

PR #644, opened 2023-03-19
Add code examples in outliers module's docstrings

datacubeR:
Added examples for the Outliers module:

- Winsorizer
- Arbitrary Outlier Capper 
- OutlierTrimmer

### DIFF
--- a/feature_engine/outliers/artbitrary.py
+++ b/feature_engine/outliers/artbitrary.py
@@ -91,6 +91,36 @@ class ArbitraryOutlierCapper(BaseOutlier):
     transform:
         Cap the variables.
 
+    Examples
+    --------
+
+    >>> import pandas as pd
+    >>> import numpy as np
+    >>> from feature_engine.outliers import ArbitraryOutlierCapper
+    >>> np.random.seed(42)
+    >>> X = pd.DataFrame(dict(x1 = np.random.exponential(size = 100),
+    >>>                     x2 = np.random.randn(100)))
+    >>> aoc = ArbitraryOutlierCapper(max_capping_dict={'x1': 3, 'x2': 1},
+    >>>                             min_capping_dict={'x2': 0.2})
+    >>> print("Max value prior Arbitrary Capper:")
+    >>> X.max()
+    Max value prior Arbitrary Capper:
+    x1    4.334146
+    x2    2.463242
+    dtype: float64
+    >>> aoc.fit(X)
+    >>> print("Max value after Arbitrary Capper:")
+    >>> aoc.transform(X).max()
+    Max value after Arbitrary Capper:
+    x1    3.0
+    x2    1.0
+    dtype: float64
+    >>> print("Min value after Arbitrary Capper:")
+    >>> aoc.transform(X).min()
+    Min value after Arbitrary Capper:
+    x1    0.005537
+    x2    0.200000
+    dtype: float64
     """
 
     def __init__(
--- a/feature_engine/outliers/artbitrary.py
+++ b/feature_engine/outliers/artbitrary.py
@@ -95,32 +95,12 @@ class ArbitraryOutlierCapper(BaseOutlier):
     --------
 
     >>> import pandas as pd
-    >>> import numpy as np
     >>> from feature_engine.outliers import ArbitraryOutlierCapper
-    >>> np.random.seed(42)
-    >>> X = pd.DataFrame(dict(x1 = np.random.exponential(size = 100),
-    >>>                     x2 = np.random.randn(100)))
-    >>> aoc = ArbitraryOutlierCapper(max_capping_dict={'x1': 3, 'x2': 1},
-    >>>                             min_capping_dict={'x2': 0.2})
-    >>> print("Max value prior Arbitrary Capper:")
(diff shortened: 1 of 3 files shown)


datacubeR:
@solegalli why are these tests failing if I haven't modified anything on the source code or the tests?

solegalli:
It fails because the docs will look to replace anything that is in between {} by a variable that we specify. 

If you want to include {} in the code example within the docstrings, you need to write it outside as I do here:

https://github.com/feature-engine/feature_engine/blob/c2f61524cc404cfb661d10345cfbe6e7aa3608bf/feature_engine/datetime/datetime_subtraction.py#L32-L45

And then call it by using the substute function like i do here:

https://github.com/feature-engine/feature_engine/blob/c2f61524cc404cfb661d10345cfbe6e7aa3608bf/feature_engine/datetime/datetime_subtraction.py#L48-L55

And then pass the example in the corresponding part like here:

https://github.com/feature-engine/feature_engine/blob/c2f61524cc404cfb661d10345cfbe6e7aa3608bf/feature_engine/datetime/datetime_subtraction.py#L142

feature_engine/outliers/artbitrary.py
```diff
@@ -91,6 +91,36 @@ class ArbitraryOutlierCapper(BaseOutlier):
...
+    Examples
+    --------
+
+    >>> import pandas as pd
+    >>> import numpy as np
+    >>> from feature_engine.outliers import ArbitraryOutlierCapper
+    >>> np.random.seed(42)
+    >>> X = pd.DataFrame(dict(x1 = np.random.exponential(size = 100),
+    >>>                     x2 = np.random.randn(100)))
+    >>> aoc = ArbitraryOutlierCapper(max_capping_dict={'x1': 3, 'x2': 1},
```
solegalli, line 103: these {} make the docs fail

feature_engine/outliers/artbitrary.py
```diff
@@ -91,6 +91,36 @@ class ArbitraryOutlierCapper(BaseOutlier):
...
 
+    Examples
+    --------
+
+    >>> import pandas as pd
+    >>> import numpy as np
+    >>> from feature_engine.outliers import ArbitraryOutlierCapper
+    >>> np.random.seed(42)
+    >>> X = pd.DataFrame(dict(x1 = np.random.exponential(size = 100),
+    >>>                     x2 = np.random.randn(100)))
```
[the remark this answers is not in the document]
solegalli: Since we are using the arbitrary capper, I think it is more illustrative if we create a toy dataset with values 1,2,3,4... to 10. And when cap at 2 and 8. And show the transformed dataframe. Could we change the example?

feature_engine/outliers/artbitrary.py
```diff
@@ -91,6 +91,36 @@ class ArbitraryOutlierCapper(BaseOutlier):
...
+    >>> aoc = ArbitraryOutlierCapper(max_capping_dict={'x1': 3, 'x2': 1},
+    >>>                             min_capping_dict={'x2': 0.2})
+    >>> print("Max value prior Arbitrary Capper:")
```
solegalli, line 105: I'd remove all the intermediate prints and just show the resulting dataframe after capping at 2 and 8

feature_engine/outliers/trimmer.py
```diff
@@ -89,6 +89,31 @@ class OutlierTrimmer(WinsorizerBase):
     transform:
         Remove outliers.
 
+    Examples
+    --------
+
+    >>> import pandas as pd
+    >>> import numpy as np
```
solegalli, line 96: picky: import numpy first

feature_engine/outliers/trimmer.py
```diff
@@ -89,6 +89,31 @@ class OutlierTrimmer(WinsorizerBase):
...
+    >>> from feature_engine.outliers import OutlierTrimmer
+    >>> np.random.seed(42)
+    >>> X = pd.DataFrame(dict(x = np.random.exponential(size = 100)))
+    >>> ot = OutlierTrimmer(capping_method='gaussian', tail='right', fold=3)
```
solegalli, line 100: If we are using gaussian, I think it would be better to use np.random.normal instead of exponential.

These snippets are meant for the readers to be able to copy and paste into their code. We should just demo how to set the transformer and then how to fit it and transform data, showing the resulting dataframe.

I'd remove all the intermediate prints. For more details we have a user guide that users can read.

I'd also use a simpler data for the demo, with 10 values, instead of a random extraction of 100 values. Could we update the example ?

feature_engine/outliers/winsorizer.py
```diff
@@ -97,6 +97,34 @@ class Winsorizer(WinsorizerBase):
...
         Cap the variables.
 
+    Examples
+    --------
+
+    >>> import pandas as pd
+    >>> import numpy as np
+    >>> from feature_engine.outliers import Winsorizer
+    >>> np.random.seed(42)
+    >>> X = pd.DataFrame(dict(x = np.random.exponential(size = 100)))
```
solegalli, line 107: Same as previous

PR #646, opened 2023-03-20
Add code examples in transformation module's docstrings

datacubeR:
Adding examples for:

- Log and LogCp Transform
- Reciprocal Transform
- Arcsin Transform
- Power Transform
- BoxCox Transform
- YeoJohnson Transform

### DIFF
--- a/feature_engine/transformation/arcsin.py
+++ b/feature_engine/transformation/arcsin.py
@@ -80,6 +80,36 @@ class ArcsinTransformer(BaseNumericalTransformer):
     transform:
         Apply the arcsin transformation.
 
+    Examples
+    --------
+
+    >>> import pandas as pd
+    >>> import numpy as np
+    >>> from feature_engine.transformation import ArcsinTransformer
+    >>> np.random.seed(42)
+    >>> X = pd.DataFrame(dict(x = np.random.beta(1, 1, size = 100)))
+    >>> X.describe()
+                    x
+    count  100.000000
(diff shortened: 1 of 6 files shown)


solegalli (commented):
Hey @datacubeR 

Thank you so much for these examples. I think these are almost ready.

I'd like to remove the .describe() and just show the head of the transformed dataframe as output. Could we change that throughout all examples. So they'd look like this more or less:

```
    >>> import pandas as pd
    >>> import numpy as np
    >>> from feature_engine.transformation import PowerTransformer
    >>> np.random.seed(42)
    >>> X = pd.DataFrame(dict(x = np.random.lognormal(size = 100)))
    >>> pt = PowerTransformer()
    >>> pt.fit(X)
    >>> X = pt.transform(X)
    >>> X.head()
```
And then the output.

Also, could we first import numpy and then pandas? as per pep8?

Finally, in the yeojohson example, could we remove the scaler and the wrapper? just showcase the transformer in question.

We've got user guides with more details elsewhere.

Thank you!

datacubeR:
> Finally, in the yeojohson example, could we remove the scaler and the wrapper? just showcase the transformer in question.

Sure thing. I added Scaler in a pipeline because with no scaling the YeoJohnson transform tends to return giant values that seems wrong (but they are not). I looked into the sklearn docs and their transformer have an `standardized = True` flag to deal with this, so I  thought it could be a good idea to make results comparable.

PR #657, opened 2023-04-18
Extend PSI feature selection to categorical variables

### LINKED ISSUE #655: PSI feature selection for categorical features
Currently, PSI feature selection as implemented in selection/drop_psi_features.py can only be applied to numerical features: it first converts these numerical features to categoricals via bucketing, then calculates PSI for the discretised numerical features. 

It would be useful to also be able to use DropHighPSIFeatures on categorical features. I believe the implementation is fairly straightforward.

I'd be happy to pick up this issue if it is of interest to others.

dlaprins:
closes #655 
closes #658

### DIFF
--- a/feature_engine/selection/drop_psi_features.py
+++ b/feature_engine/selection/drop_psi_features.py
@@ -145,7 +145,15 @@ class DropHighPSIFeatures(BaseSelector):
         feature will be dropped. The most common threshold values are 0.25 (large shift)
         and 0.10 (medium shift).
         If 'auto', the threshold will be calculated based on the size of the base and
-        target dataset and the number of bins.
+        target dataset and the number of bins as
+                threshold = χ2(q, B−1) × (1/N + 1/M)
+        where
+        q = quantile of the distribution (or 1 - p-value),
+        B = number of bins/categories,
+        N = size of basis dataset,
+        M = size of test dataset.
+        See formula (5.2) from reference [1] in the class doc string.
+
 
     bins: int, default = 10
         Number of bins or intervals. For continuous features with good value spread, 10
@@ -170,7 +178,7 @@ class DropHighPSIFeatures(BaseSelector):
         when determining the PSI for that particular feature. If 'raise' the transformer
         will raise an error and features will not be selected.
 
-    p_value: float
+    p_value: float, default = 0.001
         This variable determines the p-value to test the null hypothesis that there is
         no feature drift: in that case, the PSI-value approximates a random variable
         that follows a chi-square distribution. See [1] for details. The variable is
@@ -381,7 +389,7 @@ class DropHighPSIFeatures(BaseSelector):
         if self.missing_values == "raise":
             # check if dataset contains na or inf
             _check_contains_na(X, self.variables_)
-            _check_contains_inf(X, self.num_variables_)
+            _check_contains_inf(X, num_variables_)
 
         # Split the dataframe into basis and test.
         basis_df, test_df = self._split_dataframe(X)
@@ -404,7 +412,7 @@ class DropHighPSIFeatures(BaseSelector):
             test_df, basis_df = basis_df, test_df
 
         # set up the discretizer for numerical variables
-        if len(self.num_variables_) > 0:
+        if len(num_variables_) > 0:
             if self.strategy == "equal_width":
                 bucketer = EqualWidthDiscretiser(bins=self.bins)
             else:
@@ -416,11 +424,11 @@ class DropHighPSIFeatures(BaseSelector):
 
         for feature in self.variables_:
             # Bin the features if it is numerical and determine number of bins
-            if feature in self.num_variables_:
+            if feature in num_variables_:
                 basis_discrete = bucketer.fit_transform(basis_df[[feature]].dropna())
                 test_discrete = bucketer.transform(test_df[[feature]].dropna())
                 n_bins = self.bins
-            elif feature in self.cat_variables_:
+            elif feature in cat_variables_:
                 basis_discrete = basis_df[[feature]]
                 test_discrete = test_df[[feature]]
                 n_bins = X[feature].nunique()
--- a/feature_engine/selection/drop_psi_features.py
+++ b/feature_engine/selection/drop_psi_features.py
@@ -408,8 +408,6 @@ class DropHighPSIFeatures(BaseSelector):
             basis_distrib, test_distrib = self._observation_frequency_per_bin(
                 basis_discrete, test_discrete
             )
-            print("BASIS DIST:", basis_distrib)
-            print("TEST DISTRIB:", test_distrib)
 
             # Calculate the PSI value
             self.psi_values_[feature] = np.sum(
--- a/feature_engine/selection/drop_psi_features.py
+++ b/feature_engine/selection/drop_psi_features.py
@@ -34,7 +34,7 @@ from feature_engine.variable_handling._init_parameter_checks import (
     _check_init_parameter_variables,
 )
 from feature_engine.variable_handling.variable_type_selection import (
-    find_or_check_numerical_variables,
+    find_categorical_and_numerical_variables
 )
 
 Variables = Union[None, int, str, List[Union[str, int]]]
@@ -251,6 +251,7 @@ class DropHighPSIFeatures(BaseSelector):
         missing_values: str = "raise",
         variables: Variables = None,
         confirm_variables: bool = False,
+        p_value: float = 0.001,
     ):
 
         if not isinstance(split_col, (str, int, type(None))):
@@ -335,6 +336,7 @@ class DropHighPSIFeatures(BaseSelector):
         self.strategy = strategy
         self.min_pct_empty_bins = min_pct_empty_bins
         self.missing_values = missing_values
+        self.p_value = p_value
 
     def fit(self, X: pd.DataFrame, y: pd.Series = None):
         """
@@ -354,8 +356,12 @@ class DropHighPSIFeatures(BaseSelector):
         # If required exclude variables that are not in the input dataframe
         self._confirm_variables(X)
 
-        # find numerical variables or check those entered are present in the dataframe
-        self.variables_ = find_or_check_numerical_variables(X, self.variables_)
+        # # find variables or check those entered are present in the dataframe
+        self.cat_variables_,  self.num_variables_ = find_categorical_and_numerical_variables(X, self.variables_)
+        self.variables_ = self.num_variables_ + self.cat_variables_
+        print("NUM VARIABLES ARE: ", self.num_variables_)
+        print("CAT VARIABLES ARE: ", self.cat_variables_)
+        print("VARIABLES ARE: ", self.variables_)
 
         # Remove the split_col from the variables list. It might be added if the
         # variables are not defined at initialization.
@@ -365,7 +371,7 @@ class DropHighPSIFeatures(BaseSelector):
         if self.missing_values == "raise":
             # check if dataset contains na or inf
             _check_contains_na(X, self.variables_)
-            _check_contains_inf(X, self.variables_)
+            _check_contains_inf(X, self.num_variables_)
 
         # Split the dataframe into basis and test.
         basis_df, test_df = self._split_dataframe(X)
@@ -387,28 +393,29 @@ class DropHighPSIFeatures(BaseSelector):
         if self.switch:
             test_df, basis_df = basis_df, test_df
 
-        if self.threshold == "auto":
-            threshold = self._calculate_auto_threshold(
-                basis_df.shape[0], test_df.shape[0]
-            )
-        else:
-            threshold = self.threshold
 
-        # set up the discretizer
-        if self.strategy == "equal_width":
-            bucketer = EqualWidthDiscretiser(bins=self.bins)
-        else:
-            bucketer = EqualFrequencyDiscretiser(q=self.bins)
+
+        # set up the discretizer for numerical variables
+        if len(self.num_variables_) > 0: 
+            if self.strategy == "equal_width":
+                bucketer = EqualWidthDiscretiser(bins=self.bins)
+            else:
+                bucketer = EqualFrequencyDiscretiser(q=self.bins)
 
         # Compute the PSI by looping over the features
         self.psi_values_ = {}
         self.features_to_drop_ = []
 
         for feature in self.variables_:
-            # Discretize the features.
-
-            basis_discrete = bucketer.fit_transform(basis_df[[feature]].dropna())
-            test_discrete = bucketer.transform(test_df[[feature]].dropna())
+            # Bin the features if it is numerical and determine number of bins
+            if feature in self.num_variables_:
+                basis_discrete = bucketer.fit_transform(basis_df[[feature]].dropna())
+                test_discrete = bucketer.transform(test_df[[feature]].dropna())
+                n_bins = self.bins
+            elif feature in self.cat_variables_: 
+                basis_discrete = basis_df[[feature]]
+                test_discrete = test_df[[feature]]
+                n_bins = X[[feature]].nunique()[0]
 
             # Determine percentage of observations per bin
             basis_distrib, test_distrib = self._observation_frequency_per_bin(
@@ -419,6 +426,14 @@ class DropHighPSIFeatures(BaseSelector):
             self.psi_values_[feature] = np.sum(
                 (test_distrib - basis_distrib) * np.log(test_distrib / basis_distrib)
             )
+
+            # Determine the appropriate threshold
+            if self.threshold == "auto":
+                threshold = self._calculate_auto_threshold(
+                    basis_df.shape[0], test_df.shape[0], n_bins,
+                )
+            else:
+                threshold = self.threshold
             # Assess if feature should be dropped
             if self.psi_values_[feature] > threshold:
                 self.features_to_drop_.append(feature)
@@ -577,10 +592,16 @@ class DropHighPSIFeatures(BaseSelector):
 
         return cut_off
 
-    def _calculate_auto_threshold(self, N, M, q=0.999):
-        # threshold = χ2(q,B−1) × (1/N + 1/M)
-        # where q - quantile (or 1 - p-value) B - number of bins,
-        # N - size of basis dataset, M - size of test dataset
-        # see formula (5.2) from reference
-        # taking q = 0.999 to get higher threshold
-        return stats.chi2.ppf(q, self.bins - 1) * (1.0 / N + 1.0 / M)
+    def _calculate_auto_threshold(self, N, M, bins):
+        """ Threshold computation for chi-square test.
+
+        The threshold is given by 
+            threshold = χ2(q,B−1) × (1/N + 1/M)
+        where 
+        q = quantile of the distribution (or 1 - p-value), 
+        B = number of bins/categories, 
+        N = size of basis dataset, 
+        M = size of test dataset. 
+        See formula (5.2) from reference.
+        """
+        return stats.chi2.ppf(1 - self.p_value, bins - 1) * (1.0 / N + 1.0 / M)
--- a/feature_engine/selection/drop_psi_features.py
+++ b/feature_engine/selection/drop_psi_features.py
@@ -411,27 +411,40 @@ class DropHighPSIFeatures(BaseSelector):
         if self.switch:
             test_df, basis_df = basis_df, test_df
 
-        # set up the discretizer for numerical variables
+        # Set up parameters for numerical features
         if len(num_variables_) > 0:
+            n_bins_num = self.bins
+
+            # Set up the discretizer for numerical features
             if self.strategy == "equal_width":
                 bucketer = EqualWidthDiscretiser(bins=self.bins)
             else:
                 bucketer = EqualFrequencyDiscretiser(q=self.bins)
 
+            # Set up the threshold for numerical features
+            if self.threshold == "auto":
+                threshold_num = self._calculate_auto_threshold(
+                    basis_df.shape[0],
+                    test_df.shape[0],
+                    n_bins_num,
+                )
+            else:
+                threshold_num = self.threshold
+
+        # Set up the generic threshold for categorical features if used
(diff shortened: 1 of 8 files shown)


ClaudioSalvatoreArcidiacono:
Hey @dlaprins ping me once this PR is ready for review!

feature_engine/selection/drop_psi_features.py
```diff
@@ -7,42 +7,31 @@
...
-    _check_init_parameter_variables,
-)
-from feature_engine.variable_handling.variable_type_selection import (
-    find_or_check_numerical_variables,
-)
+from feature_engine.tags import _return_tags
+from feature_engine.variable_handling._init_parameter_checks import \
+    _check_init_parameter_variables
+from feature_engine.variable_handling.variable_type_selection import \
+    find_categorical_and_numerical_variables
```
ClaudioSalvatoreArcidiacono, line 27: Could you format this with `black`?

```
black feature_engine/selection/drop_psi_features.py
```

You should first run `isort`, then `black`. Not the other way around

feature_engine/selection/drop_psi_features.py
```diff
@@ -354,8 +345,12 @@ def fit(self, X: pd.DataFrame, y: pd.Series = None):
         # If required exclude variables that are not in the input dataframe
         self._confirm_variables(X)
 
-        # find numerical variables or check those entered are present in the dataframe
-        self.variables_ = find_or_check_numerical_variables(X, self.variables_)
+        # # find variables or check those entered are present in the dataframe
```
ClaudioSalvatoreArcidiacono, line 348:
```suggestion
        # find variables or check those entered are present in the dataframe
```

feature_engine/selection/drop_psi_features.py
```diff
@@ -387,28 +382,27 @@ def fit(self, X: pd.DataFrame, y: pd.Series = None):
...
-            test_discrete = bucketer.transform(test_df[[feature]].dropna())
+            # Bin the features if it is numerical and determine number of bins
+            if feature in self.num_variables_:
+                basis_discrete = bucketer.fit_transform(basis_df[[feature]].dropna())
+                test_discrete = bucketer.transform(test_df[[feature]].dropna())
+                n_bins = self.bins
+            elif feature in self.cat_variables_:
+                basis_discrete = basis_df[[feature]]
+                test_discrete = test_df[[feature]]
+                n_bins = X[[feature]].nunique()[0]
```
ClaudioSalvatoreArcidiacono, line 405:
```suggestion
                n_bins = X[feature].nunique()
```

feature_engine/selection/drop_psi_features.py
```diff
@@ -354,8 +345,12 @@ def fit(self, X: pd.DataFrame, y: pd.Series = None):
         # If required exclude variables that are not in the input dataframe
         self._confirm_variables(X)
 
-        # find numerical variables or check those entered are present in the dataframe
-        self.variables_ = find_or_check_numerical_variables(X, self.variables_)
+        # # find variables or check those entered are present in the dataframe
+        (
+            self.cat_variables_,
+            self.num_variables_,
```
ClaudioSalvatoreArcidiacono, line 351: Is it necessary for these two variables to be a new instance attribute? How about keeping them as simple variables?

feature_engine/selection/drop_psi_features.py
```diff
@@ -577,10 +582,24 @@ def _get_cut_off_value(self, split_column):
...
+        """Threshold computation for chi-square test.
+
+        The threshold is given by
+            threshold = χ2(q,B−1) × (1/N + 1/M)
+        where
+        q = quantile of the distribution (or 1 - p-value),
+        B = number of bins/categories,
+        N = size of basis dataset,
+        M = size of test dataset.
+        See formula (5.2) from reference.
```
ClaudioSalvatoreArcidiacono, line 595: I know this comes from the previous version of docstring, but is this reference actually valid? Does It refer to something in the class docstring? If it is referring to outside resources, like feature engine documentation, isn't it better to refer with a link?

feature_engine/selection/drop_psi_features.py
```diff
@@ -577,10 +582,24 @@ def _get_cut_off_value(self, split_column):
...
 
-    def _calculate_auto_threshold(self, N, M, q=0.999):
-        # threshold = χ2(q,B−1) × (1/N + 1/M)
-        # where q - quantile (or 1 - p-value) B - number of bins,
-        # N - size of basis dataset, M - size of test dataset
-        # see formula (5.2) from reference
-        # taking q = 0.999 to get higher threshold
-        return stats.chi2.ppf(q, self.bins - 1) * (1.0 / N + 1.0 / M)
+    def _calculate_auto_threshold(self, N, M, bins):
+        """Threshold computation for chi-square test.
```
ClaudioSalvatoreArcidiacono, line 586: could you also write docstring for the arguments?

feature_engine/selection/drop_psi_features.py
```diff
@@ -251,6 +240,7 @@ def __init__(
         missing_values: str = "raise",
         variables: Variables = None,
         confirm_variables: bool = False,
+        p_value: float = 0.001,
```
ClaudioSalvatoreArcidiacono, line 243: Docstring for this new parameter is missing.

ClaudioSalvatoreArcidiacono, line 243: I would suggest to add this parameter on a different order, for example after threshold, to show they are more closely related. Moreover usually feature engine selector ends with the parameters `variables` and `confirm_variables`.

ClaudioSalvatoreArcidiacono: Not sure about this comment, as adding it at the end works better with backwards compatibility. I think the maintainers should have the final say on it.

feature_engine/selection/drop_psi_features.py
solegalli: Yep, I second this one. Not sure what happened here, but if there are no new imports, I'd prefer we don't change what was originally there. It is also easier to review.

Could we fix @dlaprins ?

feature_engine/selection/drop_psi_features.py
solegalli: Interesting discussion. But if we put it after threshold, users that are already familiar with the order of the params will be frustrated. Let's leave it for the time being.

feature_engine/selection/drop_psi_features.py
```diff
@@ -335,6 +325,7 @@ def __init__(
         self.strategy = strategy
         self.min_pct_empty_bins = min_pct_empty_bins
         self.missing_values = missing_values
+        self.p_value = p_value
```
solegalli, line 328: we'd need a check for this value, that is a float within so and so.

feature_engine/selection/drop_psi_features.py
solegalli: Please fix as per @ClaudioSalvatoreArcidiacono 's suggestion

feature_engine/selection/drop_psi_features.py
solegalli: I guess @dlaprins  is copying from what we do in another selector that works with both numerical and categorical.

We probably do not need to generate these new attributes. We could drop it. But if we do, then the user would not be able to easily find out which variables the transformer considered numerical and which ones categorical.

Simple controversial example: discrete variables encoded as object, will be considered categorical. Will the user be aware of this? 

I am not sure what is best here. I appreciate your input.

feature_engine/selection/drop_psi_features.py
```diff
@@ -387,28 +382,27 @@ def fit(self, X: pd.DataFrame, y: pd.Series = None):
...
+            if self.strategy == "equal_width":
+                bucketer = EqualWidthDiscretiser(bins=self.bins)
+            else:
+                bucketer = EqualFrequencyDiscretiser(q=self.bins)
 
         # Compute the PSI by looping over the features
         self.psi_values_ = {}
         self.features_to_drop_ = []
 
         for feature in self.variables_:
```
solegalli, line 396: instead of looping in `self.variables_` and then `self.numerical_` we should just loop over numerical and then over categorical, and, like this, remove the nested looping.

Could we fix that please?

feature_engine/selection/drop_psi_features.py
solegalli: pls fix as per suggestion :)

feature_engine/selection/drop_psi_features.py
solegalli: Why are we changing the default parameters of the auto-threshold? We used to default q to 0.999 and now this is an obligatory parameter.

What's the logic?

fyi @glevv

feature_engine/selection/drop_psi_features.py
solegalli: This is a hidden method, so it won't be exposed to users. It is mostly for developers. If it is not clear which reference it refers to, then we should spell it out though.

feature_engine/selection/drop_psi_features.py
```diff
@@ -464,6 +468,7 @@ def _observation_frequency_per_bin(self, basis, test):
                 how="outer",
             )
             .fillna(self.min_pct_empty_bins)
+            .replace(to_replace=0, value=self.min_pct_empty_bins)
```
solegalli, line 471: why do we add this line?

feature_engine/selection/drop_psi_features.py
solegalli: Ok, I see that p_value comes here, that's why we change the logic of the auto-threshold. We need to update the dosctrings.

@glevv could you review this bit please?

solegalli (commented):
Hey @dlaprins and @ClaudioSalvatoreArcidiacono 

Thank you so much for the active discussions and code changes!.

I added my comments at the back of @ClaudioSalvatoreArcidiacono 's

@dlaprins will you be able to update the code implementation?

Thanks a lot!

feature_engine/selection/drop_psi_features.py
ClaudioSalvatoreArcidiacono: I see now that saving categorical and numerical variables as instance attributes have been done in [_prediction/base_predictor.py](https://github.com/feature-engine/feature_engine/tree/main/feature_engine/_prediction#L12) indeed. It can make sense to save them as instance attributes for that purpose, but in that case I would comment that documentation for the new instance attributes is missing.

On the other hand, in the [information value selector](https://github.com/feature-engine/feature_engine/blob/main/feature_engine/selection/information_value.py#L217), those are not saved.

I personally think that both approaches can be correct.

feature_engine/selection/drop_psi_features.py
```diff
@@ -387,28 +382,27 @@ def fit(self, X: pd.DataFrame, y: pd.Series = None):
...
-        else:
-            threshold = self.threshold
-
-        # set up the discretizer
-        if self.strategy == "equal_width":
-            bucketer = EqualWidthDiscretiser(bins=self.bins)
-        else:
-            bucketer = EqualFrequencyDiscretiser(q=self.bins)
+        # set up the discretizer for numerical variables
+        if len(self.num_variables_) > 0:
```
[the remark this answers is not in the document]
ClaudioSalvatoreArcidiacono: Using PSI if there are only categorical features is perfectly fine and correct, so in my opinion, I would not raise an error or a warning in case there are no numerical values.

feature_engine/selection/drop_psi_features.py
solegalli: Let's not save them. @dlaprins could we update the code please?

feature_engine/selection/drop_psi_features.py
dlaprins: Thanks for letting me know,  I was unaware of the correct ordering.

feature_engine/selection/drop_psi_features.py
dlaprins: Will do

feature_engine/selection/drop_psi_features.py
dlaprins: The reference is to reference [1] in the docstring of the class (i.e., Yurdakul (2018, https://scholarworks.wmich.edu/dissertations/3208/). I have left it as is, let me know if you want to see the reference spelled out more explicitly

feature_engine/selection/drop_psi_features.py
dlaprins: If a categorical variable is of type 'object', and the split misses a category, the dataframe will have a NaN which is then solved by .fillna. However, in case the categorical variable is of type 'category', and the split misses a category, it will be correctly put to 0: thus it won't get filled and the PSI becomes infinite. 

This line ensures that empty buckets have a minimal value such that the PSI doesn't explode, similar to how it functions for 'object' type categoricals.

feature_engine/selection/drop_psi_features.py
dlaprins: Wrote some form of docstring for the arguments, but not sure if it's in the format desired. Please let me know if you want it to be something else. 

The logic behind changing the arguments is that the auto-threshold is called on by the class, but the p-value (or quantile, as previously used) could not be defined by the user. I would deem this to be a necessary option in order to make use of the auto-threshold (which I think is very useful in and of itself). 

The bins argument needs to be added for auto-threshold to be compatible with categorical features. For details, please see my discussion with Claudio in the Issues thread.

dlaprins:
Thank you both @ClaudioSalvatoreArcidiacono @solegalli  for your helpful suggestions and directions. I have tried to fix all the shortcomings you pointed out in the comments in my commit. Please let me know which issues remain unresolved and I'll solve those as well.

solegalli (commented):
Hey @dlaprins 

Thanks for the code changes. I went over the code and discussion, here are some more thoughts:

- [ ] we are now not capturing numerical and categorical variables in new attributes, that's great. Later on in the logic we are still looping over self.num_variables_ even though it does not exist any more. Could you please fix?
- [ ] It is not necessary to loop over self.variables_ we can remove the nested looping. 
- [ ] n_bins is being reassigned for each feature, this is not necessary for numerical, only for categorical. Could we update?
- [ ] the threshold is calculated once for all variables, I think the idea is to use the same threshold for numerical, and a changing threshold for categorical. Could we reflect this in the code?
- [ ] The question still remains: should we expose p_values?
- [ ] it would be great if we could add the threshold calculation formula in the init docstrings as well.
- [ ] @glevv mentioned that the calculation would not work for this example: current implementation will give erroneous results if you have some differences in categories in the same columns (cat_1_train have categories 'A', 'B' and 'C', and cat_1_test have categories 'A', 'B' and 'D'). Could we add that as a test and make sure it works somehow? or raises an exception to alert users on what to do?

Thank you for the great work!

feature_engine/selection/drop_psi_features.py
```diff
@@ -419,6 +434,16 @@ def fit(self, X: pd.DataFrame, y: pd.Series = None):
             self.psi_values_[feature] = np.sum(
                 (test_distrib - basis_distrib) * np.log(test_distrib / basis_distrib)
             )
+
+            # Determine the appropriate threshold
+            if self.threshold == "auto":
```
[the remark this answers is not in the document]
dlaprins: Hi @solegalli,

Let me try to address all the comments (including the other ones you left) in this one reply: 

> Hey @dlaprins
> 
> Thanks for the code changes. I went over the code and discussion, here are some more thoughts:
> 
>     * [ ]  we are now not capturing numerical and categorical variables in new attributes, that's great. Later on in the logic we are still looping over self.num_variables_ even though it does not exist any more. Could you please fix?

An oversight indeed.

>     * [ ]  It is not necessary to loop over self.variables_ we can remove the nested looping.

I'm a bit confused here. What do you mean by nested looping? As it stands, I believe there is a single for-loop over self.variables and only an if-statement inside of that for-loop, no other for-loops (for what it's worth: the for-loop over self.variables is necessary due to the dropna(), I tried vectorizing the for-loop before I noticed this and got wrong results). 

 >     * [ ]  n_bins is being reassigned for each feature, this is not necessary for numerical, only for categorical. Could we update?

The only way I see to update this is to use 2 loops, one over categoricals and one over numericals. Then n_bins doesn't have to be checked for each feature in the numerical loop. Is that what you have in mind? 

>     * [ ]  the threshold is calculated once for all variables, I think the idea is to use the same threshold for numerical, and a changing threshold for categorical. Could we reflect this in the code?

The threshold calculation is inside the loop over self.variables. Similar to n_bins, we can have a single numerical threshold outside of a for-loop over numericals, while for categoricals the threshold has to be calculated inside of the for-loop. 

>     * [ ]  The question still remains: should we expose p_values?

I believe we should. As a user I would find it rather unpleasant if I cannot set my own p-value threshold for a statistical test. It's fine if a user does not know what p_value they should set, as there's a default value for the p-value that they can use. 

>     * [ ]  it would be great if we could add the threshold calculation formula in the init docstrings as well.

 Will do

>     * [ ]  @glevv mentioned that the calculation would not work for this example: current implementation will give erroneous results if you have some differences in categories in the same columns (cat_1_train have categories 'A', 'B' and 'C', and cat_1_test have categories 'A', 'B' and 'D'). Could we add that as a test and make sure it works somehow? or raises an exception to alert users on what to do?

Is this really the case? I believe this is tested for by Claudio's test (see line 35, 85, 100 of test_drop_high_psi_features.py): "drift_cat_1" is exactly such that basis_test contains only "A" while test_df contains only "B". This is also why I got errors before modifying _observation_frequency_per_bin() when I changed the "drift_cat_1" feature in the test to be a categorical feature. Please correct me if I'm wrong here. 

> 
> Thank you for the great work!

Thank you for the suggestions on how to improve the code. 

I have left the loop over self.variables as it is for now, but if you would prefer to have 2 loops instead (or if I'm just not understanding you properly), let me know and I'll correct it.

feature_engine/selection/drop_psi_features.py
dlaprins: I fixed the reference as per your suggestion. 

It is not necessary for the user to set the p-value, but it is optional, so in that sense a user who just wants to run it without thinking about parameters can still do so with the auto-setting. I am of the opinion that it would be a very useful feature to be able to set your own p-value, especially if one has to justify such values to stakeholders. The functionality of `auto' threshold versus float threshold are not identical, as the former varies per number of categories whereas the latter is identical for all features.

feature_engine/selection/drop_psi_features.py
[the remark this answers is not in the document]
glevv: Not sure about using PSI on categorical features is "perfectly fine". Chi2 is a better way of doing it.

feature_engine/selection/drop_psi_features.py
```diff
@@ -387,28 +403,27 @@ def fit(self, X: pd.DataFrame, y: pd.Series = None):
...
         for feature in self.variables_:
-            # Discretize the features.
-
-            basis_discrete = bucketer.fit_transform(basis_df[[feature]].dropna())
-            test_discrete = bucketer.transform(test_df[[feature]].dropna())
+            # Bin the features if it is numerical and determine number of bins
+            if feature in self.num_variables_:
+                basis_discrete = bucketer.fit_transform(basis_df[[feature]].dropna())
+                test_discrete = bucketer.transform(test_df[[feature]].dropna())
+                n_bins = self.bins
```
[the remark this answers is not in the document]
glevv: I agree, if psi selection is extended to categorical features, then we need to use list of auto thresholds depending on number of bins in these features.

feature_engine/selection/drop_psi_features.py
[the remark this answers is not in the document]
dlaprins: What's the reason for considering chi2 to be better than PSI?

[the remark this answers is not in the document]
solegalli: chi2 is very well studied and specifically designed to compare the frequency of categorical distributions (here more info: https://uk.sagepub.com/en-gb/eur/discovering-statistics-using-r/book236067). I think at some point I also read the chi2 article, but long forgotten.

I am also not sure I'd go and say "perfectly fine" to use PSI instead of chi2. Is this statement just based on that thesis? How well accepted is it in the community?

In my opinion, what happens is that if you make something available out of the box, like we are doing here, then less experienced users may be mislead to think that this is a proper way of assessing categorical variables. And I think, they should use chi2, based on my understanding of the consensus.

The thesis says that the PSI behaves like chi2, OK, it doesn't change the consensus the way I see it.

A follow up question could be, why does one use PSI instead of using other tests to compare continuous distributions, like the KolmogorovSmirnov test. I personally do not have the answer to that question, I haven't used the PSI much. So on that line, if we do discretize continuous variables to use the PSI, then why not apply it on categoricals as well, it seems like less work.

So in summary, what we need to decide is if we should we raise a warning if there are no numerical variables. And we should probably expand the user documentation with some discussion on chi2 vs PSI for categorical variables.

feature_engine/selection/drop_psi_features.py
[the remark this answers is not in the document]
solegalli: Hi @dlaprins yes, that was what I meant, 2 for loops instead of the for and the ifs. Sorry I wasn't clear.

Thank you for answering all the points!

feature_engine/selection/drop_psi_features.py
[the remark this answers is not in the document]
ClaudioSalvatoreArcidiacono: I think my comment had been misinterpreted. I did not want to say that it is perfectly fine to use one metric over the other. What I said is that it is perfectly fine to use PSI to measure feature drift over a dataset of categorical variables only and thus it is not needed to raise a warning in this situation.

solegalli:
FYI @dlaprins 

Please rebase main, we just merged #660 :)

feature_engine/selection/drop_psi_features.py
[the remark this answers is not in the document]
solegalli: Thank you!

dlaprins:
> FYI @dlaprins
> 
> Please rebase main, we just merged #660 :)

Fixed, as well as split up the loop over variables to 2 loops over cat variables and num variables separately as requested.

PR #658, opened 2023-04-18
Add support for categorical variables in PSI selector

ClaudioSalvatoreArcidiacono:
Fixes #655

PR #660, opened 2023-04-18
refactor code to work with pandas 2.0

ClaudioSalvatoreArcidiacono:
fixes #659

PR #664, opened 2023-04-27
extend drop psi functionality to categorical variables

### LINKED ISSUE #655: PSI feature selection for categorical features

solegalli:
closes #655
closes #658
closes #657

### DIFF
--- a/feature_engine/selection/drop_psi_features.py
+++ b/feature_engine/selection/drop_psi_features.py
@@ -35,7 +35,6 @@ from feature_engine.variable_handling._init_parameter_checks import (
 )
 from feature_engine.variable_handling.variable_type_selection import (
     find_categorical_and_numerical_variables,
-    find_or_check_numerical_variables,
 )
 
 Variables = Union[None, int, str, List[Union[str, int]]]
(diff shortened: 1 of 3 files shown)


solegalli:
Hey @dlaprins and @ClaudioSalvatoreArcidiacono

I've got the impression that rebasing didn't work in #657 because I can see the files that we merged in #660 in there.

I tried to rebase again, but got a lot of errors, so decided to start afresh.

The changes here in summary:

- I mostly tidied the docstrings
- fixed doc error
- did minor cosmetics in the class code
- added some tests for p-value

Would you be able to have a quick look and if OK we merge?

Thank you!

solegalli:
Thank you @dlaprins 

I am considering now on adding an option to the variables parameter in the init, not to break backwards compatibility.

In the live version, if variables=None, the transformer will operate only on numerical variables. If we deploy this version as is, when variables=None the transformer will operate on all variables, and thus break backward compatibility.

So I am thinking of leaving variable=None with the current behaviour and adding the option variables="all" to operate on all variables.

dlaprins:
A bit unwieldy for future use, but I suppose backwards compatibility takes precedence. I'm in favor.

PR #667, opened 2023-04-28
OHE: User can select which categories to encoded for selected variables

### LINKED ISSUE #303: OHE: allow encoding of specific, user desired categories
As per [this thread](https://stackoverflow.com/questions/65153240/categorical-data-encoder-for-unseen-data-in-python), the user may want to encode certain categories, that may not be the most frequent.

Morgan-Sell:
Closes #303.

**Objective:**
Create functionality so a user can encode certain categories that may not be the most frequent. The functionality is explained in this [thread](https://stackoverflow.com/questions/65153240/categorical-data-encoder-for-unseen-data-in-python).

Will create a new init param called `custom_categories` that accepts a dictionary with selected features as keys and lists of the desired categories as values.

Both `top_categories` and `user_categories` cannot be used at the same time.

### DIFF
--- a/feature_engine/encoding/one_hot.py
+++ b/feature_engine/encoding/one_hot.py
@@ -199,7 +199,7 @@ class OneHotEncoder(CategoricalInitMixin, CategoricalMethodsMixin):
                 f"Got {custom_categories} for custom_categories."
             )
 
-        if custom_categories and not isinstance(custom_categories, dict):
+        if custom_categories is not None and not isinstance(custom_categories, dict):
             raise ValueError(
                 "custom_categories must be a dictionary. "
                 f"Got {custom_categories} instead."


Morgan-Sell:
hi @solegalli,

I think we're getting close on this one.

Are there any other unit tests that we should implement? According to the coverage report, OHE has 100% coverage.

I'll dig into the type error.

Morgan-Sell:
Disregard my comment about code coverage. It seems that the transformer is very poorly "covered". I'll work on this.

feature_engine/encoding/one_hot.py
```diff
@@ -82,6 +86,17 @@ class OneHotEncoder(CategoricalInitMixin, CategoricalMethodsMixin):
...
+        If `top_categories` is being used, `custom_categories` must equal None.
+
+    custom_categories: dict, default=None
+        Accepts a dictionary in which the keys are the variables that the use would like
+        to transform. The keys must match the values of `variables` param.
+
+        The dicitonary values are lists of the categories for each selected variable
+        that are to be one-hot encoded.
+
+        If `custom_categories` is being used, `top_categories` must equal None.
```
solegalli, line 98: A few things to think:

- I think custom_categories should override top_categories. So no matter what the user enters, custom_categories takes precedence. It is easier to use like this imo. So here we should say "when custom_categories is entered, top_categories will be ignored.

The other thing to consider is, how do we want custom_categories to interplay with variables. Are we going to allow the user to specify categories for 3 variables for example, while encoding 6? What did you have in mind for this? 

For examples vars a,b,c,d,e are categorical, if variables=None, all will be encoded. If user passes a dictionary with keys, b and c, will variables a,d and e also be encoded? or are we making custom_categories take precedence over variables?

feature_engine/encoding/one_hot.py
```diff
@@ -160,9 +175,12 @@ class OneHotEncoder(CategoricalInitMixin, CategoricalMethodsMixin):
     def __init__(
         self,
         top_categories: Optional[int] = None,
+        custom_categories: Optional[Dict] = None,
         drop_last: bool = False,
         drop_last_binary: bool = False,
-        variables: Union[None, int, str, List[Union[str, int]]] = None,
+        variables: Union[
+            None, int, str, List[Union[str, int]], Iterable[Union[str, int]]
```
solegalli, line 182: why did you add iterable?

feature_engine/encoding/one_hot.py
```diff
@@ -302,3 +361,17 @@ def _add_new_feature_names(self, feature_names) -> List:
...
+
+    def _check_custom_categories_in_dataset(self, X: pd.DataFrame) -> None:
+        """
+        Raise an error if user entered categories in custom_categories that do
+        not exist within dataset.
+
+        """
+        for var, categories in self.custom_categories.items():
+            unique_values = set(X[var].unique())
+            if not set(categories).issubset(unique_values):
```
solegalli, line 373: what if, the user wants a dummy for a category that is not present in the train set? should we contemplate this or is too unlikely?

solegalli (commented):
Hey @Morgan-Sell 

I had a quick look at the logic of the new functionality. There are some things that we need to give more thought to. Would be great it you could have a look and tell me what you think?

Thank you!

feature_engine/encoding/one_hot.py
Morgan-Sell: Oh, Dr. Che, you never let me get away easy. No wonder I enjoy working w/ you ;)

I'm going to suggest that `variables` is superior. I envision scenarios in which users would like to encode all variables but would like to select which categories are encoded in certain variables.

Consequently, this transformer will require a check when `variables` does not equal `None` and the user utilizes `custom_categories`. The transformer should confirm that the variables included in `custom_categories` are a subset of the variables inputted into the `variables` param.

What do you think? ;)

feature_engine/encoding/one_hot.py
```diff
@@ -174,6 +192,40 @@ def __init__(
...
+            non_lists_custom_categories = [
+                val for val in custom_categories.values() if not isinstance(val, list)
+            ]
+            if len(non_lists_custom_categories) > 0:
+                raise ValueError(
+                    "custom_categories must be a dictionary that has lists as "
+                    f"its values. Got {custom_categories} instead."
+                )
+
+            # check that custom_categories variable match variables
```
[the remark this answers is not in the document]
Morgan-Sell: I'm reading this comment after responding to your prior question.

I agree that it will be a headache for us and useful for the user. I'll think about this more too.

feature_engine/encoding/one_hot.py
Morgan-Sell: When someone is OHE-ing a category that is not present in the training set, are they putting it into a "catch-all" bucket, e.g. `unlearned_categories`? Is this common?

I think it's unlikely and defeats the purpose of the `custom_categories` param. I see `custom_categories` being used to protect against unlearned categories.

feature_engine/encoding/one_hot.py
Morgan-Sell: Before refactoring the code, MyPy raised the following errors:

```
feature_engine/encoding/one_hot.py:221: error: Argument 1 to "sorted" has incompatible type "Union[int, str, List[Union[str, int]], None]"; expected "Iterable[Union[str, int]]"
feature_engine/encoding/one_hot.py:223: error: Argument 1 to "sorted" has incompatible type "Union[int, str, List[Union[str, int]], None]"; expected "Iterable[Union[str, int]]"
```

Morgan-Sell:
Hi @solegalli,

When you're back from vacay, lmk what you think.

Also, I don't know why I said that OHE has a low code coverage score. I just ran the coverage report - the transformer received 100% coverage!

feature_engine/encoding/one_hot.py
solegalli: Sounds good.

feature_engine/encoding/one_hot.py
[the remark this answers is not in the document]
solegalli: Alright then, let me know if you find an elegant solution :)

feature_engine/encoding/one_hot.py
solegalli: Here is an example: 

I work in a car insurance company. My historical data has 3 types of fuel: gas, petrol, diesel. But, I know that electric vehicles are coming. So I want an additional category: electric. 

In the above scenario, it might help if we allow the additional bespoke category.

On the other hand, a model can't be train in a category that does not exist, so the results would be unreliable. Maybe we don't allow them for now?

solegalli:
Hey @Morgan-Sell 

I disappeared for a while, so I am a bit lost. I guess you were waiting for my input here, no? You let me know when I need to take another look? I am away from next Thursday, for a month, ehem, (red face)

PR #676, opened 2023-06-06
return name of variables with 0s in denominator in woe encoder

### LINKED ISSUE #56: add test to determine if WoERatioCategoricalEncoder returns an error when the probability in the denominator is 0
At the moment the test is commented out:
https://github.com/solegalli/feature_engine/blob/master/tests/test_encoding/test_woe_encoder.py#L112

the aim is to test this bit of code [here](https://github.com/solegalli/feature_engine/blob/master/feature_engine/encoding/woe.py#L205)

I think the test should work, and I commented out because I changed the error for a warning. But now, we decided to go back to returning an error.

So in short, the aim is to corroborate that the commented out test checks the intended bit of code and if yes, uncomment it, otherwise, replace it by the suitable test.

### LINKED ISSUE #430: `WoEEncoder` should return a list with all the variables that have 0 in the denominator of the WoE
At the moment, the transformer fails when it encounters one variable with 0 in the denominator of the WoE formula. We would like it to assess and raise an error with all the variables that show this behaviour.

### LINKED ISSUE #672: woe encoder: inform users which variables have zero denominator

### LINKED ISSUE #673: woe encoder: add parameter to avoid division by zero

solegalli:
closes #672 
closes #430 
closes #56 

and also
closes #673

### DIFF
--- a/feature_engine/encoding/woe.py
+++ b/feature_engine/encoding/woe.py
@@ -53,7 +53,13 @@ class WoE:
             y = pd.Series(np.where(y == y.min(), 0, 1))
         return X, y
 
-    def _calculate_woe(self, X: pd.DataFrame, y: pd.Series, variable: Union[str, int], fill_value: Union[float, None] = None):
+    def _calculate_woe(
+        self,
+        X: pd.DataFrame,
+        y: pd.Series,
+        variable: Union[str, int],
+        fill_value: Union[float, None] = None,
+    ):
         total_pos = y.sum()
         inverse_y = y.ne(1).copy()
         total_neg = inverse_y.sum()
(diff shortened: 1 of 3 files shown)


solegalli:
Hey @glevv 

I am expanding the functionality of the WoE encoder to do 2 things:

1. return in the error message the name of all variables that had categories with zero in the numerator or denominator of the WoE calculation (that happens when the sum of target=0 or target=1 is zero for that category)
2. when the denominator or numerator is zero, replace by an arbitrary value and proceed with the calculation

Regarding 1: the error was there already. The improvement consists in letting users know which variables are the problematic

Regarding 2: I think this defeats the point of the calculation, but I guess we leave this to the user. The implementation is different from category encoders because 1) category encoders adds the regularization to all categories, regardless of whether they have zero or not. Here we modify only those with 0. And 2) we replace the 0 with the fill value. Category encoders adds the regularization to both denominator and numerator)

I would appreciate if you had 2 minutes to let me know what you think about the implementation, and if you can go over the code, even better.

Thanks a lot!

PR #679, opened 2023-06-11
Add the inverse_transform method in the yeoJohnson transformer

GiorgioSgl:
Add the inverse_transform function in the yeoJohnson transformer.

In sklearn PowerTransformer documentation is explained in the following way:

    The inverse of the Box-Cox transformation is given by:

```
    if lambda_ == 0:
    X = exp(X_trans)
    else:
    X = (X_trans * lambda_ + 1) ** (1 / lambda_)

    The inverse of the Yeo-Johnson transformation is given by:

    if X >= 0 and lambda_ == 0:
    X = exp(X_trans) - 1
    elif X >= 0 and lambda_ != 0:
    X = (X_trans * lambda_ + 1) ** (1 / lambda_) - 1
    elif X < 0 and lambda_ != 2:
    X = 1 - (-(2 - lambda_) * X_trans + 1) ** (1 / (2 - lambda_))
    elif X < 0 and lambda_ == 2:
    X = 1 - exp(-X_trans)
```

### DIFF
--- a/feature_engine/transformation/yeojohnson.py
+++ b/feature_engine/transformation/yeojohnson.py
@@ -7,17 +7,21 @@ import numpy as np
 import pandas as pd
 import scipy.stats as stats
 
-from feature_engine._base_transformers.base_numerical import \
+from feature_engine._base_transformers.base_numerical import (
     BaseNumericalTransformer
+)
 from feature_engine._docstrings.fit_attributes import (
     _feature_names_in_docstring, _n_features_in_docstring,
-    _variables_attribute_docstring)
-from feature_engine._docstrings.init_parameters.all_trasnformers import \
+    _variables_attribute_docstring
+)
+from feature_engine._docstrings.init_parameters.all_trasnformers import (
     _variables_numerical_docstring
+)
 from feature_engine._docstrings.methods import _fit_transform_docstring
 from feature_engine._docstrings.substitute import Substitution
-from feature_engine.variable_handling._init_parameter_checks import \
+from feature_engine.variable_handling._init_parameter_checks import (
     _check_init_parameter_variables
+)
 
 
 @Substitution(
--- a/feature_engine/transformation/yeojohnson.py
+++ b/feature_engine/transformation/yeojohnson.py
@@ -3,6 +3,7 @@
 
 from typing import List, Optional, Union
 
+import numpy as np
 import pandas as pd
 import scipy.stats as stats
 
@@ -103,7 +104,6 @@ class YeoJohnsonTransformer(BaseNumericalTransformer):
     def __init__(
         self, variables: Union[None, int, str, List[Union[str, int]]] = None
     ) -> None:
-
         self.variables = _check_init_parameter_variables(variables)
 
     def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
@@ -152,3 +152,45 @@ class YeoJohnsonTransformer(BaseNumericalTransformer):
             X[feature] = stats.yeojohnson(X[feature], lmbda=self.lambda_dict_[feature])
 
         return X
+
+    def inverse_transform(self, X: pd.DataFrame) -> pd.DataFrame:
+        """
+        Apply the inverse-transform function Yeo-Johnson.
+
+        Parameters
+        ----------
+        X: Pandas DataFrame of shape = [n_samples, n_features]
+            The data to be inversed.
+
+        Returns
+        -------
+        X: pandas dataframe
+            The dataframe with the inversed variables, so original values.
+        """
+        # check input dataframe and if class was fitted
+        X = self._check_transform_input_and_state(X)
+
+        for feature in self.variables_:
+            X[feature] = self._inverse_transform_series(
+                X[feature], lmbda=self.lambda_dict_[feature]
+            )
+
+        return X
+
+    def _inverse_transform_series(self, X: pd.Series, lmbda: float) -> pd.Series:
+        x_inv = pd.Series(np.zeros_like(X))
+        pos = X >= 0
+
+        # when x >= 0
+        if lmbda == 0:
+            x_inv[pos] = np.exp(X[pos]) - 1
+        else:  # lmbda != 0
+            x_inv[pos] = np.power(X[pos] * lmbda + 1, 1 / lmbda) - 1
+
+        # when x < 0
+        if lmbda != 2:
+            x_inv[~pos] = 1 - np.power(-(2 - lmbda) * X[~pos] + 1, 1 / (2 - lmbda))
+        else:  # lmbda == 2
+            x_inv[~pos] = 1 - np.exp(-X[~pos])
+
+        return x_inv
--- a/feature_engine/transformation/yeojohnson.py
+++ b/feature_engine/transformation/yeojohnson.py
@@ -7,20 +7,17 @@ import numpy as np
 import pandas as pd
 import scipy.stats as stats
 
-from feature_engine._base_transformers.base_numerical import BaseNumericalTransformer
+from feature_engine._base_transformers.base_numerical import \
+    BaseNumericalTransformer
 from feature_engine._docstrings.fit_attributes import (
-    _feature_names_in_docstring,
-    _n_features_in_docstring,
-    _variables_attribute_docstring,
-)
-from feature_engine._docstrings.init_parameters.all_trasnformers import (
-    _variables_numerical_docstring,
-)
+    _feature_names_in_docstring, _n_features_in_docstring,
+    _variables_attribute_docstring)
+from feature_engine._docstrings.init_parameters.all_trasnformers import \
+    _variables_numerical_docstring
 from feature_engine._docstrings.methods import _fit_transform_docstring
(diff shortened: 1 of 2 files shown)


GiorgioSgl:
Idk why it's failing some tests in the rare label encoder. In the previous commit 3bfc6d4047b44cb84bce0ef42e5e5dad9f450695 it was passed, and now I only modify style (directly with black tool)

feature_engine/transformation/yeojohnson.py
```diff
@@ -152,3 +152,45 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
             X[feature] = stats.yeojohnson(X[feature], lmbda=self.lambda_dict_[feature])
 
         return X
+
+    def inverse_transform(self, X: pd.DataFrame) -> pd.DataFrame:
+        """
+        Apply the inverse-transform function Yeo-Johnson.
```
solegalli, line 158: All inverse_transform methods take the same info: "Convert the data back to the original representation." Could you please update this sentence?

feature_engine/transformation/yeojohnson.py
```diff
@@ -152,3 +152,45 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+
+        Parameters
+        ----------
+        X: Pandas DataFrame of shape = [n_samples, n_features]
+            The data to be inversed.
+
+        Returns
+        -------
+        X: pandas dataframe
```
solegalli, line 167: Also the return, should be the same in all transformers. See this:

https://github.com/feature-engine/feature_engine/blob/3343305a01d1acfeff846b65d33a5686c6e8c84f/feature_engine/transformation/log.py#L173-L185

Could you update the docstrings here as well please?

tests/test_transformation/test_yeojohnson_transformer.py
```diff
@@ -1,6 +1,7 @@
 import pandas as pd
 import pytest
 from sklearn.exceptions import NotFittedError
+import numpy as np
```
solegalli, line 4: we follow pep8 for imports. this should be at the top. 

You could use isort: `isort path/to/this/file on the command line`.

More info here: https://feature-engine.trainindata.com/en/latest/contribute/contribute_code.html#test-code-style

tests/test_transformation/test_yeojohnson_transformer.py
```diff
@@ -84,3 +85,30 @@ def test_non_fitted_error(df_vartypes):
     with pytest.raises(NotFittedError):
         transformer = YeoJohnsonTransformer()
         transformer.transform(df_vartypes)
+
+
+def test_inverse_transform_wit_columns_of_other_types(df_vartypes):
```
solegalli, line 90: "wit" has a typo. What do you men by "column of other types"? I don't understand :_(

You mean different data types? as in int and floats?

solegalli (commented):
Hey @GiorgioSgl 

Thank you so much for this code contribution. The code looks really good and it is a valuable addition.

And, sorry for the delayed review, I was on holidays :)

There are 4 possible scenarios for the inverse_transform:

- variable is positive and lambda is 0
- variable is positive and lambda is not 0
- variable is negative and lambda is 2
- variable is negative and lambda is not 2.

We need to ensure that the inverse_transform has access to all of these lines in the tests.

The 2 tests added check the overall functionality of inverse transform, and that is very good.

I think we could add fabricated tests where we have positive and negative variables, we then modify the lambda in the lambda_dict to make it 0 or 2, and then test that the result is what we expect (which we could obtain with a manual calculation.) Would you agree and be able to add tests for this as well?

The error as you say is unrelated. I'll try to have it fixed in a separate PR.

Thanks a lot!

solegalli:
Hey @glevv @ClaudioSalvatoreArcidiacono @dlaprins, would any of you have time to have a look at these 2 tests and see why they fail?

```
=========================== short test summary info ============================
FAILED tests/test_encoding/test_rare_label_encoder.py::test_correctly_ignores_nan_in_fit - AssertionError: assert {'var_A': ['B...C', 'B', 'A']} == {'var_A': ['B...D'...
FAILED tests/test_encoding/test_rare_label_encoder.py::test_correctly_ignores_nan_in_fit_when_var_is_numerical - AssertionError: assert {'var_A': ['B....0, 2.0, 1.0]} == {'var_A': ['B... [...
========== 2 failed, 1434 passed, 2924 warnings in 176.15s (0:02:56) ===========
py39: exit 1 (177.96 seconds) /home/circleci/project> pytest tests pid=498
  py39: FAIL code 1 (202.37=setup[24.41]+cmd[177.96] seconds)
  evaluation failed :( (202.44 seconds)

```

I think it my be related to some randomization, because it fails in only the latest Python versions. 

If you have time, I'd greatly appreciate if you could fixed them in a different PR :)

GiorgioSgl:
This weekend I will check all the comments

tests/test_transformation/test_yeojohnson_transformer.py
GiorgioSgl: by running 

`isort feature_engine/transformation/yeojohnson.py`

I got this:

```python
from typing import List, Optional, Union

import numpy as np
import pandas as pd
import scipy.stats as stats

from feature_engine._base_transformers.base_numerical import \
    BaseNumericalTransformer
from feature_engine._docstrings.fit_attributes import (
    _feature_names_in_docstring, _n_features_in_docstring,
    _variables_attribute_docstring)
from feature_engine._docstrings.init_parameters.all_trasnformers import \
    _variables_numerical_docstring
from feature_engine._docstrings.methods import _fit_transform_docstring
from feature_engine._docstrings.substitute import Substitution
from feature_engine.variable_handling._init_parameter_checks import \
    _check_init_parameter_variables
```

Is it OK like that?

feature_engine/transformation/yeojohnson.py
GiorgioSgl: Change the entire dosctring to this

```python
        """
        Convert the data back to the original representation.

        Parameters
        ----------
        X: Pandas DataFrame of shape = [n_samples, n_features]
            The data to be transformed.

        Returns
        -------
        X_tr: pandas dataframe
            The dataframe with the transformed variables.
        """
```

tests/test_transformation/test_yeojohnson_transformer.py
GiorgioSgl: Me neither actually, bad name and mistake by my side. I think I mean to test it with data with a more types like str, int, float and date in this case.

Change its name to 
`test_inverse_transform_automatically_select_only_transformed_columns`

GiorgioSgl:
Just fixed the little problems, tell if have to fix something more.

This weekend I will check the problems with the tests of the different lambda.

tests/test_transformation/test_yeojohnson_transformer.py
solegalli: I don't think so, we don't use \

GiorgioSgl: Sorry I was misconfusing the files, and I apply the isort on the normal file and not on the tests
So just fix the import in the tests, now it's all okay.

Also I remove the \ from the yehojonson.py file.

GiorgioSgl:
Just fixed all the review you give me, I think. Let me know if there is something more!

tests/test_transformation/test_yeojohnson_transformer.py
```diff
@@ -84,3 +85,68 @@ def test_non_fitted_error(df_vartypes):
...
+
+    transformer = YeoJohnsonTransformer(variables=None)
+    transformer = transformer.fit(X)
+
+    transformer.lambda_dict_ = {"var1": 0, "var2": 0}
+
+    X_trans = transformer.transform(X)
+    X_inverse = transformer.inverse_transform(X_trans)
+
+    pd.testing.assert_frame_equal(X, X_inverse, check_dtype=False)
```
solegalli, line 133: Hey @GiorgioSgl 

Thank you so much for the revised version of the code. We are almost there!!

I am not sure I understand this test. Maybe you can help me.

in line 126, the transformer is fit to X, hence it learns some lambdas. Which are the lambda values? Not 0, are they?

in line 128, we re-define the lambda dictionary, because we want to test the inverse_transform functionality for lambda 0, in short, that when lambda is zero, this tests that inverse transform applies the correct calculation.

So my idea here was, to apply inverse_transform to X, it will return some values. Then, we need to perform the calculations manually, to know what the transformation of X would return when it enters the inverse transform with lambda zero, and then assert that X_inverse is the same as a pandas dataframe, that we would hard-code with the expected values.

But here, something else is happening. What was your idea?

The same for the coming test.

Thank you for your help!

solegalli:
Hey @GiorgioSgl 

I made a PR to your repo to sort the failing tests: https://github.com/GiorgioSgl/feature_engine/pull/1

Would you be able to merge so it updates here?

tests/test_transformation/test_yeojohnson_transformer.py
solegalli: Not sure what I was thinking here, but the tests are great. Thank you!

PR #686, opened 2023-07-22
Raise informative error on duplicated column names

david-cortes:
This PR adds an informative error message in cases in which the user supplies inputs having duplicated column names, which otherwise manifest in hard-to-track errors (e.g. https://github.com/feature-engine/feature_engine/pull/681).


feature_engine/variable_handling/_variable_type_checks.py
```diff
@@ -47,3 +48,11 @@ def _is_categorical_and_is_datetime(column: pd.Series) -> bool:
...
 
     return is_dt
+
+
+def _raise_on_duplicated_column_names(X: pd.DataFrame, variables: Any) -> None:
+    df_check = X if variables is None else X[variables]
+    if isinstance(df_check, pd.Series):
+        return
+    if df_check.columns.duplicated().any():
+        raise ValueError("Input data contains duplicated variable names.")
```
solegalli, line 58: Quick questions:

Does df_check.columns.duplicated().any() raise an error on a series?

Is this implementation better practice? or would it be better to say:

```
    if not isinstance(df_check, pd.Series):
        if df_check.columns.duplicated().any():
            raise ValueError("Input data contains duplicated variable names.")
```

solegalli (commented):
Hey @david-cortes 

Thank you so much for taking care of this issue. 

I wonder if instead of modifying all the check variable functions, it would not be better to modify the check_X that checks the input dataframe. For example here:

https://github.com/feature-engine/feature_engine/blob/3343305a01d1acfeff846b65d33a5686c6e8c84f/feature_engine/dataframe_checks.py#L49

In this case, if the dataframe contains duplicated variables, it will raise an error and stop all computations right at the top of the fit method.

Do you know if / how pandas warns users of duplicated column names somehow?

feature_engine/variable_handling/_variable_type_checks.py
david-cortes: a. It does raise an error, since `Series` objects do not have `columns`. Nevertheless, if slicing a `DataFrame` with a column name returns a `Series`, it means there was only one column with that name.
b. No idea. The linter didn't complain about the one in my initial commit.

david-cortes:
Moved the check to `check_X` instead as suggested above.

Regarding errors from `pandas`, it does throw errors sometimes when there are duplicates, but only under some particular situations:
https://pandas.pydata.org/pandas-docs/stable/user_guide/duplicates.html

Also checked the mechanism towards the attribute `is_unique` as it seems that's what they recommend in their guide.

solegalli (commented):
Hey @david-cortes 

Thank you so much for the link to pandas documentation on duplicate labels. That was very useful. And thanks for the code changes in check_X. That looks very good.

After reading the documentation, I think this addition you are proposing is important and very useful.

I think, it'd be important to add a test in the test_check_x file: https://github.com/feature-engine/feature_engine/blob/main/tests/test_dataframe_checks.py

Our test_files mirror our .py files. So we would not have a file in variable_handling if we do not add a function in variable handling. So I think we need to remove the file you added called `test_error_on_duplicated`, and pass the test to other files.

This test is a generic test that all transformers should pass. We have generic tests here: https://github.com/feature-engine/feature_engine/tree/main/tests/estimator_checks

But I need to refurbish that file a lot. So I wouldn't want to waste your time there. If you could just add a test to the test_check_x file I mention previously, and then create an issue saying that we need to add a generic test for all transformers to test that they raise error on duplicates, that would be great and we could finish this PR.

Thank you!

tests/test_variable_handling/test_error_on_duplicated_cols.py
```diff
@@ -0,0 +1,63 @@
...
+        return True
+    
+    def fit(self, X, y=None, *args, **kwargs):
+        return self
+
+
+def test_error_on_autodetected_numerical():
+    with pytest.raises(ValueError):
+        est = BaseNumericalTransformer()
+        est.variables = None
```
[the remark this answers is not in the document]
solegalli: We should remove this file anyways, so feel free to ignore my comments here.

david-cortes:
Moved the tests to `test_dataframe_checks.py`.

tests/test_dataframe_checks.py
```diff
@@ -170,3 +170,17 @@ def test_contains_inf(df_na):
...
+    df = pd.DataFrame(
+        {
+            "col1": [1, 2, 3],
+            "col2": ["a", "b", "c"],
+            "col3": pd.date_range("2023-01-01", periods=3),
+        }
+    )
+    df.columns = ["same", "unique", "same"]
+
+    with pytest.raises(ValueError):
```
solegalli, line 185: nitpick: could we add that it tests the error message as well? And example on how to do so here:

https://github.com/feature-engine/feature_engine/blob/3343305a01d1acfeff846b65d33a5686c6e8c84f/tests/test_encoding/test_woe/test_woe_encoder.py#L189-L197

david-cortes:
Added a check on the error message.

solegalli:
Hey @david-cortes 

I made a PR to your repo: https://github.com/david-cortes/feature_engine/pull/4

Where I rebase main and add this contribution to the changelog.

Would you have time to merge over there, so it updates here and I can merge and close?

Thanks a lot!

david-cortes:
> Hey @david-cortes
> 
> I made a PR to your repo: [david-cortes#4](https://github.com/david-cortes/feature_engine/pull/4)
> 
> Where I rebase main and add this contribution to the changelog.
> 
> Would you have time to merge over there, so it updates here and I can merge and close?
> 
> Thanks a lot!

Thanks, although I think you should also be able to push changes to the branch directly.

PR #694, opened 2023-09-12
Features/distances

VascoSch92:
Just a first sketch.

Let me know what do you think :-)

### DIFF
--- a/feature_engine/creation/distance_features.py
+++ b/feature_engine/creation/distance_features.py
@@ -1,18 +1,39 @@
-from typing import List, Optional, Tuple, Union
+from typing import List, Optional, Union
 
 import numpy as np
 import pandas as pd
 
-from feature_engine._base_transformers.base_numerical import BaseNumericalTransformer
-from feature_engine._base_transformers.mixins import (
-    FitFromDictMixin,
-    GetFeatureNamesOutMixin,
+from feature_engine._docstrings.fit_attributes import (
+    _feature_names_in_docstring,
+    _n_features_in_docstring,
+    _variables_attribute_docstring,
 )
+from feature_engine._docstrings.init_parameters.all_trasnformers import (
+    _drop_original_docstring,
+    _missing_values_docstring,
+    _variables_numerical_docstring,
+)
+from feature_engine._docstrings.methods import (
+    _fit_not_learn_docstring,
+    _fit_transform_docstring,
+    _transform_creation_docstring,
+)
+from feature_engine._docstrings.substitute import Substitution
+from feature_engine.creation.base_creation import BaseCreation
 
 
-class DistanceFeatures(
-    BaseNumericalTransformer, FitFromDictMixin, GetFeatureNamesOutMixin
-):
+@Substitution(
+    variables=_variables_numerical_docstring,
+    missing_values=_missing_values_docstring,
+    drop_original=_drop_original_docstring,
+    variables_=_variables_attribute_docstring,
+    feature_names_in_=_feature_names_in_docstring,
+    n_features_in_=_n_features_in_docstring,
+    fit=_fit_not_learn_docstring,
+    transform=_transform_creation_docstring,
+    fit_transform=_fit_transform_docstring,
+)
+class DistanceFeatures(BaseCreation):
     """
     DistanceFeatures() computes the distance between pairs of columns containing
     coordinates. The distance between two pairs of coordinates is computed using the
@@ -94,46 +115,77 @@ class DistanceFeatures(
         drop_original: bool = False,
     ) -> None:
 
+        if (
+            not isinstance(coordinate_columns, list)
+            or not all(
+                isinstance(sublist, list)
+                or not all(isinstance(item, (int, str)) for item in sublist)
+                for sublist in coordinate_columns
+            )
+            or len(set(coordinate_columns)) != len(coordinate_columns)
+        ):
+            raise ValueError(
+                "coordinate_columns must be a list of lists of strings or integers. "
+                f"Got {coordinate_columns} instead."
+            )
+        if not all(
+            len(coordinate_column) == 4 for coordinate_column in coordinate_columns
+        ):
+            invalid_coordinate_columns = [
+                coordinate_column
+                for coordinate_column in coordinate_columns
+                if len(coordinate_column) != 4
+            ]
+            raise ValueError(
+                f"coordinate_columns must be a list of lists of 4 elements."
+                f"Got {invalid_coordinate_columns}."
+            )
+
+        # TODO: check for the output_column_names
+        # TODO: what about missing values
+
+        super().__init__(drop_original=drop_original)
         # the coordinate_columns variable is rewritten in this way to speed up
         # computation later, i.e., to use vectorization
-        (self.a_latitudes, self.a_longitudes, self.b_latitudes, self.b_longitudes) = (
-            self._check_coordinate_columns(columns=coordinate_columns)
-        )
         self.output_column_name = self._check_output_columns_names(
             column_name=output_column_names,
             coordinate_columns=coordinate_columns,
         )
 
-        self.drop_original = self._check_drop_original(parameter=drop_original)
+        self.variables = coordinate_columns
 
-        self.variables = None
+    def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
+        X = super().fit(X)
 
-    def _check_drop_original(self, parameter: bool) -> bool:
-        if isinstance(parameter, bool) is False:
-            raise ValueError(
-                "Expected boolean value for parameter `drop_original`, "
-                f"but got {parameter} with type {type(parameter)}"
-            )
-        return parameter
+        self._check_correctness_of_coordinates(X)
 
-    def _check_coordinate_columns(
-        self, columns: List[List[Union[str, int]]]
-    ) -> Tuple[List[Union[str, int]], ...]:
-        if not columns:
-            raise ValueError("Empty list for `coordinate_columns` not allowed!")
-        for idx, coordinate_column in enumerate(columns):
-            if len(coordinate_column) != 4:
-                raise ValueError(
-                    f"Needed 4 values to compute a distance, "
-                    f"but got {len(coordinate_column)} columns \n"
-                    f"at the index {idx} of the list coordinate columns."
-                )
-        return (
-            [coordinate[0] for coordinate in columns],
-            [coordinate[1] for coordinate in columns],
-            [coordinate[2] for coordinate in columns],
-            [coordinate[3] for coordinate in columns],
+        (self.a_latitudes, self.a_longitudes, self.b_latitudes, self.b_longitudes) = [
+            coordinate[0] for coordinate in self.variables_
+        ]
+
+    def _check_correctness_of_coordinates(self, X: pd.DataFrame):
+        # recall that the latitude is a number between -90 and +90,
+        # while longitudes is between -180 and +180.
+        irregular_latitudes = (
+            (X[[*self.a_latitudes, *self.b_latitudes]].abs() > 90).sum().sum()
         )
+        irregular_longitudes = (
+            (X[[*self.a_longitudes, *self.b_longitudes]].abs() > 180).sum().sum()
+        )
+
+        error_message = ""
+        if irregular_latitudes > 0:
+            error_message += (
+                "The dataframe contains irregular latitudes. "
+                "Recall that a latitude is a number between -90 and 90. \n"
+            )
+        if irregular_longitudes > 0:
+            error_message += (
+                "The dataframe contains irregular longitudes"
+                "Recall that a longitude is a number between -180 and 180."
+            )
+        if error_message:
+            raise ValueError(error_message)
 
     def _check_output_columns_names(
         self,
@@ -150,11 +202,6 @@ class DistanceFeatures(
             )
         return column_name
 
-    def fit(self, X: pd.DataFrame):
-        # there is no fit for this transformer
-        super().fit(X)
-        return self
-
     def transform(self, X: pd.DataFrame):
         """
         Compute the distance on heart using the Haversine formula.
@@ -170,7 +217,6 @@ class DistanceFeatures(
             The original dataframe plus the distances between the given coordinates.
         """
         X = self._check_transform_input_and_state(X)
-        X = self._check_lat_lon_columns_are_in_df(X)
         X = self._check_correctness_of_coordinates(X)
 
         self._compute_distance(X)
@@ -187,37 +233,6 @@ class DistanceFeatures(
             )
 
         return X
-
-    def _check_lat_lon_columns_are_in_df(self, X) -> pd.DataFrame:
-        coordinate_columns = [
-            *self.a_latitudes,
-            *self.a_longitudes,
-            *self.b_latitudes,
-            *self.b_longitudes,
-        ]
-        if set(coordinate_columns).issubset(set(X.columns)) is False:
-            raise ValueError(
-                f"The columns {set(coordinate_columns).issubset(set(X.columns))} "
-                f"were not found in the dataframe."
-            )
-        return X
-
-    def _check_correctness_of_coordinates(self, X: pd.DataFrame) -> pd.DataFrame:
-        # recall that the latitude is a number between -90 and +90,
-        # while longitudes is between -180 and +180.
-        irregular_latitudes = (
-            (X[[*self.a_latitudes, *self.b_latitudes]].abs() > 90).sum().sum()
-        )
-        irregular_longitudes = (
-            (X[[*self.a_longitudes, *self.b_longitudes]].abs() > 180).sum().sum()
-        )
-
-        if irregular_latitudes > 0:
-            raise ValueError("The dataframe contains irregular latitudes")
-        elif irregular_longitudes > 0:
-            raise ValueError("The dataframe contains irregular longitudes")
-        else:
-            return X
 
     def _compute_distance(self, X: pd.DataFrame):
 
--- a/feature_engine/creation/__init__.py
+++ b/feature_engine/creation/__init__.py
@@ -2,14 +2,17 @@
 The module creation includes classes to create new variables by combination of existing
 variables in the dataframe.
 """
+
 from .cyclical_features import CyclicalFeatures
 from .decision_tree_features import DecisionTreeFeatures
 from .math_features import MathFeatures
 from .relative_features import RelativeFeatures
+from .distance_features import DistanceFeatures
 
 __all__ = [
     "DecisionTreeFeatures",
     "MathFeatures",
     "RelativeFeatures",
     "CyclicalFeatures",
+    "DistanceFeatures",
 ]


feature_engine/creation/distance_features.py
```diff
@@ -0,0 +1,118 @@
...
+
+        self.output_column_name = self._check_column_name(column_name=output_column_name)
+
+        _check_param_drop_original(drop_original=drop_original)
+        self.drop_original = drop_original
+
+        self.variables = None
+
+    @staticmethod
+    def _check_column_name(column_name: str) -> str:
```
solegalli, line 44: We allow integers as column names. I think I introduced this because it helps accommodate some of the compatibility tests with sklearn. In short, I don't think we need this check. We "outsource" this to pandas.

Could we remove?

feature_engine/creation/distance_features.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+class DistanceFeatures(
+    BaseNumericalTransformer, FitFromDictMixin, GetFeatureNamesOutMixin
+):
+    EARTH_RADIUS: float = 6371.  # radius of Earth in kms
+
+    def __init__(
+            self,
+            a_latitude: str,
```
solegalli, line 23: We allow integers as column names.

To think: is this transformer intended to compare the distance between 2 points?

Because what we normally do (see for example RelativeFeatures), is that we allow a list of variables in the first parameter, that would be we allow a list in a_latitude and a_longitude, and we allow for example a variable, or we could also think of allowing a list for b latitude and longitude, and then we calculate the distance between those in a and those in b.

This way, if the user wants to calculate more than 1 distance, they don't need to set multiple transformers.

What do you think?

For clarity, my df has for example, address victim 1, lat and lon, address victim 2, lat and lon, address perpetrator, lat and lon, and we want to calculate the distance between victim 1 and perpetrator and victim 2 and perpetrator. And if we had 2 perpetrators, then we want the distances of both victims to both perpetrators. Not sure why I am talking about victims and perpetrators, it sounds horrible. But it illustrates the issue :p

feature_engine/creation/distance_features.py
```diff
@@ -0,0 +1,118 @@
...
+            a_longitude: str,
+            b_latitude: str,
+            b_longitude: str,
+            output_column_name: Union[str, None] = None,
+            drop_original: bool = False,
+    ) -> None:
+
+        self.a_latitude = self._check_column_name(a_latitude)
```
solegalli, line 31: I'd probably just assign whatever the user passes. We need to check if it is a string or a list only, and we have that function ready to import.

feature_engine/creation/distance_features.py
```diff
@@ -0,0 +1,118 @@
...
+
+        # compute delta, i.e., difference, between radians
+        delta_phi = phi_2 - phi_1
+        delta_lambda = lambda_2 - lambda_1
+
+        # compute distance using Haversine formula
+        inner_part = np.sin(delta_phi / 2) ** 2 + np.cos(phi_1) * np.cos(phi_2) * np.sin(delta_lambda / 2) ** 2
+        X[self.output_column_name] = self.EARTH_RADIUS * 2 * np.arcsin(np.sqrt(inner_part))
+
+    def _check_lat_lon_columns_are_in_df(self, X) -> pd.DataFrame:
```
solegalli, line 100: we have the check_X function that we make part of the init, and it is also in check_transform_input_and state that take care of this. We don't need this extra methods.

feature_engine/creation/distance_features.py
```diff
@@ -0,0 +1,118 @@
...
+            raise ValueError(f'The columns {input_columns.difference(df_columns)} were not found in the dataframe.')
+
+        return X
+
+    def _check_correctness_of_coordinates(self, X: pd.DataFrame) -> pd.DataFrame:
+        irregular_latitudes = X[(X[self.a_latitude].abs() > 90) | (X[self.b_latitude].abs() > 90)]
+        irregular_longitudes = X[(X[self.a_longitude].abs() > 180) | (X[self.b_longitude].abs() > 180)]
+
+        if irregular_latitudes.empty is False:
+            raise ValueError(f'The dataframe contains irregular latitudes: {irregular_latitudes}')
```
solegalli, line 114: what does this error message return? the portion of the df with irregular latitudes?  I think it would be more useful to the user if the error says "variables so and so have latitude values greater than 90 for example.

Also, do we want to raise an error? or we want to return nan? or we want to make it optional? for example control the output through a parameter that I think we call error in other transformers, if ignore, we set to nan, if raise we raise these errors you have here?

feature_engine/creation/distance_features.py
```diff
@@ -0,0 +1,118 @@
...
+        self.a_longitude = self._check_column_name(a_longitude)
+        self.b_latitude = self._check_column_name(b_latitude)
+        self.b_longitude = self._check_column_name(b_longitude)
+
+        self.output_column_name = self._check_column_name(column_name=output_column_name)
+
+        _check_param_drop_original(drop_original=drop_original)
+        self.drop_original = drop_original
+
+        self.variables = None
```
solegalli, line 41: I don't think we need this parameter. I'd suggest using RelativeFeatures as template to model this class: https://github.com/VascoSch92/feature_engine/blob/e1e927625678ee73c5c3a9edcf79e955ff9c5e8e/feature_engine/creation/relative_features.py

solegalli (commented):
Hey @VascoSch92 

Thank you so much for the contribution! This is looking really good.

I think the logic for the distance calculation is good. We need to make this transformer look like the other transformers that we have in the library.

Why don't you have a look at the class RelativeFeatures in the creation module, and try to incorporate that logic in this class as well? Mostly related to which parameters we need in the init, and which checks we normally do. And there you will see as well how you can import many premade bits of text for the documentation.

Let me know how you get along! Thank you!

feature_engine/creation/distance_features.py
VascoSch92: I see your point. 

do you would like something as: `variables=[[lat_1, lon_1, lat_2, lon_2], [lat_1, lon_1, lat_3, lon_3]]` and the transformer compute the distance between lat_11, lon_1 and lat_2, lon_2 and the distance between lat_1, Lon_1 and lat_3, lon_3 ?

feature_engine/creation/distance_features.py
```diff
@@ -0,0 +1,118 @@
...
+        phi_1 = np.radians(X[self.a_latitude])
+        phi_2 = np.radians(X[self.b_latitude])
+        lambda_1 = np.radians(X[self.a_longitude])
+        lambda_2 = np.radians(X[self.b_longitude])
+
+        # compute delta, i.e., difference, between radians
+        delta_phi = phi_2 - phi_1
+        delta_lambda = lambda_2 - lambda_1
+
+        # compute distance using Haversine formula
```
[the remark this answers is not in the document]
VascoSch92: There is wikipedia for example: https://en.wikipedia.org/wiki/Haversine_formula.

I will add the link as a comment

feature_engine/creation/distance_features.py
VascoSch92: ok... but I don't understand. Should we allow a column name to be an integer?

Because you wrote:

> We allow integers as column names. I think I introduced this because it helps accommodate some of the compatibility tests with sklearn. In short, I don't think we need this check. We "outsource" this to pandas.

Therefore, I should checks that what we get is a list of str or integer, not?

feature_engine/creation/distance_features.py
VascoSch92: yes, I will change that

feature_engine/creation/distance_features.py
VascoSch92: Yes, I see your point.

For me the responsibility should be on the final user, i.e., if you call the transformer on a column with coordinates, then you are sure that the coordinates contained in the column are correct. If you forgot to check that, we will remember you raising an error. 

But I'm open to also do differently :-)

glevv:
Is it possible to calculate Haversine distance using sklearn? It is quite fast and well optimized, reimplementing it seems like a not so good idea

VascoSch92:
> Is it possible to calculate Haversine distance using sklearn? It is quite fast and well optimized, reimplementing it seems like a not so good idea

Yes of course I know that. The question is: Can you vectorise it? it is faster than vectorisation?

glevv:
> > Is it possible to calculate Haversine distance using sklearn? It is quite fast and well optimized, reimplementing it seems like a not so good idea
> 
> Yes of course I know that. The question is: Can you vectorise it? it is faster than vectorisation?

I'm not sure I understand the question.
[Scikit-learn implementation is vectorized by default](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise.haversine_distances.html)

kylegilde:
> > > Is it possible to calculate Haversine distance using sklearn? It is quite fast and well optimized, reimplementing it seems like a not so good idea
> > 
> > 
> > Yes of course I know that. The question is: Can you vectorise it? it is faster than vectorisation?
> 
> I'm not sure I understand the question. [Scikit-learn implementation is vectorized by default](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise.haversine_distances.html)

I think the issue with the sklearn implementation is that it does a cartesian product between X and Y and yields a matrix.

We only need a pairwise calculation between X and Y that yields a vector.

glevv:
`np.diag(haversine_distances(X, Y) * R)` would give you the vector you want

kylegilde:
> haversine_distances

I know that it is a simple way to code this, but from a time complexity perspective, it's not a great idea to use quadratic complexity when only linear complexity is needed.

glevv:
> > haversine_distances
> 
> I know that it is a simple way to code this, but from a time complexity perspective, it's not a great idea to use quadratic complexity when only linear complexity is needed.

Yea, you are right, this way it will be better

feature_engine/creation/distance_features.py
solegalli: If we have:

- variables=[lat_1, lon_1, lat_2, lon_2]
- reference=[lat_3, lon_3, lat_4, lon_4]

The result would be the following distances:

- lat_1;lon_1 and lat_3;lon3
- - lat_1;lon_1 and lat_4;lon4
- - lat_2;lon_2 and lat_3;lon3
- - lat_2;lon_2 and lat_3;lon3

feature_engine/creation/distance_features.py
solegalli: We have some example checks here:

https://github.com/feature-engine/feature_engine/blob/77ea405f18372bf3bcd8ef566bc7c3df5d66a056/feature_engine/creation/relative_features.py#L140-L157

feature_engine/creation/distance_features.py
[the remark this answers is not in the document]
solegalli: Awesome, thank you! We should include it in the docstrings of the class :)

feature_engine/creation/distance_features.py
solegalli: Let's go with your suggestion then.

solegalli:
> Is it possible to calculate Haversine distance using sklearn? It is quite fast and well optimized, reimplementing it seems like a not so good idea.
> 
> P.S. it could be quite interesting to add more measures of distance, for example [Ruler distance](https://blog.mapbox.com/fast-geodesic-approximations-with-cheap-ruler-106f229ad016)

Hey @glevv thanks for the suggestion.

If I understood this blog correctly, it has 3 computations: euclidean, harvesine (the one we are trying to implement here) and a more complicated one that has a smaller error (vincenty's formula). Is this correct?

I'd suggest we stick to harvesine in this PR, and see if we create an issue to expand the class later with the Vincenty's. Is this formula commonly used? do we really need an error as small as 0.5mm for geo coordinates?

glevv:
> > Is it possible to calculate Haversine distance using sklearn? It is quite fast and well optimized, reimplementing it seems like a not so good idea.
> > P.S. it could be quite interesting to add more measures of distance, for example [Ruler distance](https://blog.mapbox.com/fast-geodesic-approximations-with-cheap-ruler-106f229ad016)
> 
> Hey @glevv thanks for the suggestion.
> 
> If I understood this blog correctly, it has 3 computations: euclidean, harvesine (the one we are trying to implement here) and a more complicated one that has a smaller error (vincenty's formula). Is this correct?

They are all measures of distance between two points on ellipsoid. There were no Vincenty formula, but it's quite heavy to compute. In this particular blog post they talked about two simpler and faster formulas (Cheap Ruler and FCC equation) but with higher error.

> I'd suggest we stick to harvesine in this PR, and see if we create an issue to expand the class later with the Vincenty's. Is this formula commonly used? do we really need an error as small as 0.5mm for geo coordinates?

Ye, let's go with haversine only, not sure about Vincenty tho

VascoSch92:
Hey @solegalli 
Sorry if I disappeared. I had a lot to do with work and life. I will try to give a look at this pull request next week ,-)

solegalli:
No Problem at all @VascoSch92 . Same here.

I am doing some big changes to the correlation transformers, I think we could release a new version when i got those finished, hopefully during February.

It would be great if we can squeeze this transformer in that release 2. If you find the time, we look forward to your contribution :)

VascoSch92:
> No Problem at all @VascoSch92 . Same here.
> 
> I am doing some big changes to the correlation transformers, I think we could release a new version when i got those finished, hopefully during February.
> 
> It would be great if we can squeeze this transformer in that release 2. If you find the time, we look forward to your contribution :)

Hey @solegalli :-) 
is it time to give another try to this transformer? What do you think?

solegalli:
Sure! Contributions are welcome any time :)

VascoSch92:
ok perfect. I will work on it.

VascoSch92:
Hey @solegalli 
finally I have something. 

I still need some guidance for some point:
- From which classes should I extend? I'm extending from `BaseNumericalTransformer`, `FitFromDictMixin` and `GetFeatureNamesOutMixin` but I don't know if is a good idea
- I have a `fit` method also if I'm not using it. Should I have it anyway or can I delete it?

feature_engine/creation/distance_features.py
```diff
@@ -0,0 +1,241 @@
...
+            column_name=output_column_names,
+            coordinate_columns=coordinate_columns,
+        )
+
+        self.drop_original = self._check_drop_original(parameter=drop_original)
+
+        self.variables = None
+
+    def _check_drop_original(self, parameter: bool) -> bool:
+        if isinstance(parameter, bool) is False:
```
solegalli, line 112: For consistency with the rest of our codebase, let's check the init allowed values without extra functions.

feature_engine/creation/distance_features.py
```diff
@@ -0,0 +1,241 @@
...
+            raise ValueError(
+                "Expected boolean value for parameter `drop_original`, "
+                f"but got {parameter} with type {type(parameter)}"
+            )
+        return parameter
+
+    def _check_coordinate_columns(
+        self, columns: List[List[Union[str, int]]]
```
solegalli, line 120: Is this list of lists?  I think it is just a list.

feature_engine/creation/distance_features.py
```diff
@@ -0,0 +1,241 @@
...
+    ) -> Tuple[List[Union[str, int]], ...]:
+        if not columns:
+            raise ValueError("Empty list for `coordinate_columns` not allowed!")
```
solegalli, line 123: this test should be in the init, and actually, defaulting the parameter to None, would raise this error automatically. So we don't need it.

feature_engine/creation/distance_features.py
```diff
@@ -0,0 +1,241 @@
...
+        for idx, coordinate_column in enumerate(columns):
+            if len(coordinate_column) != 4:
```
solegalli, line 125: this test needs to be in the init, straightaway without a function.

feature_engine/creation/distance_features.py
```diff
@@ -0,0 +1,241 @@
...
+                raise ValueError(
+                    f"Needed 4 values to compute a distance, "
+                    f"but got {len(coordinate_column)} columns \n"
+                    f"at the index {idx} of the list coordinate columns."
+                )
+        return (
+            [coordinate[0] for coordinate in columns],
```
solegalli, line 132: I think we could just unpack: var1, var2, var3, var4  = self.coordinates parameter. And we should do this in fit.

feature_engine/creation/distance_features.py
```diff
@@ -0,0 +1,241 @@
...
+    def __init__(
+        self,
+        coordinate_columns: List[List[Union[str, int]]],
+        output_column_names: Optional[List[Union[str, None]]] = None,
+        drop_original: bool = False,
+    ) -> None:
+
+        # the coordinate_columns variable is rewritten in this way to speed up
+        # computation later, i.e., to use vectorization
+        (self.a_latitudes, self.a_longitudes, self.b_latitudes, self.b_longitudes) = (
```
solegalli, line 99: In the init, we check that the user enters allowed values, and then we assign those values to the parameters. We can't have any more functionality than that. 

The unpacking needs to be in the fit() part of the class.

feature_engine/creation/distance_features.py
solegalli: variables is a parameter that we have in all transformers, so I would stick to this name instead of using `coordinate_columns`

In short, let's replace coordinate_columns by variables.

feature_engine/creation/distance_features.py
```diff
@@ -0,0 +1,241 @@
...
+        )
+
+    def _check_output_columns_names(
+        self,
+        column_name: Optional[List[Union[str, None]]],
+        coordinate_columns: List[List[Union[str, int]]],
+    ) -> Optional[List[Union[str, None]]]:
+        if column_name is None:
+            return [f"distance_{c[0]}_{c[1]}_{c[2]}_{c[3]}" for c in coordinate_columns]
+        if len(column_name) != len(coordinate_columns):
```
solegalli, line 145: I don't understand this check, we are returning only 1 distance variable, so we don't need to check how many names, or we need to check that it is only 1 name

feature_engine/creation/distance_features.py
```diff
@@ -0,0 +1,241 @@
...
+            raise ValueError(
+                "Not enough output column names provided.\n "
+                f"Expected {len(coordinate_columns)} column names, "
+                f"but got {len(column_name)}."
+            )
+        return column_name
+
+    def fit(self, X: pd.DataFrame):
+        # there is no fit for this transformer
+        super().fit(X)
```
solegalli, line 155: here we need to unpack the coordinate variables.

feature_engine/creation/distance_features.py
```diff
@@ -0,0 +1,241 @@
...
+        X: Pandas DataFrame of shame = [n_samples, n_features]
+            The data to be transformed.
+
+        Returns
+        -------
+        X_new: Pandas dataframe.
+            The original dataframe plus the distances between the given coordinates.
+        """
+        X = self._check_transform_input_and_state(X)
+        X = self._check_lat_lon_columns_are_in_df(X)
```
solegalli, line 173: we don't need this test. The former test has it covered.

feature_engine/creation/distance_features.py
```diff
@@ -0,0 +1,241 @@
...
+                    *self.a_longitudes,
+                    *self.b_latitudes,
+                    *self.b_longitudes,
+                ],
+                inplace=True,
+            )
+
+        return X
+
+    def _check_lat_lon_columns_are_in_df(self, X) -> pd.DataFrame:
```
solegalli, line 191: delete

feature_engine/creation/distance_features.py
```diff
@@ -0,0 +1,241 @@
...
+            *self.b_longitudes,
+        ]
+        if set(coordinate_columns).issubset(set(X.columns)) is False:
+            raise ValueError(
+                f"The columns {set(coordinate_columns).issubset(set(X.columns))} "
+                f"were not found in the dataframe."
+            )
+        return X
+
+    def _check_correctness_of_coordinates(self, X: pd.DataFrame) -> pd.DataFrame:
```
solegalli, line 205: we need this check also in fit to control the input variables in the training set.

feature_engine/creation/distance_features.py
```diff
@@ -0,0 +1,241 @@
...
+        # while longitudes is between -180 and +180.
+        irregular_latitudes = (
+            (X[[*self.a_latitudes, *self.b_latitudes]].abs() > 90).sum().sum()
+        )
+        irregular_longitudes = (
+            (X[[*self.a_longitudes, *self.b_longitudes]].abs() > 180).sum().sum()
+        )
+
+        if irregular_latitudes > 0:
+            raise ValueError("The dataframe contains irregular latitudes")
```
solegalli, line 216: User may not be familiar with what irregular latitudes are. We need more informative errors. If the data has irregular lats and longs, this error will come first, and we will not let the user know that they also have irregular lons. We could unify the error and return both error messages together

solegalli (commented):
Hey @VascoSch92 

This is looking good. We need to tidy the code a bit. In the init, we can only test that the user enters allowed inputs, and then assign them. We don't need extra functions/methods for this checks. 

In the fit, we need to unpack the variables and check the allowed values.

If you could re-arrange that, I can then take another look.

It would be great if you could rebase main after you sync it to the latest version of the repo, because we made a lot of changes since this PR.

Thank you!

feature_engine/creation/distance_features.py
VascoSch92: I will do in fit. But I need to unpack in this way because I need lat and Lon in separate lists for convenience and not a 4-tuple of `lat_a, Lon_a, lat_b, Lon_b`

PR #698, opened 2023-09-18
Add missing_only init param to all imputers. missing_only is used in combination w/ the variables param.

### LINKED ISSUE #388: add missing_only=True to all imputers to use in combination with variables=None
Add missing_only functionality to all imputers to use in combination with variables=None

When variables is None, the imputers select all numerical, or categorical or all variables by default. With the missing_only, it would select only those from each subgroup that show missing data during fit.

Morgan-Sell:
Closes #388.

When `missing_only=True` and `variables=None`, all numerical variables that do **not** contain missing values will be omitted from `self.variables_`.

This functionality does not apply to categorical variables.

### DIFF
--- a/feature_engine/_docstrings/init_parameters/imputers.py
+++ b/feature_engine/_docstrings/init_parameters/imputers.py
@@ -1,0 +1,6 @@
+_missing_only_docstring = """missing_only: bool, default=False
+        If `missing_only` equals `True` and `variables` equals `None` then the transformer 
+        will only select the numerical variables that show missing values during `fit()`. 
+        All categorical variables, even those with no missing values, will be returned even 
+        when `missing_only` equals `True`.
+"""
--- a/feature_engine/imputation/arbitrary_number.py
+++ b/feature_engine/imputation/arbitrary_number.py
@@ -8,6 +8,7 @@ import pandas as pd
 from feature_engine._check_input_parameters.check_input_dictionary import (
     _check_numerical_dict,
 )
+
 from feature_engine._docstrings.fit_attributes import (
     _feature_names_in_docstring,
     _imputer_dict_docstring,
--- a/feature_engine/imputation/base_imputer.py
+++ b/feature_engine/imputation/base_imputer.py
@@ -9,6 +9,16 @@ from feature_engine.tags import _return_tags
 
 class BaseImputer(BaseEstimator, TransformerMixin, GetFeatureNamesOutMixin):
     """shared set-up checks and methods across imputers"""
+    def __init__(
+            self,
+            missing_only: bool = False,
+    ):
+        if not isinstance(missing_only, bool):
+            raise ValueError(
+                f"missing_only must be a boolean type. Got {missing_only} instead."
+            )
+
+        self.missing_only = missing_only
 
     def _transform(self, X: pd.DataFrame) -> pd.DataFrame:
         """


Morgan-Sell:
hi @solegalli,

When `missing_only=True` and `variables=None`, how should this impact the transformed dataset after `transform()` is applied? Should the `transform()` method only return those variables that were identified as having missing values? Or, should `transform()` return the complete dataset?

I'm guessing the latter because I can't think of a scenario in which we would want to eliminate variables that do not have missing values.

If my intuition is correct, is the sole purpose of this new feature for computational efficiency? If not, what are the other objectives of this new feature?

Morgan-Sell:
hola @solegalli, are you on vacation? Just wanted to see if you saw my question. No rush! I'm accustomed to your rapid responses, so I thought I check-in.

solegalli:
Hey @Morgan-Sell 
I was indeed on vacation. But the most funny thing is that I did see the question before, and I thought I answered it. LOL. Sorry about that.

`transform()` in feature engine always returns the complete dataset, unless the transformation adds variables, in which case it would return the complete dataframe plus the new variables. 

The aim is not to expand the feature space unnecessarily. If you add indicators for all variables, not just those with NA, you will end up with a lot of variables that just contain the value 0. Is this what you mean by efficiency?

Cheers
Sole

Morgan-Sell:
@solegalli, welcome back! Hope you enjoyed your "holiday" as they say on your side of the pond ;)

Based on your explanation, it seems that the `missing_only` param solely applies to the AddMissingIndicator class. The other imputers do not add additional variables. They fill in the missing values of the existing variables with the desired value(s).

If this is the case, then should this functionality only reside in the AddMissingIndicator class?

solegalli:
Its most meaningful / useful function is indeed for the MissingIndicator class. But we should add it in all imputers. It will play a role when variables =None. At the moment, when `variables=None` the transformers will learn imputation parameters for all numerical or categorical variables. If we add `missing_only=True`, they will learn parameters only for those numerical or categorical with NA. More efficient, smaller dictionaries stored, and avoids unexpected imputations when the model is live.

Morgan-Sell:
@solegalli,

I added `missing_value`, its functionality, and appropriate tests to the ArbitraryNumberImputer and CategorialEncoder classes. Before implementing the changes in all imputers, will you please validate or nullify what I've done thus far?

feature_engine/_docstrings/init_parameters/imputers.py
```diff
@@ -0,0 +1,6 @@
+_missing_only_docstring = """missing_only: bool, default=False
+        If `missing_only` equals `True` and `variables` equals `None` then the transformer 
+        will only select the numerical variables that show missing values during `fit()`.
```
solegalli, line 3:
```suggestion
        will add missing indicators to the variables that show missing values during `fit()`. 
```

feature_engine/_docstrings/init_parameters/imputers.py
```diff
@@ -0,0 +1,6 @@
+_missing_only_docstring = """missing_only: bool, default=False
+        If `missing_only` equals `True` and `variables` equals `None` then the transformer 
+        will only select the numerical variables that show missing values during `fit()`. 
+        All categorical variables, even those with no missing values, will be returned even
```
solegalli, line 4:
```suggestion
```

feature_engine/imputation/__init__.py
```diff
@@ -3,6 +3,7 @@
 """
 
 from .arbitrary_number import ArbitraryNumberImputer
+from .base_imputer import BaseImputer
```
solegalli, line 6:
```suggestion
```

feature_engine/imputation/__init__.py
```diff
@@ -18,4 +19,5 @@
     "AddMissingIndicator",
     "RandomSampleImputer",
     "DropMissingData",
+    "BaseImputer",
```
solegalli, line 22:
```suggestion
```

feature_engine/imputation/arbitrary_number.py
```diff
@@ -154,6 +163,11 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
             self.imputer_dict_ = self.imputer_dict
         else:
             self.variables_ = find_or_check_numerical_variables(X, self.variables)
+
+            # identify variables with missing values
+            if self.missing_only and self.variables is None:
```
solegalli, line 168:
```suggestion
            if self.missing_only is True and self.variables is None:
```

feature_engine/imputation/base_imputer.py
```diff
@@ -9,6 +9,16 @@
 
 class BaseImputer(BaseEstimator, TransformerMixin, GetFeatureNamesOutMixin):
     """shared set-up checks and methods across imputers"""
+    def __init__(
+            self,
+            missing_only: bool = False,
+    ):
+        if not isinstance(missing_only, bool):
+            raise ValueError(
+                f"missing_only must be a boolean type. Got {missing_only} instead."
```
solegalli, line 18:
```suggestion
                f"missing_only must be a boolean. Got {missing_only} instead."
```

feature_engine/imputation/categorical.py
```diff
@@ -175,6 +185,10 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
             # select all variables or check variables entered by the user
             self.variables_ = find_all_variables(X, self.variables)
 
+        # identify variables with missing values
+        if self.missing_only and self.variables is None:
```
solegalli, line 189: same as previous, we need to combine this with find_all_variables so it doesnt search twice.

feature_engine/imputation/__init__.py
solegalli: In the init we include the files that are for external use. BaseImputer is for internal use.

solegalli (commented):
Hey @Morgan-Sell !

Thank you so much for the changes. This is well ahead in the development :)

I made a few code suggestions and added a few comments, so here I summarize the main things:

- for numerical imputers, when `variables=None` and missing=True, it selects all **numerical** variables with nan
-  for categorical imputers, when variables=None and missing=True, it selects all **categorical** variables with nan
- for the transformer that adds missing indicators, or the random sample imputer, when variables=None and missing=True, it selects all **all** variables with nan

I think the function that you created needs a bit of refinement.

feature_engine/imputation/arbitrary_number.py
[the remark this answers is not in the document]
Morgan-Sell: Yes,  but we need line 165 to identify and check for the numerical variables first. Note how line 169 references `self.variables_` not `self.variables`.

If we write it as 

```
elif self.missing_only and self.variables is None:
     self.variables_ = find_variables_with_missing_values(X, self.variables)
```

`self.variables_` will include both numerical and categorical variables that have missing values. We only want the numerical variables.

We could do the following:

```
elif self.missing_only and self.variables is None:
     self.variables_ = find_or_check_numerical_variables(X, self.variables)
     self.variables_ = find_variables_with_missing_values(X, self.variables_)
```

However, the code is repeating `find_or_check_numerical_variables()`.

feature_engine/imputation/categorical.py
Morgan-Sell: See my prior response. By implementing the elif approach, we'll encounter a similar issue. This time we only want categorical variables, but if we create a new elif then we'll have both categorical and numerical variables

Morgan-Sell:
hi @solegalli,

I responded to two of your comments. Once we agree on the approach, we will quickly move forward with this PR. It's, more or less, copying/pasting the agreed-upon approach on all transformers.

feature_engine/imputation/arbitrary_number.py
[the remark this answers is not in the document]
solegalli: Can we think of a way in which we do not need to scan for variables twice? Maybe we need 2 functions to find vars with NA, one for categorical and one for numerical?

Morgan-Sell:
hi @solegalli, i created two global functions to identify either categorical or numerical variables that have missing values. These functions incorporate the "check_or_find_all..." functions.

I have one question about the ArbitraryImputer's fit() order of operations.

In meantime, I will implement the new code in the other imputation transformers.

Morgan-Sell:
hi @solegalli,

AddMissingIndicator() already has the `missing_only` param. I don't recall coding it ;)

Should I refactor the code so the AddMissingIndicator() code base has a similar structure to the other imputers? Or, keep the code as is?

If you have time, take a look at the other imputers. We're almost there!!!

Morgan-Sell:
hi @solegalli,

Did you see my above questions regarding the AddMissingIndicator() class?

PR #710, opened 2023-11-08
Add py.typed file

david-cortes:
This PR adds a `py.typed` file for the package, which signals to `mypy` that the library contains types that can be checked during checks/tests in user code.

This is useful for example to use classes from `feature_engine` in type hints and check that the types match when executing that code.


solegalli:
Hey @david-cortes 

Thanks for the PR. I didn't know about this functionality. Is this for users of Feature-engine? or for feature-engine developers? Do you have by any chance a code snippet where this would be useful? or a link?

Thanks!

david-cortes:
> Hey @david-cortes
> 
> Thanks for the PR. I didn't know about this functionality. Is this for users of Feature-engine? or for feature-engine developers? Do you have by any chance a code snippet where this would be useful? or a link?
> 
> Thanks!

This is for users of the package who wish to run `mypy` on code that uses it.

For example, suppose you create a  file `test_transformers.py` like this:
```python
from feature_engine.preprocessing import MatchVariables

def get_transformer() -> MatchVariables:
    return MatchVariables()

def test_get_transformer():
    tr = get_transformer()
    assert True
```

Then running it with `pytest test_transformers.py` will succeed as expected.

But running it with `pytest --mypy test_transformers.py` will error out:
```
======================================== FAILURES =========================================
__________________________________ test_transformers.py ___________________________________
1: error: Skipping analyzing "feature_engine.preprocessing": module is installed, but missing library stubs or py.typed marker  [import-untyped]
1: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
______________________________________ test session _______________________________________
mypy exited with status 1.
========================================== mypy ===========================================
Found 1 error in 1 file (checked 1 source file)
================================= short test summary info =================================
FAILED test_transformers.py::mypy
FAILED test_transformers.py::mypy-status
=============================== 2 failed, 1 passed in 1.14s ===============================
```

Adding this file allows checking the type hints that involve classes from this package.

solegalli:
Super thank you! That's very helpful :)

PR #721, opened 2024-01-16
speed-up correlation selection transformers, make them deterministic

### LINKED ISSUE #446: [ENHANCEMENT] change pandas.corr() by numpy alternative in correlation selectors

For the following feature selectors:
- [SmartCorrelatedSelection](https://feature-engine.readthedocs.io/en/1.3.x/api_doc/selection/SmartCorrelatedSelection.html)
- [DropCorrelatedFeatures](https://feature-engine.readthedocs.io/en/1.3.x/api_doc/selection/DropCorrelatedFeatures.html)

`pandas.corr(method='pearson')` is currently being used to calculate the correlation matrix. This is very slow in performance compared to using the `numpy` equivalent method.

An example of where `pandas.corr(method='pearson')` is used can be found [here](https://github.com/feature-engine/feature_engine/blob/1b558821a58552bfb1addf96a86d04302ff619c2/feature_engine/selection/smart_correlation_selection.py#L239).

Refer to here for specific code change: [smart_correlation_selection.py](https://github.com/feature-engine/feature_engine/blob/1b558821a58552bfb1addf96a86d04302ff619c2/feature_engine/selection/smart_correlation_selection.py#L239)

Instead of this:

```python
 _correlated_matrix = X[self.variables_].corr(method=self.method)
```

We can do this instead:

```python
if self.method == 'pearson':
    _correlated_matrix = pd.DataFrame(np.corrcoef(X[self.variables_], rowvar=False), columns=X[self.variables_].columns)
    _correlated_matrix.index = _correlated_matrix.columns
```

**Additional context**
I currently haven't looked into how to make the other methods faster (e.g.` method=spearman`), but this can also be considered in this feature.

### LINKED ISSUE #570: SmartCorrelatedSelection  may not replicate the same result
Hi, 

I've found this issue when I'm doing experiment on some datasets.

**Describe the bug**
`SmartCorrelatedSelection` may not replicate the same result for some datasets.
This is because `set` is unordered data structure, and thus `.add()`  would not preserve the order.

When some features match the same score by selection method, the result would be different.

It happened only when I restart my development environment. (otherswise, the result might be the same.)

 I have solved this by transforming set to  list: `_temp_set = list(set([feature]))` and replace `.add(f2)` to `_temp_set.append(f2)` in `SmartCorrelatedSelection.py`

**To Reproduce**
1. Run `SmartCorrelatedSelection` on some dataset which has plenty correlated features. 
(In my case, I assign `selection_method="variance"`)
2. record `features_to_keep_`
3. restart your environment. (including `PyThon`)
4. Run again and you will see different `features_to_keep_` result

**Expected behavior**
Drop the same features when the given parameters is the same. 

NAME="CentOS Linux"
VERSION="8"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="8"
PLATFORM_ID="platform:el8"
PRETTY_NAME="CentOS Linux 8"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:8"
HOME_URL="https://centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"
CENTOS_MANTISBT_PROJECT="CentOS-8"
CENTOS_MANTISBT_PROJECT_VERSION="8"

 Thank you for developing this wonderful tool! 🌟

### LINKED ISSUE #612: Correlation-based feature selection is column-order dependent

### LINKED ISSUE #684: SmartCorrelatedSelection inconsistency in features to drop
**Describe the bug**
The SmartCorrelatedSelection from feature engine version 1.5.2 is not reproducing the same result after I restart the kernel. I have done a dummy test on the titanic dataset where one hot encoding has been done and then I am searching for the correlated features to drop using the variance method from the documentation.

**Screenshots**
In the initial run of the notebook, I get the following results:

image

After restarting and re-running the notebook, I get a second set of results as follows:

image

I have ensured that "encoded_data" is the same reproducible dataset after restarting the kernel, the transformers used here are the ones from feature_engine to impute and encode respectively.

solegalli:
closes #619
closes #648 
closes #633 
closes #612
closes #446 
closes #684 
closes #570 
closes #703 

FYI @dlaprins @glevv 

Sorts variables alphabetically to avoid different results.
Adds adaptation of @dlaprins numpy implementation to speed search.

I'd like to keep the brute force correlation plain simple. And then we add the ordering per cardinality, etc to the smart correlation, which sort of has it already.

- [x] speed up search in smartcorrelationselection
- [x] check ordering implementation in smartcorrelationselection
- [x] take care of transitivity in smart corr selection
- [x] add tests for find_single_features
- [x] fix tests for smart correlation selection - need to create specific dfs to control for variance, cardinality and missing data
- [x] sort out unrelated error (newer versions of some libraries)
- [x] update correlation documentation (rst files in user_guide)
- [x] expand user guide on smart correlation selection
- [x] improve tests for recursive feature elimination (they are failing at the moment)
- [x] check python 3.11 and 3.12 images on circle ci
- [ ] check check_classification_targets and why it fails for decision tree encoder

The unrelated error is in the recursive feature selectors. Something change in the estimators, or random state or something that now the values are different. Need to make that deterministic somehow.


feature_engine/selection/drop_correlated_features.py
```diff
@@ -177,52 +187,35 @@ def fit(self, X: pd.DataFrame, y: pd.Series = None):
...
-                    # if the correlation is higher than the threshold
-                    # we are interested in absolute correlation coefficient value
-                    if abs(_correlated_matrix.loc[f2, feature]) > self.threshold:
-
-                        # add feature (f2) to our correlated set
-                        self.features_to_drop_.add(f2)
-                        _temp_set.add(f2)
-                        _examined_features.add(f2)
+        # the correlated pairs
+        correlated_mask = (abs(correlated_matrix) > self.threshold) - np.eye(
```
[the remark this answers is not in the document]
solegalli: Yep, it works. Thanks. Would this be faster? or more memory efficient?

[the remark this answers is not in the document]
glevv: It will take upper triangle of the square matrix without the main diagonal (the one with the 1's). Should reduce the number of feature combinations to iterate over

solegalli:
Observations:

I could take the opportunity to speed up the selection tests for the recursive feature elimination, those are the ones failing.

The documentation for the smart correlation selector, could also be improved by showcasing the other methodologies.

.circleci/config.yml
```diff
@@ -60,9 +48,33 @@ jobs:
           name: Run tests with Python 3.10
           command: |
             tox -e py310
+  test_feature_engine_py311:
+    docker:
+      - image: cimg/python:3.11.7
```
solegalli, line 53: @ChristopherGS any obvious idea why this is failing?

.circleci/config.yml
```diff
@@ -60,9 +48,33 @@ jobs:
...
+      - checkout:
+          path: ~/project
+      - *prepare_tox
+      - run:
+          name: Run tests with Python 3.11
+          command: |
+            tox -e py311
+  test_feature_engine_py312:
+    docker:
+      - image: cimg/python:3.12.1
```
solegalli, line 65: @ChristopherGS same here?

.circleci/config.yml
ChristopherGS: @solegalli you need to add the test environment to the tox.ini: https://github.com/feature-engine/feature_engine/blob/main/tox.ini#L25

Tutorial: https://christophergs.com/python/2020/04/12/python-tox-why-use-it-and-tutorial/

solegalli: Life saver, thank you so much @ChristopherGS !

PR #724, opened 2024-02-18
add method to transfrom x and y to classes that drop rows from X, add pipeline to handle transform_x_y

solegalli:
- [x] add test for transform x y mixin
- [x] fix test window features
- [x] update expanding windows transformer
- [x] add test for expanding windows
- [x] create pipeline
- [x] test pipeline
- [x] expand doc dropna
- [x] expand doc outlier trimmer
- [x] expand doc lag features
- [x] expand doc window features
- [x] expand doc expanding window features
- [x] create doc pipeline


feature_engine/timeseries/forecasting/base_forecast_transformers.py
```diff
@@ -38,7 +41,9 @@
     fit=_fit_not_learn_docstring,
     n_features_in_=_n_features_in_docstring,
 )
-class BaseForecastTransformer(BaseEstimator, TransformerMixin, GetFeatureNamesOutMixin):
+class BaseForecastTransformer(
+    BaseEstimator, TransformerMixin, GetFeatureNamesOutMixin, TransformXyMixin
+):
```
VascoSch92, line 46: nit:
```python
class BaseForecastTransformer(
     BaseEstimator, 
     TransformerMixin, 
     GetFeatureNamesOutMixin, 
     TransformXyMixin,
 ):
```

tests/test_base_transformers/test_transform_xy_mixin.py
```diff
@@ -0,0 +1,22 @@
...
+class MockTransformer(TransformXyMixin):
+    def transform(self, X):
+        return X.iloc[1:-1].copy()
+
+
+def test_transform_x_y_method(df_vartypes):
+    y = pd.Series(pd.Series(0, index=np.arange(len(df_vartypes))))
+    transformer = MockTransformer()
+    Xt, yt = transformer.transform_x_y(df_vartypes, y)
+    print(Xt)
```
VascoSch92, line 16: I suppose the print was just for debugging :-)

tests/test_time_series/test_forecasting/test_lag_features.py
```diff
@@ -214,6 +214,131 @@ def test_correct_lag_when_using_freq(df_time):
...
+    df_tr = transformer.fit_transform(df)
+    assert df_tr.equals(expected_df)
+
+
+def test_transform_x_y(df_time):
+    df = df_time.head(5).copy()
+    y = pd.Series(np.zeros(len(df)), index=df.index)
+
+    # Expected
+    date_time = [
```
VascoSch92, line 310: date_time and expected_results are equal to the one in the previous test. Right?

If yes, you could use a fixture

tests/test_base_transformers/test_transform_xy_mixin.py
solegalli: yes, thank you!

solegalli:
Tests for new pipeline functionality are done. 

I need to add remaining tests to corroborate that the original functionality of the pipeline (inherited from sklearn) also continues to work.

feature_engine/pipeline/pipeline.py
```diff
@@ -442,3 +442,51 @@ def fit_predict(self, X, y=None, **params):
...
             )
         return y_pred
+
+
+def make_pipeline(*steps, memory=None, verbose=False):
+    """Construct a :class:`Pipeline` from the given estimators.
+
+    This is a shorthand for the :class:`Pipeline` constructor; it does not
+    require, and does not permit, naming the estimators. Instead, their names
+    will be set to the lowercase of their types automatically.
```
VascoSch92, line 452: Just to know:
Can we have other types outside string that can be set to lowercase?

tests/test_pipeline/test_pipeline_all.py
```diff
@@ -0,0 +1,842 @@
...
+    "the burger beer beer copyright",
+    "the coke burger coke copyright",
+    "the coke burger burger",
+)
+
+R_TOL = 1e-4
+
+def test_pipeline_invalid_parameters():
+    test_pipeline_invalid_parameters
+
```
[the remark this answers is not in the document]
ChristopherGS: @solegalli copying tests from another library like that is generally not best practice because:

- The test functions are not typically designed to be part of a library's public API. This means they might not be as stable or intended for external use
- If scikit-learn updates their tests or implementation in a way that is incompatible with your extension, your tests might suddenly fail. Yes you can mitigate that by pinning the version, but I doubt you want to do that and miss out on patches

I'd just write your own tests, even if it means lifting and dropping the code

PR #729, opened 2024-03-13
refactor: exposing the unseen var of the categorical encoder

### LINKED ISSUE #588: replace attribute encoding pipeline for encoding dictionary

### LINKED ISSUE #728: Parametrize unseen for the OrdinalEncoder in DecisionTreeEncoder
I started to use the [DecisionTreeEncoder](https://github.com/feature-engine/feature_engine/blob/main/feature_engine/encoding/decision_tree.py) for categorical variables. Lately I ran into an exception and I found out it was about when in there is an unseen category. 
I understand this as a default behaviour, but I was thinking that having the flexibility to decide that in the constructor can be helpful.   

To give more context:
From [the code](https://github.com/feature-engine/feature_engine/blob/main/feature_engine/encoding/decision_tree.py#L245) I saw that the `OrdinalEncoder` has by default the unseen variable set to `"raise"`
```python 
        cat_encoder = OrdinalEncoder(
            encoding_method=self.encoding_method,
            variables=variables_,
            missing_values="raise",
            ignore_format=self.ignore_format,
            unseen="raise",
        )
```
**Suggestion**

I would like to suggest to make that param something more flexible having it in the constructor 
```python 
     
    def __init__(
        self,
        cv=3,
        regression: bool = True,
        ignore_format: bool = False,
        # added var
        unseen: str = 'raise',
    ) -> None:

        super().__init__(variables, ignore_format)
        self.cv = cv
        self.scoring = scoring
        self.regression = regression
        self.param_grid = param_grid
        self.unseen = unseen


[ ... other code ...]



   cat_encoder = OrdinalEncoder(
            encoding_method=self.encoding_method,
            variables=variables_,
            missing_values="raise",
            ignore_format=self.ignore_format,
            unseen=self.unseen,
        )
```

Of course there are alternatives, in my project I am currently overriding the function to take into account that behaviour, but I think that this flexibility might help others .
In case it is found a reasonable change I would be happy to help on making the change

93lorenzo:
closes #728
closes #588 

Exposing the `unseen` parameter for the `DecisionTreeEncoder`.
Changing the encoder_ pipeline for an encoding_dict_ that maps from categories to numerical mappings (the predictions of the tree)

### DIFF
--- a/feature_engine/encoding/decision_tree.py
+++ b/feature_engine/encoding/decision_tree.py
@@ -347,7 +347,7 @@ class DecisionTreeEncoder(CategoricalInitMixin, CategoricalMethodsMixin):
         if self.unseen == "encode":
             X = X.mask(mask_unseen, other=self.fill_value)
         if self.unseen == "ignore":
-            X = X.mask(mask_unseen, other=None)
+            X = X.mask(mask_unseen, other=np.nan)
 
         return X
 
--- a/feature_engine/encoding/decision_tree.py
+++ b/feature_engine/encoding/decision_tree.py
@@ -32,9 +32,8 @@ from feature_engine.tags import _return_tags
 
 _unseen_docstring = (
     _unseen_docstring
-    + """ `'ignore'` will then make the DecisionTreeDiscretiser throw an error.
-        If `'encode'` unseen categories will be encoded with fill_value param,
-        that must be a numeric value.Any other value will throw an error.
+    + """ If `'encode'` unseen categories will be encoded with fill_value param,
+        that must be a numeric value.
         """
 )
 
@@ -330,15 +329,19 @@ class DecisionTreeEncoder(CategoricalInitMixin, CategoricalMethodsMixin):
         X = self._check_transform_input_and_state(X)
         _check_contains_na(X, self.variables_)
 
-        if self.unseen == "encode":
-            for column, col_values in self._categories.items():
-                X[column] = np.where(X[column].isin(col_values) | X[column].isna(),
-                                     X[column], self.fill_value)
-        elif self.unseen == "ignore":
-            for column, col_values in self._categories.items():
-                X[column] = np.where(X[column].isin(col_values) | X[column].isna(),
-                                     X[column], np.nan)
+        if self.unseen != "raise":
+            # replace unseen with a seen value
+            mask_unseen = ~X.apply(lambda x: x.isin(self._categories[x.name]))
+            for col, values in self._categories.items():
+                X.loc[mask_unseen[col],[col]] = values[0]
+
         X = self.encoder_.transform(X)
+
+        if self.unseen == "encode":
+            X = np.where(mask_unseen, self.fill_value, X)
+        elif self.unseen == "ignore":
+            X = np.where(mask_unseen, None, X)
+
         return X
 
     def inverse_transform(self, X: pd.DataFrame):
--- a/feature_engine/encoding/decision_tree.py
+++ b/feature_engine/encoding/decision_tree.py
@@ -331,7 +331,15 @@ class DecisionTreeEncoder(CategoricalInitMixin, CategoricalMethodsMixin):
 
         if self.unseen != "raise":
             # replace unseen with a seen value - later it will be replaced
-            mask_unseen = ~X.apply(lambda x: x.isin(self._categories[x.name]))
+            mask_unseen = ~X.apply(
+                lambda x: x.isin(
+                    self._categories[x.name]
+                    )
+                if x.name in list(
+                        self._categories.keys()
+                        )
+                else True
+                )
             for col, values in self._categories.items():
                 X.loc[mask_unseen[col], [col]] = values[0]
 
--- a/feature_engine/encoding/decision_tree.py
+++ b/feature_engine/encoding/decision_tree.py
@@ -4,6 +4,7 @@
 from typing import List, Optional, Union
 
 import pandas as pd
+import numpy as np
 from sklearn.pipeline import Pipeline
 from sklearn.utils.multiclass import check_classification_targets, type_of_target
 
@@ -25,6 +26,8 @@ from feature_engine.encoding.base_encoder import (
     CategoricalMethodsMixin,
 )
 from feature_engine.encoding.ordinal import OrdinalEncoder
+from feature_engine.encoding._helper_functions import check_parameter_unseen
+
 from feature_engine.tags import _return_tags
 
 
@@ -106,6 +109,24 @@ class DecisionTreeEncoder(CategoricalInitMixin, CategoricalMethodsMixin):
         of the parameters of the Scikit-learn's DecisionTreeRegressor() or
         DecisionTreeClassifier(). For reproducibility it is recommended to set
         the random_state to an integer.
+
+    unseen: str, default='raise'
+        The unseen param of the OrdinalEncoder used before DecisionTreeDiscretiser
+        in the fit method. It tells the encoder how to handle unseen categories.
+        Acceptable values are:
+
+            - If 'ignore', then unseen categories will be encoded as nan
+                - That might make the OrdinalEncoder throw an error
+
+            - If 'raise', then unseen categories will raise an error.
+
+            - If 'encode', unseen categories will be encoded with fill_value param.
+
+        Any other value will throw an error.
+
+    fill_value: int, default=None
+        The value used to fill for unseen categories. It is only used in
+        case of unseen=encode
 
     {variables}
 
@@ -198,8 +219,13 @@ class DecisionTreeEncoder(CategoricalInitMixin, CategoricalMethodsMixin):
         random_state: Optional[int] = None,
         variables: Union[None, int, str, List[Union[str, int]]] = None,
         ignore_format: bool = False,
-    ) -> None:
+        unseen: str = "raise",
+        fill_value: Optional[int] = None,
 
+    ) -> None:
+        check_parameter_unseen(unseen, ["ignore", "raise", "encode"])
+
+        check_parameter_unseen(unseen, ["ignore", "raise", "encode"])
         super().__init__(variables, ignore_format)
         self.encoding_method = encoding_method
         self.cv = cv
@@ -207,6 +233,24 @@ class DecisionTreeEncoder(CategoricalInitMixin, CategoricalMethodsMixin):
         self.regression = regression
         self.param_grid = param_grid
         self.random_state = random_state
+        self.unseen = unseen
+        self.fill_value = fill_value
+        self._fill_value_check()
+
+    def _fill_value_check(self,):
+        """
+        Check if unseen and fill_value inputs are acceptable.
+        """
+        if self.unseen == "encode":
+            if self.fill_value is None:
+                raise ValueError(
+                    "If unseen is encode fill_value must be an integer and not None"
+                )
+            if not isinstance(self.fill_value, int):
+                raise ValueError(
+                    f"fill_value must be an integer, "
+                    f"{self.fill_value} of type {type(self.fill_value)} found"
+                )
 
     def fit(self, X: pd.DataFrame, y: pd.Series):
         """
@@ -239,6 +283,15 @@ class DecisionTreeEncoder(CategoricalInitMixin, CategoricalMethodsMixin):
         variables_ = self._check_or_select_variables(X)
         _check_contains_na(X, variables_)
 
+        # dictionary of categories and seen values
+        if self.unseen in ["encode", "ignore"]:
+            self._categories = {}
+            for var in variables_:
+                unique_vars = X[var].drop_duplicates().to_list()
+                self._categories.update({
+                    var: unique_vars
+                })
+
         param_grid = self._assign_param_grid()
 
         # initialize categorical encoder
@@ -247,7 +300,7 @@ class DecisionTreeEncoder(CategoricalInitMixin, CategoricalMethodsMixin):
             variables=variables_,
             missing_values="raise",
             ignore_format=self.ignore_format,
-            unseen="raise",
+            unseen=self.unseen,
         )
 
         # initialize decision tree discretiser
@@ -291,6 +344,15 @@ class DecisionTreeEncoder(CategoricalInitMixin, CategoricalMethodsMixin):
         """
         X = self._check_transform_input_and_state(X)
         _check_contains_na(X, self.variables_)
+        # replace unseen values if unseen==encode
+        if self.unseen == "encode":
+            for column, col_values in self._categories.items():
+                X[column] = np.where(X[column].isin(col_values) | X[column].isna(),
+                                     X[column], self.fill_value)
+        elif self.unseen == "ignore":
+            for column, col_values in self._categories.items():
+                X[column] = np.where(X[column].isin(col_values) | X[column].isna(),
+                                     X[column], np.nan)
         X = self.encoder_.transform(X)
         return X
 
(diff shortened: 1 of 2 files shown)


glevv (approved):
LGTM

feature_engine/encoding/decision_tree.py
```diff
@@ -107,6 +107,18 @@ class DecisionTreeEncoder(CategoricalInitMixin, CategoricalMethodsMixin):
...
+        The unseen param of the OrdinalEncoder used before DecisionTreeDiscretiser
+        in the fit method. It tells the encoder how to handle unseen categories.
+        Following OrdinalEncoder we have :
+
+            -  If 'raise', then unseen categories will raise an error.
+
+            -  If 'ignore', then unseen categories will be encoded as NaN
+               and a warning will be raised instead.
+
+            -  If 'encode', unseen categories will be encoded as -1.
```
solegalli, line 120: Is this true? it will be replaced by -1 by the ordinal encoder, but the tree encoder creates a tree based on the ordinal numbers, so I am not sure this is what will happen. Did you test that?

We'd need a test for this parameter in the test, that it is able to take the strings and error when passing something else. And, ideally we also want another test to corroborate that the unseen functionality works as expected.

tests/test_selection/test_single_feature_performance.py
```diff
@@ -136,7 +134,7 @@ def test_raises_warning_if_no_feature_selected(load_diabetes_dataset):
         threshold=10,
     )
     with pytest.warns(UserWarning):
-        warnings.warn(sel.fit(X, y), UserWarning)
+        sel.fit(X, y)
```
solegalli, line 137: thanks for this fix, it was helpful for another PR that I am working on. I will merge that PR with this solution today, so I think we can remove this fix from this PR :)

feature_engine/encoding/decision_tree.py
93lorenzo: Answering first to your second point point :
* Yes I will create some tests for that! 

Answering to your first point: 
* I tested that and the tree encoder encodes the unseen ones in the same "encoding value"

EX:
```python 

df_enc = pd.DataFrame ( {
        "var_A": ["A"] * 6 + ["B"] * 10 + ["C"] * 4,
        "var_B": ["A"] * 10 + ["B"] * 6 + ["C"] * 4,
        "target": [1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 0, 0, 0, 0, 1, 1, 0, 0],
    })

variables_ = ["var_A", "var_B"]
param_grid = {"max_depth": [1, 2, 3, 4]}
X = df_enc[variables_]
y = df_enc['target']
unseen = 'raise'

# initialize categorical encoder
cat_encoder_encode = OrdinalEncoder(
    encoding_method="arbitrary",#self.encoding_method,
    variables=variables_,
    missing_values="raise",
    ignore_format=False, #self.ignore_format,
    unseen="encode" #self.unseen,
)

# initialize decision tree discretiser
tree_discretiser = DecisionTreeDiscretiser(
    cv=2, #self.cv,
    scoring="neg_mean_squared_error", #self.scoring,
    variables=variables_,
    param_grid=param_grid,
    regression=False, #self.regression,
    random_state=None #self.random_state,
)

# pipeline for the encoder
encoder_encode = Pipeline(
    [
        ("categorical_encoder", cat_encoder_encode),
        ("tree_discretiser", tree_discretiser),
    ]
)

encoder_encode.fit(X,y)

X_new = pd.DataFrame({
    "var_A":['A','B','C','ZZZ','YYY',-1],
    "var_B":['A','B','C','A'  ,'ZZZ',-1],
})

encoder_encode.transform(X_new)

>>>>
>>>>
>>>>
>>>>

```

93lorenzo: Answering first to your second point point :
* Sure I will create a test for the possible values of the param
* And a test to check that work as expected  

Answering to your first point: 
* I quickly tested, the tree encoder encodes all the unseen values in the same "encoding bucket". I am providing here a code snippet with the output. For the `var_A` it seems 0.25 the 'Other' bucket on which encode some categories and unseen values.

EX:
```python 

df_enc = pd.DataFrame ( {
        "var_A": ["A"] * 6 + ["B"] * 10 + ["C"] * 4,
        "var_B": ["A"] * 10 + ["B"] * 6 + ["C"] * 4,
        "target": [1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 0, 0, 0, 0, 1, 1, 0, 0],
    })

variables_ = ["var_A", "var_B"]
param_grid = {"max_depth": [1, 2, 3, 4]}
X = df_enc[variables_]
y = df_enc['target']
unseen = 'raise'

# initialize categorical encoder
cat_encoder_encode = OrdinalEncoder(
    encoding_method="arbitrary",#self.encoding_method,
    variables=variables_,
    missing_values="raise",
    ignore_format=False, #self.ignore_format,
    unseen="encode" #self.unseen,
)

# initialize decision tree discretiser
tree_discretiser = DecisionTreeDiscretiser(
    cv=2, #self.cv,
    scoring="neg_mean_squared_error", #self.scoring,
    variables=variables_,
    param_grid=param_grid,
    regression=False, #self.regression,
    random_state=None #self.random_state,
)

# pipeline for the encoder
encoder_encode = Pipeline(
    [
        ("categorical_encoder", cat_encoder_encode),
        ("tree_discretiser", tree_discretiser),
    ]
)

encoder_encode.fit(X,y)

X_new = pd.DataFrame({
    "var_A":['A','B','C','ZZZ','YYY',-1],
    "var_B":['A','B','C','A'  ,'ZZZ',-1],
})

encoder_encode.transform(X_new)

>>>> var_A var_B
>>>> 0.25  0.2
>>>> 0.25  0.2
>>>> 0.50  0.4
>>>> 0.25  0.2
>>>> 0.25  0.2
>>>> 0.25  0.2 

```

93lorenzo: Doing further checks raise both "raise" and "ignore" raise an exception.
"raise" --> it raises a ValueError for new values
"ignore" --> It passes the first part of the pipeline `OrdinalEncoder` (generating Nans for unseen) but it throws an error in the second part for the `DecisionTreeDiscretiser` because it is fed with Nan values which `DecisionTreeDiscretiser` does not allow .

I am planning to change the  `DecisionTreeDiscretiser` class to allow only `raise` and `encode` for the `unseen` param

feature_engine/encoding/decision_tree.py
```diff
@@ -116,9 +116,10 @@ class DecisionTreeEncoder(CategoricalInitMixin, CategoricalMethodsMixin):
 
             -  If 'raise', then unseen categories will raise an error.
 
-            -  If 'encode', unseen categories will be encoded as -1 by the 
+            -  If 'encode', unseen categories will be encoded as -1 by the
```
solegalli, line 119: Thanks for the fixes. Here are my thoughts regarding the replacement for unseen categories:

I don't think -1 is a good idea, because we don't really know what the output of the tree is going to be. If it is a probability, then -1 is a safe choice, because it clearly signals that this category was encoded by a different logic.

But if the target is continuous, then -1 might be a common output of the tree, and then we'd be treating this category like perhaps other group of categories that appear quite often in the data.

If the target is binary, perhaps we could leave -1. But if the target is continuous, I think it won't be possible for the API to understand what a would replacement value would be.

So my suggestion is: 

- add a `fill_value` parameter at init, defaults to `None`, where the user can enter the value they wish to use for the encoding (this is the logic currently used in sklearn ordinal encoder).
- then, add logic to have unseen categories replaced by the fill_value when unseen=encode
- if unseen equals raise or ignore, we should treat unseen as we normally do: raise throws an error, ignore i think replaces by nan, or leaves it as it is, I don;t remember any more, we should do whatever ordinaleconder does. 

At init, we need a check that `fill_value` is numeric, and another check to corroborate that `fill_value` is not none when `unseen = encode`.

Then we need checks for the values that `unseen` can take. I don't remember if we can inherit this from a parent class, we should check what ordinal encoder does and try to repeat it for this class.

Regarding the logic to replace the unseen categories:

- if raise, we just leave it to ordinalencoder to handle
- if ignore or encode, we need to add extra logic in this class to have those categories either replaced with nan (or whatever ordinalencoder does, or with the `fill_value`.

And then, we need tests for all the new logic, the tests at init, plus the replacements during transform.

What do you think? Do you think you could work on this?

Thank you!

solegalli: You could also rebase main now so that the tests can pass :)

93lorenzo: I understand your point! Indeed, I think what you are suggesting would the most flexible way. 
I will work on that this week :)

solegalli: awesome! thank you!

93lorenzo: Now that I am rereading the messages and the code I  have a question. 
With your phrase
<em>So my suggestion is:
*    add a fill_value parameter at init, defaults to None, where the user can enter the value they wish to use for the encoding (this is the logic currently used in sklearn ordinal encoder).
*   then, add logic to have unseen categories replaced by the fill_value when unseen=encode
*   if unseen equals raise or ignore, we should treat unseen as we normally do: raise throws an error, ignore i think replaces by nan, or leaves it as it is, I don;t remember any more, we should do whatever ordinaleconder does.</em>

You mean to have as a init param **fill_value** and **unseen** in the  `DecisionTreeEncoder`  (to forward them to the ordinal enc) and also in the `OrdinalEncoder`. To add within the  `OrdinalEncoder` class the logic of unseen fill_value ?
 
I am asking that, because it is still not 100% clear to me if you meant the change to happen only in the `DecisionTreeEncoder`. 

After having a look I feel the change must happen also in the `OrdinalEncoder` to have the flexibility for the  `OrdinalEncoder` to fill [unseen values with the passed param instead of always with -1](https://github.com/feature-engine/feature_engine/blob/main/feature_engine/encoding/ordinal.py#L224)


More than a question is a confirmation if what I understood (and tried to explain here) is aligned with what you meant  :)

solegalli: Hi @93lorenzo 

We would not be changing the OrdinalEnocoder in this PR. We would forward only the "raise" to the OrdinalEncoder's unseen, because it has that functionality already, but for "ignore" and "encode" we need to add the logic to the tree encoder. The fill_value will exist only in tree encoder as well.

93lorenzo: Hey @solegalli ! 
I did the changes! 
Please have a look and let me know if there is something to add/change or discuss :)

feature_engine/encoding/decision_tree.py
```diff
@@ -239,6 +281,15 @@ def fit(self, X: pd.DataFrame, y: pd.Series):
         variables_ = self._check_or_select_variables(X)
         _check_contains_na(X, variables_)
 
+        # dictionary of categories and seen values
+        if self.unseen in ["encode", "ignore"]:
```
solegalli, line 285: would it not be faster if self.unseen != "raise" ? instead of searching over a list?

feature_engine/encoding/decision_tree.py
```diff
@@ -239,6 +281,15 @@ def fit(self, X: pd.DataFrame, y: pd.Series):
...
+            self._categories = {}
+            for var in variables_:
+                unique_vars = X[var].drop_duplicates().to_list()
```
solegalli, line 288: is this trying to capture the unique categories per variable? can we not do X[var].unique(() instead?

the name unique_vars is confusing, should it not be unique_categories? or simply categories?

feature_engine/encoding/decision_tree.py
```diff
@@ -291,6 +342,15 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
         """
         X = self._check_transform_input_and_state(X)
         _check_contains_na(X, self.variables_)
+        # replace unseen values if unseen==encode
```
solegalli, line 345: I think this comment is self-explained in the logic, can we remove please?

feature_engine/encoding/decision_tree.py
```diff
@@ -291,6 +342,15 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+        if self.unseen == "encode":
+            for column, col_values in self._categories.items():
+                X[column] = np.where(X[column].isin(col_values) | X[column].isna(),
+                                     X[column], self.fill_value)
+        elif self.unseen == "ignore":
+            for column, col_values in self._categories.items():
+                X[column] = np.where(X[column].isin(col_values) | X[column].isna(),
+                                     X[column], np.nan)
```
solegalli, line 353: I am not sure this logic works as intended. First we replace the unseen with either a number or nan. If it is nan, then the transform step below should fail, because sklearn decision trees do not take nan as input. And if it replaces nan with a number, then the tree is going to use that number as input and return a prediction at the back of that. 

Somehow, we should first obtain the predictions over the permitted values and then replaced unseen by either nan or the fill_value.

solegalli (commented):
Hey @93lorenzo 

Thank you so much for the changes! I think it goes in the right direction.

Most of my comments are cosmetic, to try and keep the API homogeneous. But at the end I have one regarding the logic, which as it stands, it should not return what we want it to return.

Could you please take a look?

Thanks a lot!

feature_engine/encoding/decision_tree.py
93lorenzo: Hey @solegalli ! 
I implemented the changes you suggested. 
About the np.nan or fill_value replacement I used a workaround in the transform step:
* first I get a mask of those unknown
* I convert them to a known value to make succeed the encoding 
* in the end I replace the values in the mask with the fill_value or np.nan

Another point I would like to ask is that I am defining the `unseen` parameter of the OrdinalEncoder as `self.unseen`, but with the current developments I think I should keep it to `"raise"` because in any case it will receive seen categories. 
Here the code I am referring to. 
```
cat_encoder = OrdinalEncoder(
            encoding_method=self.encoding_method,
            variables=variables_,
            missing_values="raise",
            ignore_format=self.ignore_format,
            unseen="raise", # unseen=self.unseen,
        )
```

solegalli:
Hey @93lorenzo Sorry for the delayed response.

I made a PR to your repo with some changes: https://github.com/93lorenzo/feature_engine/pull/1

It turns out, that this is harder than I thought.

At the moment, this is what I think could work:

if encode=raise, we just let the ordinal encoder raise the error.
if encode is not raise, we let the encoder do whatever (i set it to encode so it will return a value for unseen categories and then the tree will make a prediction at the back of that, but we don;t care)

In transform, when encode is not raise, we need to check if there is a category in the dataframe, that was not present in the training set (during fit).

We have the unique categories for each variable in the encoding_dict_ parameter of the ordinalencoder.

What we need to do is to find a fast way to find categories in the new dataframe that are not among the keys of the corresponding variable's dictionary. We could use set(a).intersection(b) where a is the variables in the encoding_dict_ and b is the unique categories seen in the variables in the new df. And if there are unseen, then we create a boolean vector with True where the observation is unseen.

We do this for each variable, so we will end up with an array of Trues and Falses with the same size of the dataframe and the variables that are being encoded. And then using pandas loc we set the Trues to fill_value.

I don't think pandas.isin and apply are the fastest options.

@glevv would you know of a better way? and by better I mean faster.

solegalli:
@93lorenzo  Would you be able to pick up from where I left and try to modify the logic in transform to create the mask array and then replace by fill_value?

solegalli:
Looking at it again, the changes needed are instead of calling the list of unique values that you created in fit, we'd use the one that we already have within OrdinalEncoder. And instead of using apply and isin, we'd try to use sets and difference and numpy

solegalli:
Change of plans. We are tackling issue #588 together with this one, which makes our lives much easier.

solegalli:
superseeded by #757

PR #734, opened 2024-04-02
expand decision tree discretizer functionality

### LINKED ISSUE #452: DecisionTreeDiscretiser to output integers in addition to the predictions
At the moment, the DecisionTreeDiscretiser returns the values of the tree predictions as the replacement of the original variables.

I would like to add the option to return integers from 1 to k (or 0 to k-1), where k is the number of final leaves. The numbers increase with the mean value target per leave.

Not sure how difficult it is to implement, I think we sort of need to navigate the tree somehow, pick up the final values at the leaves, and then create a mapping from final value to the integer. And add a parameter in the init where the user can specify if they want integers or predictions as output.

The addition is suggested after the Self-Guided via CART method available in MINITAB and described [here](https://www.minitab.com/content/dam/www/en/uploadedfiles/content/products/spm/IntroDataBinning.pdf).

### LINKED ISSUE #588: replace attribute encoding pipeline for encoding dictionary

solegalli:
closes #588 
closes #452


feature_engine/discretisation/decision_tree.py
```diff
@@ -189,13 +190,19 @@ def __init__(
...
+        if bin_output == "boundaries" and precision is None:
+            raise ValueError(
+                "When `bin_output == 'boundaries', `precision` cannot be None. "
+                "Change precision's value to a positive integer."
+            )
         if not isinstance(regression, bool):
-            raise ValueError("regression can only take True or False. "
-                             f"Got {regression} instead.")
+            raise ValueError(
+                "regression can only take True or False. " f"Got {regression} instead."
```
VascoSch92:
```suggestion
                f"regression can only take True or False. Got {regression} instead."
```

solegalli:
```suggestion
                f"regression can only take True or False. Got {regression} instead."
```

solegalli: Thanks

PR #748, opened 2024-04-26
Fix pandas qcut link for the EqualFrequencyDiscretiser user guide

cmcouto-silva:
The current text (Pandas cut) links to the Pandas cut library. I am just replacing cut with qcut since that's what EqualFrequencyDiscretiser uses under the hood.


solegalli:
Thank you!

PR #755, opened 2024-05-02
add support for 'auto' value in 'fold' parameter

### LINKED ISSUE #754: allow outlier trimmer and winsorizer to select optimal fold automatically based on the method to detect outliers
At the moment, the parameter fold from OutlierTrimmer and Winsorizer are set to 3. 

I'd like it to be set to "auto" which would have this behaviour:

- if iqr: then fold = 1.5
- if gaussian, then fold=3
- if MAD, then fold=3.29 (coincides with r package: https://easystats.github.io/performance/articles/check_outliers
- if quantile: then 0.05

Easy fix, plus tests.

glevv:
closes #754


feature_engine/outliers/base_outlier.py
```diff
@@ -149,23 +149,29 @@ def __init__(
...
             )
 
         if tail not in ["right", "left", "both"]:
-            raise ValueError("tail takes only values 'right', 'left' or 'both'")
-
-        if fold <= 0:
-            raise ValueError("fold takes only positive numbers")
-
-        if capping_method == "quantiles" and fold > 0.2 and fold != 3:
+            raise ValueError("tail takes only values 'right', 'left' or 'both'.")
```
VascoSch92, line 163:
```suggestion
            raise ValueError(f"Parameter ``tail`` takes only values 'right', 'left' or 'both', but got {tail}")
```
Or perhaps use also a Literal for the parameter `tail`.

feature_engine/outliers/base_outlier.py
```diff
@@ -149,23 +149,29 @@ def __init__(
...
+
+        if (isinstance(fold, str) and (fold != "auto")) or (
```
VascoSch92, line 165: `(isinstance(fold, str) and (fold != "auto"))` this can never happen right? Because fold can be just "auto" if it is a string

feature_engine/outliers/base_outlier.py
```diff
@@ -149,23 +149,29 @@ def __init__(
...
+            isinstance(fold, (int, float)) and (fold <= 0)
+        ):
+            raise ValueError("fold takes only positive numbers or 'auto'.")
```
VascoSch92, line 168:
```suggestion
            raise ValueError(f"The parameter ``fold`` takes only positive numbers or 'auto', but go {fold}")
```

feature_engine/_docstrings/init_parameters/outliers.py
```diff
@@ -14,9 +14,14 @@
         distribution. Can take 'left', 'right' or 'both'.
     """.rstrip()
 
-_fold_docstring = """fold: int or float, default=0.05 if `quantile`, or 3 otherwise.
+_fold_docstring = """fold: int or float, default='auto'.
         The factor used to multiply the std, MAD or IQR to calculate
         the maximum or minimum allowed values.
+        When 'auto' is set, `'fold'` value will be chosen based on `capping_method`: \n
```
solegalli, line 20:
```suggestion
        When 'auto', `'fold'` is set based on the `capping_method`: \n
```

feature_engine/outliers/base_outlier.py
solegalli: I guess the user can enter a typo.

feature_engine/outliers/base_outlier.py
```diff
@@ -149,34 +149,48 @@ def __init__(
...
+                f"tail must be 'right', 'left' or 'both'. Got {tail} instead."
+            )
 
-        if fold <= 0:
-            raise ValueError("fold takes only positive numbers")
+        if (isinstance(fold, str) and (fold != "auto")) or (
+            isinstance(fold, (int, float)) and (fold <= 0)
+        ):
+            raise ValueError(
+                f"fold must be a positive numbers or 'auto'. Got {fold} instead."
```
solegalli, line 172:
```suggestion
                f"fold must be a positive number or 'auto'. Got {fold} instead."
```

feature_engine/outliers/base_outlier.py
```diff
@@ -210,15 +224,20 @@ def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
         self.right_tail_caps_ = {}
         self.left_tail_caps_ = {}
 
+        if self.fold == "auto":
+            self.fold_ = self._calculate_auto_fold()
```
solegalli, line 228: We need to add `fold_` to the attributes list in the docstrings. Could you do that?

feature_engine/outliers/winsorizer.py
```diff
@@ -1,7 +1,7 @@
 # Authors: Soledad Galli <solegalli@protonmail.com>
 # License: BSD 3 clause
 
-from typing import List, Union
+from typing import List, Literal, Union
```
solegalli, line 4: I think we may need this change also in the OutlierTrimmer? Did you check?

solegalli: And we probably need to add `fold_` in both docstrings, here and outliertrimmer, so might be worth adding it as a common string in the _docstring module.

solegalli:
Hi @glevv Thank you so much for the super quick turnaround!

Could I ask you an additional change? Could you update the corresponding rst files within the user_guide in the docs file? It'd be important to reflect the current limits, and if you can give a shoutout at the auto parameter it would be great :)

Thank you!

feature_engine/outliers/winsorizer.py
```diff
@@ -87,6 +87,9 @@ class Winsorizer(WinsorizerBase):
 
     {n_features_in_}
 
+    fold_:
+        The multiplier calculated during fit stage and used in transformation stage.
```
solegalli, line 91:
```suggestion
        Factor multiplying the std, mad, iqr or alternative the percentile. Only different 
        from `fold` when `fold="auto".
```

solegalli (commented):
Hey @glevv Thank you so much! If we change the wording of fold_ we are good to merge!

solegalli:
Thanks a lot!

PR #757, opened 2024-05-06
refactor tree encoder: add encoding dict, param unseen, inverse transform

### LINKED ISSUE #588: replace attribute encoding pipeline for encoding dictionary

### LINKED ISSUE #728: Parametrize unseen for the OrdinalEncoder in DecisionTreeEncoder

solegalli:
closes #729 
closes #728 
closes #588


solegalli:
Hey @VascoSch92 @glevv @93lorenzo

If you have a minute to spare, a quick look to see that there is nothing big that I am missing will be much appreciated :)

tests/test_encoding/test_decision_tree_encoder.py
```diff
@@ -2,24 +2,97 @@
...
-def test_encoding_method_param(df_enc):
-    # defaults
+# init parameters
+@pytest.mark.parametrize("enc_method", ["count", False, 1])
+def test_error_if_encoding_method_not_permitted_value(enc_method):
+    msg = (
+        "`encoding_method` takes only values 'ordered' and 'arbitrary'."
+        f" Got {enc_method} instead."
+    )
+    with pytest.raises(ValueError) as record:
```
glevv, line 17: pytest.raises(ValueError, match=msg) is way cleaner

tests/test_encoding/test_decision_tree_encoder.py
```diff
@@ -2,24 +2,97 @@
...
+@pytest.mark.parametrize("precision", ["string", 0.1, -1, np.nan])
+def test_error_if_precision_gets_not_permitted_value(precision):
+    msg = "Parameter `precision` takes integers or None. " f"Got {precision} instead."
+    with pytest.raises(ValueError) as record:
+        DecisionTreeEncoder(precision=precision)
+    str(record.value) == msg
+
+
+@pytest.mark.parametrize(
+    "params",
```
glevv, line 54: Another IMO more readable way would be:

"encoding_method,ignore_format,precision,unseen,fill_value" instead of "params"
and then 
def test_init_param_assignment(encoding_method, ignore_format, precision, unseen, fill_value):
    DecisionTreeEncoder(
        encoding_method=encoding_method,
        ignore_format=ignore_format,
        precision=precision,
        unseen=unseen,
        fill_value=fill_value,
    )

feature_engine/encoding/decision_tree.py
```diff
@@ -261,16 +308,27 @@ def fit(self, X: pd.DataFrame, y: pd.Series):
...
             ]
         )
 
-        encoder_.fit(X, y)
+        Xt = pipe.fit_transform(X, y)
+
+        encoder_ = {}
+        if self.precision is None:
+            for var in variables_:
+                encoder_[var] = dict(zip(X[var], Xt[var]))
```
VascoSch92, line 323: Just to be sure: the variable `variables_` doesn't contain duplicated values right?

Otherwise, this line of code could create some bugs. ;-)

93lorenzo: Great question, but It should not contain duplicates. 
`_variables` is derived from the [select_dtypes](https://github.com/feature-engine/feature_engine/blob/46155a0b51def2bf96c1c4719a3dd9bfd11f78d3/feature_engine/variable_handling/find_variables.py) function

VascoSch92: Thanks for the answer. 

It was just to be sure ;-)

tests/test_encoding/test_decision_tree_encoder.py
solegalli: thanks!

tests/test_encoding/test_decision_tree_encoder.py
solegalli: Done, thank you

solegalli:
Thank you guys! @93lorenzo @glevv @VascoSch92 

Really appreciate it :)

93lorenzo (approved):

tests/test_encoding/test_decision_tree_encoder.py
```diff
@@ -2,24 +2,96 @@
...
+
+@pytest.mark.parametrize(
+    "unseen", ["string", False, ("raise", "ignore"), ["ignore"], np.nan]
+)
+def test_error_if_unseen_gets_not_permitted_value(unseen):
+    msg = (
+        "Parameter `unseen` takes only values ignore, raise, encode. "
+        f"Got {unseen} instead."
+    )
+    with pytest.raises(ValueError) as record:
```
glevv, line 29: Miss one

tests/test_encoding/test_decision_tree_encoder.py
```diff
@@ -163,3 +222,119 @@ def test_assigns_param_grid(grid):
...
+            "var_B": ["C", "YYY", "ZZZ"],
+        }
+    )
+    var_ls = "var_A, var_B"
+    msg = (
+        "During the encoding, NaN values were introduced in the feature(s) "
+        f"{var_ls}."
+    )
+    # new categories will raise an error
+    with pytest.raises(ValueError) as record:
```
glevv, line 286: Also look at this one

solegalli: Yep, I didn't miss them. I changed them to match=msg, but then these 2 returned an error which I didn't understand and I was lazy to resolve. Although the messages were identical, the test failed, and it said something about did you want to **escape** something? Did you come across this before? Probably something to do with regex. which is not my forte :/

glevv: both lines of the msg should be f-strings in both cases

solegalli: Hmm, not sure I follow: This is the error

```
E       AssertionError: Regex pattern did not match.
E        Regex: 'During the encoding, NaN values were introduced in the feature(s) var_A, var_B.'
E        Input: 'During the encoding, NaN values were introduced in the feature(s) var_A, var_B.'
E        Did you mean to `re.escape()` the regex?
```

And this is the code:

```
def test_fit_errors_if_new_cat_values_and_unseen_is_raise_param(df_enc):
    encoder = DecisionTreeEncoder(unseen="raise", regression=False)
    encoder.fit(df_enc[["var_A", "var_B"]], df_enc["target"])
    X = pd.DataFrame(
        {
            "var_A": ["A", "ZZZ", "YYY"],
            "var_B": ["C", "YYY", "ZZZ"],
        }
    )
    var_ls = "var_A, var_B"
    msg = (
        f"During the encoding, NaN values were introduced in the feature(s) {var_ls}."
    )
    # new categories will raise an error
    with pytest.raises(ValueError, match=msg):
        encoder.transform(X)
    # assert str(record.value) == msg
```

What am I doing wrong?

glevv: Oh, I see. Match is a regex patter search, so it gets confused by `(s)` put an escape characters like this

`f"During the encoding, NaN values were introduced in the feature\(s\) {var_ls}."`

93lorenzo: I think it is the parentesis "()".
I assume the match of the msg uses some regex, and having "()" within the string trigger some regex patterns and throws the exception.

glevv: Updated the comment

tests/test_encoding/test_decision_tree_encoder.py
glevv: This one should also be fixed

solegalli: Is this comment new? I don't think I left any record in the file I committed last. Just doubled checked and didnt find any left.

The error comes from the values within parametrize, when I pass a list or a tuple, say ["raise", "ignore"] and then that tuple is passed as {unseen}, it doesn't like it. But it resolved with re.escape(message here)

tests/test_encoding/test_decision_tree_encoder.py
```diff
@@ -1,25 +1,98 @@
...
+    )
+    with pytest.raises(ValueError, match=msg):
+        DecisionTreeEncoder(encoding_method=enc_method)
+
+
+@pytest.mark.parametrize(
+    "unseen", ["string", False, ("raise", "ignore"), ["ignore"], np.nan]
+)
+def test_error_if_unseen_gets_not_permitted_value(unseen):
+    msg = re.escape(
```
solegalli, line 27: @glevv fixed here

PR #766, opened 2024-05-19
add transformer that combines features with decision trees

### LINKED ISSUE #107: create new features by combining variables with decision trees

solegalli:
closes #107 
superseeds #454 

- [x] add example to docstring
- [x] add user guide
- [x] add transformer to readme and index


feature_engine/creation/decision_tree_features.py
```diff
@@ -0,0 +1,461 @@
...
+        param_grid: Optional[Dict[str, Union[str, int, float, List[int]]]] = None,
+        regression: bool = True,
+        random_state: int = 0,
+        missing_values: str = "raise",
+        drop_original: bool = False,
+    ) -> None:
+
+        if precision is not None and (not isinstance(precision, int) or precision < 1):
+            raise ValueError(
+                "precision must be None or a positive integer. "
```
VascoSch92, line 205:
```suggestion
                "precision must be None or a non-zero positive integer. "
```

PR #772, opened 2024-06-17
Numpy2 0 0 updates np.Inf vs np.inf

olikra:
Update tests for compatibility of numpy2_0_0 regarding np.Inf vs. np.inf


olikra:
@solegalli I assume you have to merge these changes into base branch to retest the other open pulls...

solegalli:
Thank you @olikra ! It's the never-ending story with the updates from numpy, pandas and sklearn.

PR #774, opened 2024-06-24
Math features migrate to numpy

olikra:
I migrated the transform method in math_features from pandas.agg to native numpy functions with backward compatibility. At least all existing tests on math-feature pass.

Following key-functions are implemented:

- "sum", "np.sum"
- "mean", "np.mean"
- "min", "np.min"
- "max", "np.max"
- "prod", "np.prod"
- "median", "np.median"
- "std", "np.std"
- "var", "np.var"

In case a defined function is not in above collection it falls back to pandas.agg. In the Examples-section I described the way to call a custom function with numpy.apply_over_axes()

I additionally changed the deprecated 

`dob_datrange = pd.date_range("2020-02-24", periods=4, freq="T")`

to

`dob_datrange = pd.date_range("2020-02-24", periods=4, freq="min")`

to avoid FutureWarnings from pandas.

Please have a look. @solegalli

### DIFF
--- a/feature_engine/creation/custom_functions.py
+++ b/feature_engine/creation/custom_functions.py
@@ -6,6 +6,5 @@ custom function should be processed via pandas.agg() or numpy.apply_over_axes()
 
 class CustomFunctions:
 
-
     def __init__(self, scope_target):
         self.scope_target = scope_target
--- a/feature_engine/creation/custom_functions.py
+++ b/feature_engine/creation/custom_functions.py
@@ -1,0 +1,11 @@
+"""
+CustomFunctions is a wrapper, which allows MathFunctions to detect if a
+custom function should be processed via pandas.agg() or numpy.apply_over_axes()
+"""
+
+
+class CustomFunctions:
+
+
+    def __init__(self, scope_target):
+        self.scope_target = scope_target
--- a/feature_engine/creation/math_features.py
+++ b/feature_engine/creation/math_features.py
@@ -378,7 +378,5 @@ class MathFeatures(BaseCreation):
                 feature_names = [
                     f"{function}_{'_'.join(varlist)}" for function in functions
                 ]
-            else:
-                feature_names = [f"{self.func}_{'_'.join(varlist)}"]
 
         return feature_names
--- a/feature_engine/creation/math_features.py
+++ b/feature_engine/creation/math_features.py
@@ -343,8 +343,6 @@ class MathFeatures(BaseCreation):
                         result = np_df[np_variables].agg(np_function, axis=1)
                         np_result_df[new_variable_names[np_function_idx]] = result
 
-                    print(scope_target)
-
             return np_result_df
 
         X = self._check_transform_input_and_state(X)
--- a/feature_engine/creation/math_features.py
+++ b/feature_engine/creation/math_features.py
@@ -1,7 +1,7 @@
 from typing import Any, List, Optional, Union
 
+import numpy as np
 import pandas as pd
-
 from feature_engine._docstrings.fit_attributes import (
     _feature_names_in_docstring,
     _n_features_in_docstring,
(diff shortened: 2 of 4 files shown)


feature_engine/creation/custom_functions.py
```diff
@@ -0,0 +1,10 @@
+"""
+CustomFunctions is a wrapper, which allows MathFunctions to detect if a
+custom function should be processed via pandas.agg() or numpy.apply_over_axes()
+"""
+
+
+class CustomFunctions:
```
solegalli, line 7: Do we need this class? it seems to be implementing a very simple functionality. Can this not be a one liner in the main MathFeatures class? 

Besides that, the class name does not immediately tell what the class is doing, we'd need docstrings and typehints.

feature_engine/creation/math_features.py
```diff
@@ -68,6 +69,11 @@ class MathFeatures(BaseCreation):
         one name per function. If None, the transformer will assign arbitrary names,
         starting with the function and followed by the variables separated by _.
 
+    ddof: int, float, default = 1
```
solegalli, line 72: can ddof ever be a float? the ddof are usually integers, no?

feature_engine/creation/math_features.py
```diff
@@ -130,6 +136,54 @@ class MathFeatures(BaseCreation):
     0   1   4         2.5
     1   2   5         3.5
     2   3   6         4.5
+
```
solegalli, line 139: All of this, should go in the user guide, not here. It is too much information for the class docstrings. But let's hold on with the change until we finalize the class update

feature_engine/creation/math_features.py
```diff
@@ -139,8 +193,19 @@ def __init__(
         new_variables_names: Optional[List[str]] = None,
         missing_values: str = "raise",
         drop_original: bool = False,
+        ddof: Union[int, float] = 1,
```
solegalli, line 196: I think ddof should be just int

feature_engine/creation/math_features.py
```diff
@@ -139,8 +193,19 @@ def __init__(
...
     ) -> None:
 
+        # casting input parameter func to a list
```
solegalli, line 199: for compatibility with sklearn, we can't modify the parameters that the user enters at init. They need to remain the same. Anything that needs changing, needs to happen in the fit and be added as a new attribute followed by _ if necessary at all.

feature_engine/creation/math_features.py
```diff
@@ -175,18 +240,16 @@ def __init__(
...
                         "of functions."
                     )
-            else:
-                if len(new_variables_names) != 1:
-                    raise ValueError(
-                        "The number of new feature names must coincide with the number "
-                        "of functions."
-                    )
+
+        if not (isinstance(ddof, int) or isinstance(ddof, float)):
```
solegalli, line 244: if not insinstance(ddof, (int, float)): 
 assuming that we allow it to take both values.

feature_engine/creation/math_features.py
```diff
@@ -175,18 +240,16 @@ def __init__(
...
+            raise ValueError("The type of ddof has to be Integer or Float")
```
solegalli, line 245: error message needs to contain "Got {ddof} instead." to help user undestand why the error

feature_engine/creation/math_features.py
```diff
@@ -202,14 +265,117 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
         X_new: Pandas dataframe, shape = [n_samples, n_features + n_operations]
             The input dataframe plus the new variables.
         """
+
+        def np_transform(np_df, new_variable_names, np_variables, np_functions):
```
solegalli, line 269: we need typehint and better function names. What is np_transform supposed to do? not immediately obvious from the name. I'd also not define a method within a method. Let's define it outside preceded by _

feature_engine/creation/math_features.py
```diff
@@ -202,14 +265,117 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+            np_result_df = pd.DataFrame()
+            for np_function_idx, np_function in enumerate(np_functions):
+                if callable(np_function):
```
solegalli, line 272: if it is a callable, can we not apply it straightaway? is not getting the name and then the if loop adding complexity?

feature_engine/creation/math_features.py
```diff
@@ -202,14 +265,117 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+                    np_function_name = np_function.__name__
+                else:
+                    np_function_name = np_function
+
+                if np_function_name in ("sum"):
```
solegalli, line 277: I think asserting equality is faster than scanning something in set, is that the case? did you check?

feature_engine/creation/math_features.py
```diff
@@ -202,14 +265,117 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+                    result = np.nansum(
+                        np_df[np_variables],
+                        axis=1,
+                    )
+                    np_result_df[new_variable_names[np_function_idx]] = pd.Series(
+                        result
+                    )
+                    continue
```
solegalli, line 285: why do we need continue here? so that it reads the following elif? should we not just use if then?

tests/test_creation/test_math_features.py
```diff
@@ -4,6 +4,7 @@
 from sklearn.pipeline import Pipeline
 
 from feature_engine.creation import MathFeatures
+from feature_engine.creation.custom_functions import CustomFunctions
```
solegalli, line 7: if we do keep this function, we need to test it in a separate test file. We have 1 test file per class/script

tests/test_creation/test_math_features.py
```diff
@@ -39,6 +40,11 @@ def test_error_if_func_is_dictionary():
         MathFeatures(variables=["Age", "Name"], func={"A": "sum", "B": "mean"})
 
 
+def test_error_if_ddof_is_not_int_or_float():
+    with pytest.raises(ValueError):
+        MathFeatures(variables=["Age", "Name"], func={"std", "var"}, ddof="A")
```
solegalli, line 45: we also want to assert the error message is the expected one

solegalli (commented):
Hey @olikra 

Thank you so much for moving forward with these changes. And apologies for the delayed response. I was first sick and then on a short holiday break.

I added some comments, I am not convinced we need the CustomFuntions class, and I wonder if there is a better way of implementing the functionality other than so many elifs.

Could we not do something like what we have in relative features: https://github.com/feature-engine/feature_engine/blob/8cfab264e679c09fe85fc49cbc25c053b80b4fd2/feature_engine/creation/relative_features.py#L199

Would be great if you could look at the comments and let me know your thoughts. Thank you!

PR #777, opened 2024-06-25
Generell code coverage increasing

olikra:
Added some missing tests on several classes - mainly on the init-part for classes
@solegalli please have a look

PR #781, opened 2024-07-11
fix deprecations improve code coverage

solegalli:
closes #777


solegalli:
Alternatively I can merge this PR which contains your changes @olikra maybe that's easier?

PR #788, opened 2024-07-22
update user guide about ReciprocalTransformer

sergiobemar:
This PR has the goal to update the content of the Reciprocal Transfomer documentation. The new content includes:
- Explanation about the theoretical of this math function.
- Pros and cons of its using
- Code example in order to ease the reproducitibility with images of the obtained plots


solegalli:
Exito! fyi @sergiobemar

Issue #790, opened 2024-08-14
Add a Binarizer class 


d-lazenby:
**A clear and concise description of what the problem is.** 
I needed to binarize some features in a pipeline but saw that this wasn't currently implemented in the package; I instead used the sklearn Binarizer and ColumnTransformer classes. This wasn't ideal as I had to keep track of the new indexes at subsequent steps of the pipeline - one of the nice things about feature engine is that you don't have to worry about that. 

**Describe the solution you'd like**
An implementation of a Binarizer class that bypasses the need for the corresponding sklearn class and that you can use directly within the Pipeline class instead of using ColumnTransformer.

**Describe alternatives you've considered**
I wrote a class that inherits from the BaseDiscretizer which works as described above. 

Would it be OK for me to work on this and open a PR?

solegalli:
Hey @d-lazenby ! Thanks for the suggestion! You are more than welcome to create a PR! I'll stay on the look out!

PR #791, opened 2024-08-16
Feat/binarizer without column transformer

d-lazenby:
Issue raised [here](https://github.com/feature-engine/feature_engine/issues/790#issue-2466280271)

**Notes on Code**
The `BinaryDiscretiser` class is implemented in `binariser.py`, located with the other discretisers, and takes a parameter threshold to determine where to split the interval. 
- After standard checks and type checks for threshold, there's a check to see if the threshold is in min(x) < threshold < max(x) for each feature x (L167). If not, then x isn't transformed and the user is notified of this. The remaining features are passed to a list for transformation.
- Because of the above, the transform method from the `BaseDiscretiser` is repeated here, only iterating through the new list of features that passed the threshold check rather than the list in `self.variables_`. I'm not sure if there's a cleaner way of doing this. We could also modify the `self.variables_` attribute directly in the fit method instead, which might make sense since then it would contain only features that were actually transformed.

**Other notes**
- I updated the docs apart from the user_guide since this might change depending on further changes to the implementation
- I've tested on an sklearn Pipeline and it seems to work fine but haven't included explicit tests for that as they were missing for the other discretisers. Let me know if that's something you'd want. 
- It might be nice to have functionality where the user can pass a set of different thresholds for each feature passed to the class (could be corresponding lists for threshold and variables parameters, or a dictionary of pairs).
- The threshold check output is written to stdout at the moment, but this should perhaps be given as a warning instead. 

**Finally**
This is my first time contributing to open source – all feedback is very welcome!


d-lazenby:
Hi @solegalli, I forgot to write in the PR, could someone review this please? 

Some of the style checks don't pass. I noticed that when I run isort on `binariser.py` some changes are made to the imports section but then running black on the file changes them back, so that might be the cause of the issue.

feature_engine/discretisation/binariser.py
```diff
@@ -0,0 +1,231 @@
...
+                failed_threshold_check.append(var)
+            else:
+                self.binner_dict_[var] = [
+                    float("-inf"),
+                    np.float64(self.threshold),
+                    float("inf"),
+                ]
+
+        if failed_threshold_check:
+            print(
```
VascoSch92, line 178: this should be a logger not a print I think

feature_engine/discretisation/binariser.py
```diff
@@ -0,0 +1,231 @@
...
+
+    The BinaryDiscretiser() first finds the boundaries for the intervals for
+    each variable. Then, it transforms the variables, that is, sorts the values into
+    the intervals.
+
+    Parameters
+    ----------
+    {variables}
+
+    threshold: int, float, default=None
```
solegalli, line 63: We need to default threshold to a value that can actually be used for splitting. Sklearn binarizer defaults to 0, so we could take the same approach and make the threshold default to zero. Could you pls change that?

feature_engine/discretisation/binariser.py
```diff
@@ -0,0 +1,231 @@
...
+    binner_dict_=_binner_dict_docstring,
+    fit=_fit_discretiser_docstring,
+    transform=_transform_discretiser_docstring,
+    variables=_variables_numerical_docstring,
+    variables_=_variables_attribute_docstring,
+    feature_names_in_=_feature_names_in_docstring,
+    n_features_in_=_n_features_in_docstring,
+    fit_transform=_fit_transform_docstring,
+)
+class BinaryDiscretiser(BaseDiscretiser):
```
solegalli, line 45: I wonder whether we should use Binarizer as name, like sklearn's class. Thoughts?

feature_engine/discretisation/binariser.py
```diff
@@ -0,0 +1,231 @@
...
+    Parameters
+    ----------
+    {variables}
+
+    threshold: int, float, default=None
+        Desired value at which to divide the interval.
+
+    {return_object}
+
+    {return_boundaries}
```
solegalli, line 68: Do we need to return boundaries? It's just 2 intervals, either it is greater or smaller than threshold. A priori, I think we should drop this paramter.

feature_engine/discretisation/binariser.py
```diff
@@ -0,0 +1,231 @@
...
+
+    {precision}
```
solegalli, line 70: I also think we don't need precision, because we don't have interval boundaries. The boundaries are -inf to threshold and then threshold to inf. We can drop this one as well.

feature_engine/discretisation/binariser.py
```diff
@@ -0,0 +1,231 @@
...
+
+    Attributes
+    ----------
+    {binner_dict_}
```
solegalli, line 74: I am not sure we need a binner dict for this class. We don't have a sequence of boundaries that vary variable per variable.

feature_engine/discretisation/binariser.py
```diff
@@ -0,0 +1,231 @@
...
+    {fit_transform}
+
+    {transform}
+
+    See Also
+    --------
+    pandas.cut
+    sklearn.preprocessing.KBinsDiscretizer
+
+    References
```
solegalli, line 95: we need to drop these references.

feature_engine/discretisation/binariser.py
```diff
@@ -0,0 +1,231 @@
...
+    def __init__(
+        self,
+        threshold: Union[None, int, float] = None,
+        variables: Union[None, int, str, List[Union[str, int]]] = None,
+        return_object: bool = False,
+        return_boundaries: bool = False,
+        precision: int = 3,
+    ) -> None:
+
+        if threshold is None:
```
solegalli, line 129: We can't allow the default value to be a non permitted value. Default to 0 instead.

feature_engine/discretisation/binariser.py
```diff
@@ -0,0 +1,231 @@
...
+
+        failed_threshold_check = []
+        self.binner_dict_ = {}
+        for var in self.variables_:
+            # Check that threshold is within range
+            if (self.threshold < min(X[var])) or (self.threshold > max(X[var])):
+                # Omit these features from transformation step
+                failed_threshold_check.append(var)
+            else:
+                self.binner_dict_[var] = [
```
solegalli, line 171: the binner dict will be the same for all variables. I am not sure it is useful for the user. We should not create this param

feature_engine/discretisation/binariser.py
solegalli: This should be a warning instead of print. Could we change?

feature_engine/discretisation/binariser.py
```diff
@@ -0,0 +1,231 @@
...
+                ]
+
+        if failed_threshold_check:
+            print(
+                "threshold outside of range for one or more variables."
+                f" Features {failed_threshold_check} have not been transformed."
+            )
+
+        # A list of features that satisfy threshold check and will be transformed
+        self.variables_trans_ = [
```
solegalli, line 184: we normally store the variables that will be transformed in variables_ But I am not convinced that we need to exclude the variables that will not be transformed. This is something for the user to take care of, not feature-engine.

We can raise a warning, but that is as far as I would go.

solegalli (commented):
Hey @d-lazenby 

Thank you so much for the PR and sorry for the delayed response. You put a lot of work already well done and thanks!

This class at the moment works like the Binarizer from sklearn, it takes a threshold and then splits the variables in 2 intervals around the threshold. So far so good. The only additional thing that I could consider is whether we allow the user to pass a dictionary with different thresholds per variable for example 0 for vars a and b, and 2 for vars c and d. (See arbitrary discretiser for params at init).

Since this class returns only 2 intervals, I am inclined to suggest that we only allow the output of integers, 0 if it is smaller than threshold and 1 otherwise. I don't see that having boundaries from -inf to threshold and then threshold to inf could be useful for the user. And that would simplify the class a lot.

What do you think? Should we change that? 

I left a few comments across the files also. Would you be able to take a look?

Thanks a lot!

d-lazenby:
Hi @solegalli, thanks a lot for looking at this, all your suggestions look
good. I’m out of action for the next couple of weeks but will be in touch
with the code amendments when I’m back. All the best

solegalli:
Sounds good!

feature_engine/discretisation/binariser.py
d-lazenby: Hi @solegalli, I’m back working on this now and wanted to check something before proceeding.

The functionality that the user passes a dictionary where the features are the keys and the thresholds are the values (float or int) is good. I was wondering if we wanted just this, or this with an additional possibility of passing a list to `variables` that all default to zero threshold. My feeling is the latter might be a bit unnecessary, so I'd be inclined to insist on a user-inputted dictionary (as with the ArbitraryDiscretiser), remove `variables` as a user input and circumvent the need for default values for the thresholds but let me know if you disagree. If we take just the user-inputted dict route, we could set `self.variables_` to be the keys of the threshold dictionary so the user can easily access them in a way that’s consistent with the other transformers.

feature_engine/discretisation/binariser.py
d-lazenby: I started with Binarizer, funnily enough :) 

I changed it to be consistent with all the other discretiser class names in feature engine (EqualFrequencyDiscretiser, EqualWidthDiscretiser, etc.) since it looked strange to me to have a single -ize spelling in there and omit the 'Discretiser' part but I'm happy to go with your preference.

PR #797, opened 2024-08-24
expand cyclical encoder documentation to discuss max values

### LINKED ISSUE #765: cyclical features may not always get the right period
CyclicalFeatures obtains the period by looking for the maximum value. So if months are coded from 1-12, the period will be 12 and that'll be fine.

For hours for example, if the hrs are coded from 1-24, it will get the period as 24 and that will also be fine, but if hrs are coded as 0-23, then the period will be set to 23 automatically, but in reality it is still 24.

Maybe we could use n_unique instead of max value? 

In both cases there is still an issue to resolve: if the training set does not have the maximum value, or does not contain all elements, then the period will still be wrongly estimated, but I guess we could make that clear in the docs, I am not sure we can sort that out from our end.

solegalli:
closes #765 

FYI @VascoSch92 and @glevv

### DIFF
--- a/docs/user_guide/creation/CyclicalFeatures.rst
+++ b/docs/user_guide/creation/CyclicalFeatures.rst
@@ -25,7 +25,7 @@ circle. The idea was that, after preprocessing the cyclical data, the lowest val
 cyclical feature would appear right next to the largest value.
 
 To represent cyclical features in (x, y) coordinates, the authors created two new features,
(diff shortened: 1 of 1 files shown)


solegalli:
I decided not to change the logic and instead to expand the docs. If you still remember the discussion, let me know if this makes sense. Thank you!

Issue #722, opened 2024-02-07
RecursiveFeatureElimination Error using Ccross-validation Splitter


behzad89:
**Describe the bug**
When I try to use a cross-validation splitter instead of an integer value in `RecursiveFeatureElimination` I face an error that says `IndexError: list index out of range.` I used the the same splitter in the cross-validation function in sklearn and it worked properly. So I wonder why this happens? and if it has any solution or not?

**To Reproduce**
```
Steps to reproduce the behavior:
group_cvs =  LeaveOneGroupOut()
cvsp=group_cvs.split(X, y, groups)
model = RandomForestRegressor(random_state=123
sel.fit(X,y))
```
**Expected behavior**
---------------------------------------------------------------------------
```
IndexError                                Traceback (most recent call last)
Cell In[102], line 1
----> 1 sel.fit(X,y)

File ~/Documents/Learning/Air_Polltion_Map_Milan/myenv/lib/python3.9/site-packages/feature_engine/selection/recursive_feature_elimination.py:179, in RecursiveFeatureElimination.fit(self, X, y)
    176     break
    178 # remove feature and train new model
--> 179 model_tmp = cross_validate(
    180     self.estimator,
    181     X_tmp.drop(columns=feature),
    182     y,
    183     cv=self.cv,
    184     scoring=self.scoring,
    185     return_estimator=False,
    186 )
    188 # assign new model performance
    189 model_tmp_performance = model_tmp["test_score"].mean()

File ~/Documents/Learning/Air_Polltion_Map_Milan/myenv/lib/python3.9/site-packages/sklearn/utils/_param_validation.py:214, in validate_params.<locals>.decorator.<locals>.wrapper(*args, **kwargs)
    208 try:
    209     with config_context(
    210         skip_parameter_validation=(
    211             prefer_skip_nested_validation or global_skip_validation
    212         )
    213     ):
--> 214         return func(*args, **kwargs)
    215 except InvalidParameterError as e:
    216     # When the function is just a wrapper around an estimator, we allow
    217     # the function to delegate validation to the estimator, but we replace
    218     # the name of the estimator by the name of the function in the error
    219     # message to avoid confusion.
    220     msg = re.sub(
    221         r"parameter of \w+ must be",
    222         f"parameter of {func.__qualname__} must be",
    223         str(e),
    224     )

[...]

PR #799, opened 2024-08-24
fix generator input for rfe

Benetti-Hub:
Fixes https://github.com/feature-engine/feature_engine/issues/722

The issue arises because a generator object becomes exhausted after a single iteration. To address this, the generator has been replaced with a list, allowing multiple iterations and resolving the problem.


solegalli:
Hey @Benetti-Hub 

Thank you for the PR. 

On further investigation, this set up works without the fix:

```
from sklearn.linear_model import LinearRegression
from sklearn.model_selection import KFold
import pandas as pd
from feature_engine.selection import RecursiveFeatureElimination

linear_model = LinearRegression()

df = pd.DataFrame(
    {
        "x": [0, 0, 0, 0, 0, 1, 1, 1, 1, 1],
        "z": [1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
        "y": [0, 0, 0, 0, 0, 1, 1, 1, 1, 1],
    }
)
cv = KFold()

transformer = RecursiveFeatureElimination(
    estimator=linear_model, scoring="r2", cv=cv
)
transformer.fit(df[["x", "z"]], df["y"])
```

So in short, we should pass cv and not cv.split to the parameter.

This is how it should be set up based on sklearn's documentation.

tests/test_selection/test_recursive_feature_elimination.py
```diff
@@ -208,3 +209,21 @@ def test_stops_when_only_one_feature_remains():
...
+            "z": [1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
+            "y": [0, 0, 0, 0, 0, 1, 1, 1, 1, 1],
+        }
+    )
+    cv = KFold()
+
+    transformer = RecursiveFeatureElimination(
+        estimator=linear_model, scoring="r2", cv=cv.split(df)
+    )
+    transformer.fit(df[["x", "z"]], df["y"])
```
solegalli, line 229: this test is missing an assert or something to test. As it stands, it just passes because it does not fail.

solegalli (commented):
Hi @Benetti-Hub 

Thank you for the PR and for figuring out what's going on here.

I think the RecursiveFeatureAddition may have the same problem. Would you be able to expand the fix to that transformer as well?

Benetti-Hub:
Hi! I made a change in the `BaseRecursiveSelector` class, which also impacted `RecursiveFeatureAddition` (since it's derived from it). Nonetheless, to ensure everything works as expected, I added a test along with assertions to verify that the output matches the expected results.

Let me know if anything else needs to be changed

feature_engine/selection/base_recursive_selector.py
```diff
@@ -117,7 +118,7 @@ def __init__(
         self.estimator = estimator
         self.scoring = scoring
         self.threshold = threshold
-        self.cv = cv
+        self.cv = list(cv) if isinstance(cv, GeneratorType) else cv
```
solegalli, line 121: Hi @Benetti-Hub thank you so much! This is almost good to go. 

Unfortunately, to remain compatible with sklearn, we can't modify the parameters of init. They need to take the values that the user passes. Could you leave the parameter setting in init as is, an move this line of code to fit?

Benetti-Hub: Done! I have moved them

solegalli: Hey @Benetti-Hub 

We are still redefining the self.cv.

What I'd suggest is creating a temporary cv and then passing that onto the cross_validate. I am trying to do that but it does not resolve the issue, so it needs further digging.

The tests also need to check that the output of transform is the desired dataframe, to ensure that the class works as expected and returns the same results when the user passes cv and cv.split(X,y)

solegalli:
closed in #805

feature_engine/selection/base_recursive_selector.py
Benetti-Hub: Hey, given how generators work in Python, it's not possible to create a direct copy of one I am afraid. The issue arises because `self.cv` is used both during the initial fit (which exhausts the generator) and again in the cross-validation, leading to complications.

When I tested a similar scenario in scikit-learn, I observed that the `self.cv` attribute indeed becomes an exhausted generator after the fit call, which modifies it. In fact, you cannot fit two RFECV sequentially using the same generator.

Addressing this problem seems challenging without altering the `super().fit(X, y)` method to include a "cv" argument. Could you let me know if such a modification would be acceptable? If not, I am afraid that the only solution might be to raise a more descriptive error when a generator is passed.

Benetti-Hub: oh nvm, i checked the other PR, specifying an additionaly argument also works

solegalli: Hi @Benetti-Hub thank you so much for this message. 

I spend a good hour fighting with the generators and trying to get it to work, and I faced the same issues that you describe in your message. So I am happy to see that I am not the only one, because I was starting to think that I did something wrong and that there is something that I was not seeing through.

Your message gave me a lot of relief.

Anyways, your PR was key for me to get the solution, so I'll credit the work to you. Thank you so much. I'll make a PR during the week for a new release and tag you along the other contributors.

Thanks a lot!

Benetti-Hub: Glad I could help!

PR #804, opened 2024-08-27
Groups support in feature selection for cross validate

KananMahammadli:
Currently feature selection methods: `BaseRecursiveSelector`, `ProbeFeatureSelection`, `RecursiveFeatureAddition`, `RecursiveFeatureElimination`, `SelectByShuffling`, `SelectBySingleFeaturePerformance`, `SelectByTargetMeanPerformance` use `cross_validate` from `sklearn`, which support `groups` argument for group based splits, but feature_engine doesn't support it. Purpose of this PR is to enable feature_engine to have feature selection with cross_validation based on groups too. Example usage: While working with multiple time series data, for example, store item sales, same dates are repeated for multiple items, and during cross-validation it is good practice to use time series splits considering groups of items for each date. 

**Update:**

- Imports were put into original format, but that format doesn't align with isort, and running tox doesn't use isort
- Groups are given default None value for every class and function it is used, so current tests do not have to pass groups
- If everything is ok with current changes, I will write unit tests for groups usage


feature_engine/selection/base_selection_functions.py
```diff
@@ -1,17 +1,15 @@
...
 from sklearn.model_selection import cross_validate
 
-from feature_engine.variable_handling import (
-    check_all_variables,
-    check_numerical_variables,
-    find_all_variables,
-    find_numerical_variables,
-    retain_variables_if_in_df,
-)
+from feature_engine.variable_handling import (check_all_variables,
```
solegalli, line 8: could we put these changes back to the original representation? We use black for code styling.

feature_engine/selection/probe_feature_selection.py
```diff
@@ -4,34 +4,23 @@
 import pandas as pd
 
 from feature_engine._docstrings.fit_attributes import (
-    _feature_names_in_docstring,
-    _n_features_in_docstring,
-)
+    _feature_names_in_docstring, _n_features_in_docstring)
```
solegalli, line 7: could we put this and the below back to the original representation? we use black for styling.

feature_engine/selection/single_feature_performance.py
```diff
@@ -3,39 +3,26 @@
...
-)
+from feature_engine._check_init_parameters.check_variables import \
+    _check_variables_input_value
 from feature_engine._docstrings.fit_attributes import (
-    _feature_names_in_docstring,
-    _n_features_in_docstring,
-)
-from feature_engine._docstrings.init_parameters.selection import (
-    _confirm_variables_docstring,
-)
```
solegalli, line 15: could we put these imports back to the original representation?

tests/test_selection/test_base_selection_functions.py
```diff
@@ -238,10 +251,24 @@ def test_find_feature_importance(df_test):
...
     pd.testing.assert_series_equal(mean_.round(2), expected_mean)
     pd.testing.assert_series_equal(std_.round(4), expected_std)
 
-    mean_, std_ = find_feature_importance(X, y, rf, cv.split(X, y), scoring)
+    mean_, std_ = find_feature_importance(
+        X=X,
+        y=y,
+        estimator=rf,
+        cv=cv.split(X, y),
+        groups=None,
```
solegalli, line 270: I don't think we need to add these changes where groups is None. Could we revert back to the original?

What we do need is a test for when groups is not None. Could we add that?

solegalli (commented):
Hey @KananMahammadli 

Thank you so much for adding this functionality. 

There are a few cosmetic changes required. Could you revert the import style back to what it was? we use black for styling. Not sure if you use it too?

There is an important change: I just added functionality to accept cv.split(x,y), so all these classes were modified. Could you please rebase main onto this branch?

And the last big ask: we need tests to check that group works for all classes when it is not None.

Would you be able to take care of these? Thanks a lot!

feature_engine/selection/base_selection_functions.py
KananMahammadli: actually, I refer to all steps mentioned in [code contribution](https://feature-engine.trainindata.com/en/latest/contribute/contribute_code.html), and when I used isort and black on this file, it changed format

feature_engine/selection/probe_feature_selection.py
KananMahammadli: same as above

feature_engine/selection/single_feature_performance.py
KananMahammadli: same as above

tests/test_selection/test_base_selection_functions.py
KananMahammadli: yes you are right, as groups=None is default behaviour we do not need to specify it, but rather test when actual groups is provided. I will try to cover tests in the following days

KananMahammadli: as an update to this test, the current format of the `find_feature_importance` and `single_feature_performance` is that none of the arguments have default value, therefore in these function I also didn't assign default None, I can either keep the format, and revert tests back to have groups=None, or assign None inside functions themselves.

KananMahammadli: can you please confirm if this style is ok for `single_feature_importance` and `find_feature_importance` at `base_selection_functions.py`?

```Python
def single_feature_performance(
    X: pd.DataFrame,
    y: pd.Series,
    variables: List[Union[str, int]],
    estimator,
    cv,
    scoring,
    groups=None,
):
````

and

```Python
def find_feature_importance(
    X: pd.DataFrame,
    y: pd.Series,
    estimator,
    cv,
    scoring,
    groups=None,
):
```

If I do not add default None value here, we have to pass groups=None in test files where we do not test groups

feature_engine/selection/base_selection_functions.py
KananMahammadli: I suspect difference can derive from whether we run isort first, then black or black first then isort

KananMahammadli:
- Tests have been added to check if groups perform exactly same as passing indices
- groups option has been removed from ShuffeFeatureSelection, as it randomly changes values of features, which changes group order too, causing abnormal behaviour

@solegalli can you please check if everything is ok?

feature_engine/selection/base_selection_functions.py
solegalli: yes, it comes from there. I normally run first isort and then black.

tests/test_selection/test_base_selection_functions.py
solegalli: this is good, thanks!

feature_engine/selection/single_feature_performance.py
```diff
@@ -32,10 +33,8 @@
 from feature_engine.selection.base_selector import BaseSelector
 from feature_engine.tags import _return_tags
 
-from .base_selection_functions import (
-    _select_numerical_variables,
-    single_feature_performance,
-)
+from .base_selection_functions import (_select_numerical_variables,
```
solegalli, line 36: sorry to bother with this, could you please pass the imports to the following lines, so we keep it consistent?

tests/test_selection/test_base_selection_functions.py
```diff
@@ -184,6 +199,37 @@ def test_single_feature_performance_cv_generator(df_test):
...
+    )
+
+    mean_, std_ = single_feature_performance(
+        X=X,
+        y=y,
+        variables=variables,
+        estimator=rf,
+        cv=cv,
+        scoring=scoring,
+        groups=groups,
```
solegalli, line 226: If I understand this test correctly, we are comparing the results of the function `single_feature_performance` when groups is `None` with the results returned when groups is the `group` variable, and the results are exactly the same, which comes from the dataset containing random values, I guess. But then, we can't be sure whether the grouping variable is working or not, if passing None or something else return the same thing.

As painful as it is, we need to add hard coded values, or return the expected values by an alternative function(s), that is independent of the function we are trying to test.

solegalli (commented):
Hey @KananMahammadli 

Thank you so much for the quick turnaround! The code is looking good. The changes in the feature-engine module are good to go. There is only one small cosmetic request.

We need to make better tests. I left a comment on one of the tests, and since the same logic is used in all tests, then we need to update them all.

Tests need to be unequivocal. The safest way is to hard-code the expected values. Sometimes it is possible to obtain the expected results with independent functions. For example, we can write some logic to obtain the expected results with cross-validate, which we use under the hood and is independent from feature-engine, and then compare that with the output of our functions, but using our function to return the results that we want to use to test our own function is a bit circular. Better to avoid.

Could you take a look?

Thanks a lot!

feature_engine/selection/single_feature_performance.py
KananMahammadli: I put them back, sorry for having linting confusion again, isort doesn't work properly in my environment

tests/test_selection/test_base_selection_functions.py
KananMahammadli: if we have to make sure about the independent expected results for tests, then I will update it, otherwise let me explain the reason for the test logic I chose. Using Splitter object as cv and giving groups is clear, and for expected results, I do not use groups, and manually find the train and validation indexes for splits considering existence of the groups, then pass generator for list of indices to the class, so this is expected result that current feature_engine without groups will provide for us, hence adding groups and avoiding index based cv should provide exactly same result

solegalli: ahhhhh!!! clever. I missed that.

feature_engine/selection/single_feature_performance.py
solegalli: thank you!

solegalli:
Great work @KananMahammadli thank you so much!

Issue #763, opened 2024-05-15
transformer to apply mean normalization


solegalli:
in mean normalization, we subtract the mean from each value and then divide by the value range. This centres the variables at 0, and scales their values between -1 and 1. It is an alternative to standardization.

sklearn has no transformer to apply mean normalization. but we can combine the standard scaler and the robust scaler to do so. The thing is, that both transformers need to be fit over the raw data, so we can't use them within a pipeline, because the pipeline applies the transformation before the next transformer learns the required parameters.

My idea is to wrap both transformers within a class, so that fit is applied without transform, and then with those parameters, we can transform the data. See for example here: https://github.com/solegalli/Python-Feature-Engineering-Cookbook-Second-Edition/blob/main/ch07-scaling/Recipe-4-mean-normalization.ipynb

VascoSch92:
Instead of wrapping two transformers within a class, could also be an idea to create a brand new transformer which accomplish mean normalisation?

I understand that this could create some duplicated code, but on the other hand, the new mean normalisation transformer will be easy to understand and to debug.

I'm just thinking :-)

solegalli:
Yes, it's perhaps a better idea.

solegalli:
By any chance, would you like to give this one a go? creating a new class? and a new module. We don't have a module for scaling yet. @VascoSch92

VascoSch92:
Yes I can give it a try :-) 

Just a little summary (please correct me if I'm wrong):
1. the idea is to create a new module scaling which contains transformers to scale columns
2. the first of these transformers will be the mean_normalization transformer right?


On a side note: there is still this [PR](https://github.com/feature-engine/feature_engine/pull/694) on the distance transformer which attend a review :-) I corrected it and now should be better. Of course, if you have time :-D

solegalli:
Closing as completed

PR #806, opened 2024-08-29
Mean Normalisation Scaling

VascoSch92:
First version of the `MeanNormalizationScaling` as discussed in #763 

I create a new module scaling as discussed.

Probably, you will also add new scaling in this module.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -296,6 +296,15 @@ types and variable names match.
 - :doc:`api_doc/preprocessing/MatchCategories`: ensures categorical variables are of type 'category'
 - :doc:`api_doc/preprocessing/MatchVariables`: ensures that columns in test set match those in train set
 
+Scaling:
+~~~~~~~~
+
+Scaling the data can help to balance the impact of all variables on the model, and can improve the
+its performance.
+
+- :doc:`api_doc/scaling/MeanNormalizationScaler`: scale a given set of columns using mean normalization
+
+
 Scikit-learn Wrapper:
 ~~~~~~~~~~~~~~~~~~~~~
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -67,6 +67,7 @@ Feature-engine includes transformers for:
 - Datetime features
 - Time series
 - Preprocessing
+- Scaling
 
 Feature-engine transformers are fully compatible with scikit-learn. That means that you can assemble Feature-engine
 transformers within a Scikit-learn pipeline, or use them in a grid or random search for hyperparameters.
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -299,7 +299,7 @@ types and variable names match.
 Scaling:
 ~~~~~~~~
 
-Scaling the data can help to balance the impact of all variables on the model, and can improve the
+Scaling the data can help to balance the impact of all variables on the model, and can improve 
 its performance.
 
 - :doc:`api_doc/scaling/MeanNormalizationScaler`: scale a given set of columns using mean normalization
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -302,7 +302,7 @@ Scaling:
 Scaling the data can help to balance the impact of all variables on the model, and can improve 
 its performance.
 
-- :doc:`api_doc/scaling/MeanNormalizationScaler`: scale a given set of columns using mean normalization
+- :doc:`api_doc/scaling/MeanNormalizationScaler`: scale variables using mean normalization
 
 
 Scikit-learn Wrapper:
--- a/feature_engine/scaling/mean_normalization.py
+++ b/feature_engine/scaling/mean_normalization.py
@@ -60,7 +60,6 @@ class MeanNormalizationScaler(BaseNumericalTransformer):
 
     range_:
         a pd.Series containing the range of the selected columns, i.e., the difference
-        between the max and the min
 
     {variables_}
 
--- a/feature_engine/scaling/mean_normalization.py
+++ b/feature_engine/scaling/mean_normalization.py
@@ -56,7 +56,7 @@ class MeanNormalizationScaler(BaseNumericalTransformer):
     Attributes
     ----------
     mean_:
-        a dictionary containing the mean, max and min of every given variable
+        Dictionary containing the mean of the variables.
 
     range_:
         a pd.Series containing the range of the selected columns, i.e., the difference
--- a/feature_engine/scaling/mean_normalization.py
+++ b/feature_engine/scaling/mean_normalization.py
@@ -121,11 +121,11 @@ class MeanNormalizationScaler(BaseNumericalTransformer):
 
         # check input dataframe
         X = super().fit(X)
-        self.mean_ = X[self.variables_].mean()
-        self.range_ = X[self.variables_].max() - X[self.variables_].min()
+        self.mean_ = X[self.variables_].mean().to_dict()
+        self.range_ = (X[self.variables_].max() - X[self.variables_].min()).to_dict()
 
         # check for constant columns
-        constant_columns = self.range_[self.range_ == 0].index.tolist()
+        constant_columns = [col for col, value in self.range_.items() if value == 0]
         if constant_columns:
             raise ValueError(
                 "Division by zero in the scaling. This is because \n"
@@ -153,9 +153,7 @@ class MeanNormalizationScaler(BaseNumericalTransformer):
         X = self._check_transform_input_and_state(X)
 
         # transformation
-        X[self.variables_] = (
-            X[self.variables_] - self.mean_[self.variables_]
-        ) / self.range_[self.variables_]
+        X[self.variables_] = (X[self.variables_] - self.mean_) / self.range_
 
         return X
 
@@ -178,9 +176,6 @@ class MeanNormalizationScaler(BaseNumericalTransformer):
         X = self._check_transform_input_and_state(X)
 
         # inverse transform
-        X[self.variables_] = (
-            X[self.variables_] * self.range_[self.variables_]
-            + self.mean_[self.variables_]
-        )
+        X[self.variables_] = X[self.variables_] * self.range_ + self.mean_
 
         return X
--- a/feature_engine/scaling/mean_normalization.py
+++ b/feature_engine/scaling/mean_normalization.py
@@ -1,0 +1,194 @@
+# Authors: Vasco Schiavo <vasco.schiavo@protonmail.com>
+# License: BSD 3 clause
+
+from typing import Dict, List, Optional, Union
+
+import pandas as pd
+
+from feature_engine._base_transformers.base_numerical import BaseNumericalTransformer
+from feature_engine._check_init_parameters.check_variables import (
+    _check_variables_input_value,
+)
+from feature_engine._docstrings.fit_attributes import (
+    _feature_names_in_docstring,
+    _n_features_in_docstring,
+    _variables_attribute_docstring,
+)
+from feature_engine._docstrings.init_parameters.all_trasnformers import (
+    _variables_numerical_docstring,
+)
+from feature_engine._docstrings.methods import (
+    _fit_not_learn_docstring,
+    _fit_transform_docstring,
+    _inverse_transform_docstring,
+)
+from feature_engine._docstrings.substitute import Substitution
+
+
+@Substitution(
+    variables=_variables_numerical_docstring,
+    variables_=_variables_attribute_docstring,
+    feature_names_in_=_feature_names_in_docstring,
+    n_features_in_=_n_features_in_docstring,
+    fit=_fit_not_learn_docstring,
+    fit_transform=_fit_transform_docstring,
+    inverse_transform=_inverse_transform_docstring,
+)
+class MeanNormalizationScaling(BaseNumericalTransformer):
+    """
+    The MeanNormalizationScaling() applies the mean normalization scaling techinques
+    to one or multuple columns of a dataframe.
+
+    Mean normalization is a way to implement feature scaling. Mean normalization
+    calculates and subtracts the mean of for every feature and divide this value
+    by the range (max - min).
+
+    A list of variables can be passed as an argument. Alternatively, the transformer
+    will automatically select and transform all variables of type numeric.
+
+    If a column is constant, then it will be transformed to the zero constant column.
+
+    Parameters
+    ----------
+    {variables}
+
+
+    Attributes
+    ----------
+    params:
+        a dictionary containing the mean, max and min of every given variable
+
+    {variables_}
+
+    {feature_names_in_}
+
+    {n_features_in_}
+
+    Methods
+    -------
+    {fit}
+
+    {fit_transform}
+
+    {inverse_transform}
+
+    transform:
+        Scale the variables using mean normalization.
+
+    Examples
+    --------
+
+    >>> import numpy as np
+    >>> import pandas as pd
+    >>> from feature_engine.scaling import MeanNormalizationScaling
+    >>> np.random.seed(42)
+    >>> X = pd.DataFrame(dict(x = np.random.lognormal(size = 100)))
+    >>> mns = LogTransformer()
+    >>> mns.fit(X)
+    >>> X = mns.transform(X)
+    >>> X.head()
+            x
+    0  0.496714
+    1 -0.138264
+    2  0.647689
+    3  1.523030
+    4 -0.234153
+    """
+
+    def __init__(
+        self,
+        variables: Union[None, int, str, List[Union[str, int]]] = None,
+    ) -> None:
+
+        self.variables = _check_variables_input_value(variables)
+        # the following variable will be populated in the during the fit
+        self.params: Dict = {}
+
+    def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
+        """
+        The method populate the variable `params`.
+
+        Parameters
+        ----------
+        X: Pandas DataFrame of shape = [n_samples, n_features].
+            The training input samples. Can be the entire dataframe, not just the
+            variables to transform.
+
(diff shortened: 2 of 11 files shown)


feature_engine/scaling/mean_normalization.py
```diff
@@ -0,0 +1,194 @@
+# Authors: Vasco Schiavo <vasco.schiavo@protonmail.com>
```
solegalli, line 1: another protonmail fan, I like it!

feature_engine/scaling/mean_normalization.py
```diff
@@ -0,0 +1,194 @@
...
+    """
+
+    def __init__(
+        self,
+        variables: Union[None, int, str, List[Union[str, int]]] = None,
+    ) -> None:
+
+        self.variables = _check_variables_input_value(variables)
+        # the following variable will be populated in the during the fit
+        self.params: Dict = {}
```
solegalli, line 105: in init, we can only have parameters that the user can tweak. So either we expose it, or we assign it in fit.

feature_engine/scaling/mean_normalization.py
```diff
@@ -0,0 +1,194 @@
...
+            variables to transform.
+
+        y: pandas Series, default=None
+            It is not needed in this transformer. You can pass y or None.
+        """
+
+        # check input dataframe
+        X = super().fit(X)
+
+        for variable in self.variables_:
```
solegalli, line 124: i think the look is not optimal. X[self.variables_].mean() would return all means automatically, and the same for the below values. We should create a dictionary for mean, one for max and one for min, so then we can apply it at once to the entire dataframe without looping.

solegalli (commented):
Hey @VascoSch92 

This is looking really good. Thank you for the first draft.

I think we could tidy it a bit so that we don't loop neither in fit nor in transform.

Could you take a look?

Thank you!

feature_engine/scaling/mean_normalization.py
VascoSch92: I think it is nice if the user can see the fit parameters for the scaling, i.e., `mean`, `max`, `min`.

Or what do you think? Should I put it in the `fit` method?

Another question: private parameters ends with an underscore right? so it should be `params_` ?
I'm taking as inspiration `variables` and `variables_`.

feature_engine/scaling/mean_normalization.py
```diff
@@ -0,0 +1,194 @@
...
+        -------
+        X_new: pandas dataframe
+            The dataframe with the transformed variables.
+        """
+
+        # check input dataframe and if class was fitted
+        X = self._check_transform_input_and_state(X)
+
+        # transformation
+        for variable in self.variables_:
```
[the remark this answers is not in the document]
VascoSch92: Yes this is true. The problem is that we have not just min max and mean but an also the dimension given by the columns. I think we can create a pandas data frame with all the info we need and then use it to do the transformations in a vectorised way.

feature_engine/scaling/mean_normalization.py
```diff
@@ -0,0 +1,194 @@
...
+            numerator = X[variable] - self.params[variable]["mean"]
+            denominator = self.params[variable]["max"] - self.params[variable]["min"]
+
+            # If max and min are equal, then the column's variable is constant.
+            # We set denominator to 1, and the column will be normalized to 0
+            if denominator == 0:
```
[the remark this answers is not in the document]
VascoSch92: As you want. I was following `Sklearn`. In fact, if you have zero denominator, then `Sklearn` don't thrown an error but divide by 1 (or don't divide).

I will change to raise an error ;-)

feature_engine/scaling/mean_normalization.py
solegalli: parameters ending in _ are learned during fit. That is sklearn convention. Private parameters start with _. So variables_ is learned during fit, and _variables would be private.

Yes, min,  max and min should be exposed to the user and learned in fit :)

VascoSch92:
Hey @solegalli 

I addressed your comments.

Please let me know what you think :-)

feature_engine/scaling/mean_normalization.py
```diff
@@ -0,0 +1,185 @@
...
+            The training input samples. Can be the entire dataframe, not just the
+            variables to transform.
+
+        y: pandas Series, default=None
+            It is not needed in this transformer. You can pass y or None.
+        """
+
+        # check input dataframe
+        X = super().fit(X)
+        self.params_ = pd.DataFrame(
```
solegalli, line 121: Do we need a dataframe? can we not make it work with dictionaries? Dataframes are heavier.

We'd need 3 dictionaries, one for mean, one for min and one for max. If we can make the code work with dictionaries, without transposing, this will be faster.

feature_engine/scaling/mean_normalization.py
```diff
@@ -0,0 +1,185 @@
...
+                "min": X[self.variables_].min(),
+            },
+        ).transpose()
+
+        # check for constant columns
+        constant_columns = self.params_.columns[
+            self.params_.loc["min"] == self.params_.loc["max"]
+        ].to_list()
+        if constant_columns:
+            raise ValueError(f"The following column/s are constant: {constant_columns}")
```
solegalli, line 134: can we make the error more informative? What's the problem with constant columns? We should try to squeeze that info in.

feature_engine/scaling/mean_normalization.py
VascoSch92: Ah yes no problem. It should fast to change from `df` to `dict` ;)

feature_engine/scaling/mean_normalization.py
```diff
@@ -0,0 +1,185 @@
...
+
+        y: pandas Series, default=None
+            It is not needed in this transformer. You can pass y or None.
+        """
+
+        # check input dataframe
+        X = super().fit(X)
+        self.params_ = pd.DataFrame(
+            {
+                "mean": X[self.variables_].mean(),
```
solegalli, line 123: I am having second thoughts here. I think it is better to store the range (max-min), so that we make this computation once, and then we don't need to make it every time we transform the data. The transform will be faster. We should just store mean and range, if possible as dictionaries.

feature_engine/scaling/mean_normalization.py
```diff
@@ -0,0 +1,185 @@
...
+@Substitution(
+    variables=_variables_numerical_docstring,
+    variables_=_variables_attribute_docstring,
+    feature_names_in_=_feature_names_in_docstring,
+    n_features_in_=_n_features_in_docstring,
+    fit=_fit_not_learn_docstring,
+    fit_transform=_fit_transform_docstring,
+    inverse_transform=_inverse_transform_docstring,
+)
+class MeanNormalizationScaling(BaseNumericalTransformer):
```
solegalli, line 37: let's use Scaler as name, MeanNormalizationScaler so we keep it like sklearn's names.

solegalli (commented):
Hi @VascoSch92 

This is looking really good. The tests are great. 

Regarding the logic, I think we can speed this up if we store the range instead of min max,, and if we use dictionaries instead of dataframes. Could you give that a go?

Feel free to start working on the dosctrings and on adding a user guide :)

VascoSch92:
Hey @solegalli 

 I changed what requested.
Now,

- `params_` is a dictionary `Dict[str, pd.Series]` with keys `'mean'` and '`range`'
- `MeanNormalizationScaling` -> `MeanNormalizationScaler`
- error message for constant columns is a little better :-)

feature_engine/scaling/mean_normalization.py
```diff
@@ -0,0 +1,186 @@
...
+        X_new: pandas dataframe
+            The dataframe with the transformed variables.
+        """
+
+        # check input dataframe and if class was fitted
+        X = self._check_transform_input_and_state(X)
+
+        # transformation
+        X[self.variables_] = (
+            X[self.variables_] - self.params_["mean"][self.variables_]
```
solegalli, line 157: is the self.variables_  in `self.params_["mean"][self.variables_]`  necessary?

feature_engine/scaling/mean_normalization.py
```diff
@@ -0,0 +1,186 @@
...
+            The training input samples. Can be the entire dataframe, not just the
+            variables to transform.
+
+        y: pandas Series, default=None
+            It is not needed in this transformer. You can pass y or None.
+        """
+
+        # check input dataframe
+        X = super().fit(X)
+        self.params_ = {
```
solegalli, line 121: this is OK. I think I still prefer having one dict for mean and one dict for range. Otherwise the dict might get too big and it might get confusing when the data has too many variables. Sklearn has 2 attributes, 1 for mean or center and one for scale or range, we would keep it consistent with their api as well if we create 2 attributes.

feature_engine/scaling/mean_normalization.py
```diff
@@ -0,0 +1,186 @@
...
+    variables_=_variables_attribute_docstring,
+    feature_names_in_=_feature_names_in_docstring,
+    n_features_in_=_n_features_in_docstring,
+    fit=_fit_not_learn_docstring,
+    fit_transform=_fit_transform_docstring,
+    inverse_transform=_inverse_transform_docstring,
+)
+class MeanNormalizationScaler(BaseNumericalTransformer):
+    """
+    The MeanNormalizationScaler() applies the mean normalization scaling techinques
```
solegalli, line 39: Here I'd say that the meannormalizationscaler substracts the mean and then divides the result by the value range. And add a small formula of the trasformation

feature_engine/scaling/mean_normalization.py
```diff
@@ -0,0 +1,186 @@
...
+    to one or multuple columns of a dataframe.
+
+    Mean normalization is a way to implement feature scaling. Mean normalization
```
solegalli, line 42: Yes, more or less like this.

feature_engine/scaling/mean_normalization.py
```diff
@@ -0,0 +1,186 @@
...
+    calculates and subtracts the mean of for every feature and divide this value
+    by the range (max - min).
+
+    A list of variables can be passed as an argument. Alternatively, the transformer
+    will automatically select and transform all variables of type numeric.
+
+    If a column is constant, then it will be transformed to the zero constant column.
+
```
solegalli, line 50: here we need to add a sentence that says more details in the user guide and create a user guide file with more detail about this transformation and an example with a real dataframe.

Here we create the user guide files: https://github.com/feature-engine/feature_engine/tree/main/docs/user_guide

We need a new folder for scaling, there an index, and then the file for the meannormalization scaler.

And also, to build the docs, we need to add a file to the api: https://github.com/feature-engine/feature_engine/tree/main/docs/api_doc  here we need a folder for scaling and in that folder an index, and the file for the scaler. 

If you check out the folder transformation inside those folders, you can very quickly get an idea of the files that we need and their format. The one for the api is super straightforward, the one for the user guide needs more work but I can support.

feature_engine/scaling/mean_normalization.py
```diff
@@ -0,0 +1,186 @@
...
+    def __init__(
+        self,
+        variables: Union[None, int, str, List[Union[str, int]]] = None,
+    ) -> None:
+
+        self.variables = _check_variables_input_value(variables)
+
+    def fit(self, X: pd.DataFrame, y: Optional[pd.Series] = None):
+        """
+        The method populate the variable `params_`.
```
solegalli, line 107: Could we replace with this:

Finds the mean and value range of each variable.

solegalli (commented):
Hey @VascoSch92 really good work here. Thank you so much!

We need to add a few files to create the docs now. Would you be able to do that as well?

Thanks a lot for the hard work.

solegalli:
Last but not least: we need to add the new module on the readme and on the frontpage of the documentation, which lives here: https://github.com/feature-engine/feature_engine/blob/main/docs/index.rst

Thank you!!

feature_engine/scaling/mean_normalization.py
VascoSch92: I think yes. In this way you are taking on both side the columns in the right order

feature_engine/scaling/mean_normalization.py
VascoSch92: Ok so let's do `self.mean_` and `self.range_`

VascoSch92:
Hey @solegalli 

- added the documentation for the new scaler
- divided `params_` into `mean_` and `range_`. Now `mean_` and `range_` are `pd.Series`. We can also make them `np.array` if you want. `Sklearn` uses np.arrays, should we make the same?

VascoSch92:
Question: I added into the docstring this formula
```text
..math:: \frac{X_col - X_{mean}}{X_{max} - X_{min}}
```
is it supported by the docs of feature engine?

feature_engine/scaling/mean_normalization.py
VascoSch92: but I think that we can also do it with numpy array actually

docs/index.rst
```diff
@@ -296,6 +296,15 @@ types and variable names match.
 - :doc:`api_doc/preprocessing/MatchCategories`: ensures categorical variables are of type 'category'
 - :doc:`api_doc/preprocessing/MatchVariables`: ensures that columns in test set match those in train set
 
+Scaling:
+~~~~~~~~
+
+Scaling the data can help to balance the impact of all variables on the model, and can improve the
```
solegalli, line 302:
```suggestion
Scaling the data can help to balance the impact of all variables on the model, and can improve 
```

feature_engine/scaling/mean_normalization.py
```diff
@@ -0,0 +1,187 @@
...
+
+    Parameters
+    ----------
+    {variables}
+
+
+    Attributes
+    ----------
+    mean_:
+        a dictionary containing the mean, max and min of every given variable
```
solegalli, line 59:
```suggestion
        Dictionary containing the mean of the variables.
```

feature_engine/scaling/mean_normalization.py
```diff
@@ -0,0 +1,187 @@
...
+
+    range_:
+        a pd.Series containing the range of the selected columns, i.e., the difference
```
solegalli, line 62:
```suggestion
        Dictionary containing the value range of of the variables.
```

feature_engine/scaling/mean_normalization.py
```diff
@@ -0,0 +1,187 @@
...
+        between the max and the min
```
solegalli, line 63:
```suggestion
```

feature_engine/scaling/mean_normalization.py
```diff
@@ -0,0 +1,187 @@
...
+            The training input samples. Can be the entire dataframe, not just the
+            variables to transform.
+
+        y: pandas Series, default=None
+            It is not needed in this transformer. You can pass y or None.
+        """
+
+        # check input dataframe
+        X = super().fit(X)
+        self.mean_ = X[self.variables_].mean()
```
solegalli, line 125:
```suggestion
        self.mean_ = X[self.variables_].mean().to_dict()
```

feature_engine/scaling/mean_normalization.py
```diff
@@ -0,0 +1,187 @@
...
+        self.range_ = X[self.variables_].max() - X[self.variables_].min()
```
solegalli, line 126: can we make this a dictionary? We use dictionaries in feature-engine :)

feature_engine/scaling/mean_normalization.py
```diff
@@ -0,0 +1,187 @@
...
+        X_new: pandas dataframe
+            The dataframe with the transformed variables.
+        """
+
+        # check input dataframe and if class was fitted
+        X = self._check_transform_input_and_state(X)
+
+        # transformation
+        X[self.variables_] = (
+            X[self.variables_] - self.mean_[self.variables_]
```
solegalli, line 158: If I remember from how I implemented pandas operations combined with dictionaries, pandas does automatically match keys with columns, so from memory, I don't think we need to pass self.variables_ to the dictionary. Could you test? remove it and see if the tests fail. Thank you!

feature_engine/scaling/mean_normalization.py
```diff
@@ -0,0 +1,187 @@
...
+        ) / self.range_[self.variables_]
```
solegalli, line 159: same here

feature_engine/scaling/mean_normalization.py
```diff
@@ -0,0 +1,187 @@
...
+        X_tr: pandas dataframe
+            The dataframe with the transformed variables.
+        """
+
+        # check input dataframe and if class was fitted
+        X = self._check_transform_input_and_state(X)
+
+        # inverse transform
+        X[self.variables_] = (
+            X[self.variables_] * self.range_[self.variables_]
```
solegalli, line 183: same here

solegalli (commented):
Hey @VascoSch92 thanks for the quick turnaround. The api docs look great.

VascoSch92:
Hey @solegalli 

I changed to dictionaries instead of `pd.Series`, and it works :-)

solegalli:
Amazing! Thanks a lot!

We just need to add a description / demo to the user guide folder in the docs and we are good to go then :)

VascoSch92:
@solegalli quick question: in general feature engine use scaling transformers from sklearn. Should we also include examples of these transformers?

VascoSch92:
Hey @solegalli 

I updated the docs with a demo.

Let me know what do you think. It is just a first version :-)

VascoSch92:
Hey @solegalli :-) did you have time to look at the latest changes?

solegalli:
Pending acceptance of suggested changes: https://github.com/VascoSch92/feature_engine/pull/1

PR #813, opened 2024-09-17
Update base_imputer.py

Astha0024:
removing infer because it changed manually converted object columns back to int or float

### DIFF
--- a/feature_engine/imputation/base_imputer.py
+++ b/feature_engine/imputation/base_imputer.py
@@ -61,7 +61,7 @@ class BaseImputer(BaseEstimator, TransformerMixin, GetFeatureNamesOutMixin):
 
         # Replace missing data with learned parameters
         with pd.option_context("future.no_silent_downcasting", True):
-            X = X.fillna(value=self.imputer_dict_).infer_objects(copy=False)
+            X = X.fillna(value=self.imputer_dict_)
         return X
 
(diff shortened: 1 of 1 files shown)


solegalli:
Hi @Astha0024 thanks for the PR.

I remember vaguely adding the infer recently due to some pandas upgrade or something, By removing it, other tests fail.

What's the issue that you are having? could you add an example?

Astha0024:
So sometimes a column that could be inferred at numerical needs to be cast to object type for example : number of bathrooms in a house. Then when we pass it through this transformation, the infer converts it back to int and then we can't run encoding on top of it.

solegalli:
This seems to be a problem only for the CategoricalImputer, which is the one that takes objects. We need to ensure that it returns objects as outputs.

- [ ] add test to corroborate that integer column cast as object is returned as object after imputation
- [ ] fix functionality if test fails

solegalli:
The `CategoricalEncoder` has the option to return numerical variables that were encoded as object both as **object** or as **number** after the imputation.

To return the variables as object, ensure you set the parameter `return_object=True` when setting up the transformer:

For example:

```
    imputer = CategoricalImputer(
        imputation_method="frequent",
        variables=["City", "Studies", "Marks"],
        return_object=True,
    )
```

PR #829, opened 2024-12-03
Add `PandasCategoricalEncoder`

ClaudioSalvatoreArcidiacono:
Fixes 828


solegalli:
This transformer is intended to modify data types and not data values, hence I don't think it is suitable for feature engine.

PR #837, opened 2025-01-21
enable testing of different scikit-learn versions

solegalli:


solegalli:
Update this PR to test with sklearn 1.5 and 1.6 instead of 1 and 1.4.

PR #859, opened 2025-04-30
Remove `switch` arg in `DropHighPSIFeatures`

### LINKED ISSUE #858: The `switch` argument in `DropHighPSIFeatures` should be removed
**Describe the bug**
In `DropHighPSIFeatures` it is stated that the PSI metric is not symmetric, and hence a boolean argument `switch` is introduced. The PSI however is symmetric by definition, so this argument should be removed.

PSI(a, b) = PSI(b, a) by definition, as it is a symmetrized version of the KL divergence. See p. 85 of [Kullback & Leibler 1951](https://projecteuclid.org/journals/annals-of-mathematical-statistics/volume-22/issue-1/On-Information-and-Sufficiency/10.1214/aoms/1177729694.full).

sancholp:
Closes #858.

PSI is a symmetric metric and as such the `switch` arg doesn't add functionality.


sancholp:
@solegalli Whenever you find time please have a look

fkiraly:
two remarks from me (just browsing through PR):

* I think it is indeed correct, PSI is symmetric. To see this, switch the two batches and observe that the difference term and the log term both switch signs.
* however, removing an existing argument from an sklearn class is a breaking change. If this would be sklearn, it would have to go through a deprecation, i.e., a warning is raised that the argument will be removed, so users can adapt and their code does not break in case they use the arg.

solegalli:
As per discussion in #858, we won't remove the argument, we'll make it clear in the docs that the PSI is symetric, but the estimator calculates it is not.

solegalli:
Waiting for this PR https://github.com/sancholp/feature_engine/pull/1 to merge.

solegalli:
For some reason, this didn't work. I reverted the files to the original ones to only change the docstrings but something went wrong somewhere.

PR #862, opened 2025-07-02
[MNT] migrate various config files to `pyproject.toml`

fkiraly:
Not sure if it is desired, I made a try to migrate various config files to `pyproject.toml`.

For context, `setup.py` is officially considered legacy by python, see PEP 518 and PEP 621.


ChristopherGS:
> @ChristopherGS is this something that we want? what would be the advantage of this migration?

@solegalli pyproject.toml is the modern way to do things, as @fkiraly points out (see the PEPs). The main advantage will be making the library seem more up-to-date and hence giving off better vibes.

solegalli:
I am always up for giving off better vibes :)

Thank you both!

fkiraly:
PS: I like to avoid rebase and only use merge.
Rebase is quite error prone, especially for newcomers, so in my personal opinion merge from `main` is the choice that is more "welcoming" as a standard for a project. You squash PR anyway (a good choice imo as it keeps the `main` history "clean"), so there is no need to retain history.

PR #874, opened 2025-09-16
feat: Add DatetimeOrdinal transformer and unit tests (#818)

### LINKED ISSUE #818: Add DatetimeOrdinal
It'd be nice to have the ability to convert datetime columns to ordinal values. The DatetimeFeatures is nice for when one wants to extract the year, month, day, etc., but sometimes it's desirable to have the date simply as an ordinal value.

Here's an example I've used in code before:

```py
class DatetimeOrdinal(BaseEstimator, TransformerMixin):
    def fit(self, X, y=None):
        return self

    def transform(self, X):
        return X.apply(lambda x: x.map(pd.Timestamp.toordinal, na_action='ignore'))

    def inverse_transform(self, X):
        return X.apply(lambda x: x.map(pd.Timestamp.fromordinal, na_action='ignore'))
```

bermuchin:
Adds the DatetimeOrdinal, a new transformer to convert datetime variables
into their ordinal representation. This allows using date information as a
continuous numerical feature in machine learning models.

The transformer can automatically find datetime variables or work with a
user-defined list. It also provides an option to set a custom `start_date`
to calculate ordinals relative to a specific point in time, which can
help in creating more meaningful features.

Closes #818

### DIFF
--- a/tests/test_datetime/test_datetime_ordinal.py
+++ b/tests/test_datetime/test_datetime_ordinal.py
@@ -247,3 +247,23 @@ def test_datetime_ordinal_missing_values_raise_in_transform(
         ValueError, match="Some of the variables in the dataset contain NaN"
     ):
         transformer.transform(X_test)  # 3 columns + NA data
+
+
+def test_raises_error_for_invalid_missing_values():
+    with pytest.raises(
+        ValueError, match="missing_values takes only values 'raise' or 'ignore'"
+    ):
+        DatetimeOrdinal(missing_values="foo")
+
+
+def test_raises_error_for_invalid_drop_original():
+    with pytest.raises(
+        ValueError, match="drop_original takes only booleans True or False"
+    ):
+        DatetimeOrdinal(drop_original="bar")
+
+
+def test_more_tags_returns_expected_tags():
+    transformer = DatetimeOrdinal()
+    expected_tags = {"variables": "datetime"}
+    assert transformer._more_tags() == expected_tags
--- a/README.md
+++ b/README.md
@@ -143,6 +143,7 @@ Please share your story by answering 1 quick question
 ### Datetime
  * DatetimeFeatures
  * DatetimeSubtraction
+ * DatetimeOrdinal
  
 ### Time Series
  * LagFeatures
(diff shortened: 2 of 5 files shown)


bermuchin:
Hi, @solegalli. Can I get a review?

tests/test_datetime/test_datetime_ordinal.py
```diff
@@ -0,0 +1,224 @@
...
+        [d.toordinal() - start_ordinal + 1 for d in df_datetime_ordinal["date_col_1"]],
+        name="date_col_1_ordinal",
+    )
+
+    pd.testing.assert_series_equal(
+        X_transformed["date_col_1_ordinal"], expected_ordinal
+    )
+
+
+def test_datetime_ordinal_no_variables_specified(df_datetime_ordinal):
```
solegalli: very nitpick: this test could be put together with test_datetime_ordinal_no_start_date and pass the variable names or None with a pytest fixture, because the test and expected results (test logic) are identical.

tests/test_datetime/test_datetime_ordinal.py
```diff
@@ -0,0 +1,224 @@
...
+        X_transformed["date_col_2_ordinal"], expected_ordinal_2
+    )
+    assert "non_date_col" in X_transformed.columns
+    assert "date_col_1" not in X_transformed.columns
+    assert "date_col_2" not in X_transformed.columns
+
+
+def test_datetime_ordinal_missing_values_raise(df_datetime_ordinal_na):
+    transformer = DatetimeOrdinal(missing_values="raise")
+    with pytest.raises(ValueError):
```
solegalli: I normally like testing that the error message matches the expected message as well, because sometimes it fails somewhere else and then we can't be sure.

tests/test_datetime/test_datetime_ordinal.py
```diff
@@ -0,0 +1,224 @@
...
+        transformer.fit(df_datetime_ordinal_na)
```
solegalli, line 117: it should also fail in transform, so we could add that test as well.

tests/test_datetime/test_datetime_ordinal.py
```diff
@@ -0,0 +1,224 @@
...
+    pd.testing.assert_series_equal(
+        X_transformed["date_col_1_ordinal"], expected_ordinal_1
+    )
+    pd.testing.assert_series_equal(
+        X_transformed["date_col_2_ordinal"], expected_ordinal_2
+    )
+
+
+def test_datetime_ordinal_invalid_start_date():
+    with pytest.raises(ValueError):
```
solegalli: I like testing that the error message matches the expected one.

bermuchin:
Hi @solegalli, I've finally pushed all the updates! (tests, docs, style, and coverage).
How does it look now?
Thanks as always for your review!

PR #881, opened 2026-01-08
Add GeoDistanceTransformer for geographic distance calculation

ankitlade12:
- Add GeoDistanceTransformer class for calculating distances between coordinate pairs
- Support haversine, euclidean, and manhattan distance methods
- Support km, miles, meters, and feet output units
- Add validate_ranges parameter to skip coordinate validation
- Add comprehensive tests with pytest parametrize
- Add user guide and API documentation
- Add pipeline test to test_check_estimator_creation.py


docs/api_doc/creation/GeoDistanceTransformer.rst
```diff
@@ -0,0 +1,6 @@
+GeoDistanceTransformer
+======================
+
```
solegalli: we need to remove this file from the PR

PR #884, opened 2026-01-28
[MNT] add support for pandas 3

solegalli:
fyi @ankitlade12

### DIFF
--- a/tests/test_creation/test_math_features.py
+++ b/tests/test_creation/test_math_features.py
@@ -79,12 +79,6 @@ def test_error_new_variable_names_not_permitted():
             func=["sum", "mean"],
             new_variables_names=["sum_of_two_vars", "sum_of_two_vars"],
         )
-    with pytest.raises(ValueError):
-        MathFeatures(
-            variables=variables,
-            func=["sum", "mean"],
-            new_variables_names="sum_of_two_vars"
-        )
 
 
 def test_aggregations_with_strings(df_vartypes):


solegalli:
Will do. Thanks a lot! I only have internet again from Saturday though. I ll do my best o have it ready by beginning of next week.

-------- Original Message --------

tests/test_encoding/test_similarity_encoder.py
```diff
@@ -150,6 +150,30 @@ def test_nan_behaviour_ignore(df_enc_big_na):
...
+    # Test StringDtype with pd.NA to hit "<NA>" branch in transform
+    df = pd.DataFrame({"var_A": ["A", "B", pd.NA]}, dtype="string")
+    encoder = StringSimilarityEncoder(missing_values="impute")
+    X = encoder.fit_transform(df)
+    assert (X.isna().sum() == 0).all(axis=None)
+    # The categories will include "<NA>" or the string version of it
+    assert (
+        "<NA>" in encoder.encoder_dict_["var_A"]
+        or "" in encoder.encoder_dict_["var_A"]
+    )
```
solegalli: Hi @ankitlade12 

**Quick question**: why do we need this line: `"<NA>" in encoder.encoder_dict_["var_A"]`?

This line alone (`"" in encoder.encoder_dict_["var_A"]`) seems enough to pass the test in both pandas 3 and 2.

tests/test_encoding/test_similarity_encoder.py
```diff
@@ -139,15 +151,48 @@ def test_nan_behaviour_impute(df_enc_big_na):
...
+    df_enc_big_na.loc[0, "var_A"] = nan_value
+
     encoder = StringSimilarityEncoder(missing_values="ignore")
     X = encoder.fit_transform(df_enc_big_na)
     assert (X.isna().any(axis=1) == df_enc_big_na.isna().any(axis=1)).all()
-    assert encoder.encoder_dict_ == {
-        "var_A": ["B", "D", "G", "A", "C", "E", "F"],
-        "var_B": ["A", "D", "B", "G", "C", "E", "F"],
-        "var_C": ["C", "D", "B", "G", "A", "E", "F"],
-    }
```
solegalli, line 150: Hi @ankitlade12 @glevv 

There is a discrepancy on how SimilarityEncoder treats None, respect of np.nan or pd.NA.

Briefly, when missing_values is raise or impute, the treatment is the same. But when missing_values is ignore, SimilarityEncoder will not include the nan value in the encoder_dict if np.nan, or pd.NA, bit it will if None.

Is this something we are happy with?

tests/test_encoding/test_similarity_encoder.py
```diff
@@ -139,15 +151,48 @@ def test_nan_behaviour_impute(df_enc_big_na):
...
+
+
+def test_string_dtype_with_pd_na():
+    # Test StringDtype with pd.NA to hit "<NA>" branch in transform
+    df = pd.DataFrame({"var_A": ["A", "B", pd.NA]}, dtype="string")
+    encoder = StringSimilarityEncoder(missing_values="impute")
+    X = encoder.fit_transform(df)
+    assert (X.isna().sum() == 0).all(axis=None)
+    # The categories will include "<NA>" or the string version of it
+    assert "" in encoder.encoder_dict_["var_A"]
```
solegalli, line 176: Sorry @ankitlade12 I committed the change without wanting. Here it was:

```
    assert (
        "<NA>" in encoder.encoder_dict_["var_A"]
        or "" in encoder.encoder_dict_["var_A"]
    )
```

And my question was why the line  "<NA>" in encoder.encoder_dict_["var_A"] if the test seems to pass without it.

tests/test_encoding/test_similarity_encoder.py
solegalli: Apparently None is handled differently in pandas >=3 vs older versions. In >3, is handled as np.nan and pd.NA. In older versions it's handled as string.

tests/test_encoding/test_similarity_encoder.py
ankitlade12: Hi @solegalli,

The assertion `assert "" in encoder.encoder_dict_["var_A"]` is important because it verifies that the missing value was actually captured during [fit](cci:1://file:///Users/ankithemantlade/Desktop/feature_engine/feature_engine/encoding/similarity_encoder.py:214:4-292:19). 

When `missing_values="impute"`, the transformer is designed to replace `NaN` with `""` so that it becomes part of the learned vocabulary. This ensures that:

1. A specific feature is created to represent the missing values (e.g., `var_A_nan`).
2. Subsequent similarity calculations are made against this "missing" category specifically.

If we don't have this assertion, the test might still pass its "no NaNs in output" check, but we wouldn't be verifying if the encoder actually **learned** the missing category or if it's just treating them as unseen values.

tests/test_encoding/test_similarity_encoder.py
ankitlade12: Hi @solegalli,

You're absolutely right—the `"<NA>"` check is redundant in that assertion.

Because we use `.mask(series.isna(), "")` in the [impute](cci:1://file:///Users/ankithemantlade/Desktop/feature_engine/tests/test_encoding/test_similarity_encoder.py:136:0-150:5) logic, any `pd.NA` or `NaN` value is guaranteed to be converted to an empty string `""` before the vocabulary is learned. Therefore, `encoder_dict_` will only ever contain `""` for missing values, and the check for `"<NA>"` can be removed to keep the test cleaner.

I've updated the local tests to reflect this and confirmed that checking for `""` alone is sufficient for both pandas 2 and 3.

tests/test_encoding/test_similarity_encoder.py
ankitlade12: Hi @solegalli,

Great catch on the `None` discrepancy. You’re right—it’s caused by `None` being converted to the literal string `"None"` in older pandas versions, which slips through our current `.drop(["nan", "<NA>"])` filter as it isn't recognized as a "null" string.

To make this consistent across all versions, I propose we normalize it by using `.dropna()` **before** converting to a string in the `ignore` case:

```python
# Proposed consistent fix in fit():
self.encoder_dict_[var] = (
    X[var]
    .dropna()        # Consistently removes None, np.nan, pd.NA first
    .astype(str)     # Then convert remaining categories to string
    .value_counts()
    .head(self.top_categories)
    .index.tolist()
)

solegalli: Thank you. That resolves things :)

PR #887, opened 2026-02-01
[BUG ] fix performance warning in AddMissingIndicator

### LINKED ISSUE #886: Pandas fragmentation warning with AddMissingIndicator
Pandas fragmentation warning with AddMissingIndicator.

Tested behavior when using AddMissingIndicator in a pipeline with cross validation.

Assuming the fix would be to change line 164 and 165 of the AddMissingIndicator class from:

` indicator_names = [f"{feature}_na" for feature in self.variables_]`
`X[indicator_names] = X[self.variables_].isna().astype(int)`

To something like:

`indicator_df = (X[self.variables_].isna().astype(int).rename(columns=lambda c: f"{c}_na"))`
`X = pd.concat([X, indicator_df], axis=1)`

mo1998:
### Summary

This PR resolves the Pandas `PerformanceWarning: DataFrame is highly fragmented` that was triggered when using the `AddMissingIndicator` transformer.
The warning occurred because missing-indicator columns were previously added to the DataFrame one at a time via direct assignment, which is inefficient when handling many columns.

### What Changed

The `transform` method in `feature_engine/imputation/missing_indicator.py` has been refactored to improve performance and avoid DataFrame fragmentation:

* All missing indicators are now computed in a single step.
* A dedicated DataFrame is created to hold the indicator columns.
* The original DataFrame and the indicators DataFrame are merged using `pd.concat` in one operation, instead of repeated column assignments.

### Impact

* Eliminates the Pandas fragmentation warning.
* Improves performance and memory efficiency when generating many missing-indicator columns.
* No changes to the public API or transformer behavior.

### Related Issue

* Fixes #886

### Verification

* **Reproduction:** Confirmed that the warning no longer appears when running a script that previously triggered it.
* **Regression Testing:** All existing tests in `tests/test_imputation/` pass successfully.

### DIFF
--- a/tests/test_imputation/test_missing_indicator.py
+++ b/tests/test_imputation/test_missing_indicator.py
@@ -1,8 +1,10 @@
-import pytest
-from sklearn.pipeline import Pipeline
 import warnings
 import numpy as np
 import pandas as pd
+import pytest
+
+from sklearn.pipeline import Pipeline
+
 from feature_engine.imputation import AddMissingIndicator
 
 


mo1998:
I merged #885 Into codebase to fix the issue with pandas 3.0

tests/test_imputation/test_missing_indicator.py
```diff
@@ -96,3 +96,32 @@ def test_get_feature_names_out_from_pipeline(df_na):
 
     assert tr.get_feature_names_out(input_features=None) == feat_out
     assert tr.get_feature_names_out(input_features=original_features) == feat_out
+
+
+def test_no_performance_warning_with_many_variables():
+    # Test for issue #886: PerformanceWarning due to fragmentation
+    import numpy as np
+    import pandas as pd
+    import warnings
```
solegalli: How about this:

```
import warnings
import numpy as np
import pandas as pd


def test_no_performance_warning_with_many_variables():
    n_cols = 101
    df = pd.DataFrame(
        np.random.randn(10, n_cols),
        columns=[f"col_{i}" for i in range(n_cols)],
    )

    # Introduce missing values
    df.iloc[0, :] = np.nan

    ami = AddMissingIndicator(missing_only=False)
    ami.fit(df)

    with warnings.catch_warnings(record=True) as captured:
        warnings.simplefilter("always")
        ami.transform(df)

    assert not any(
        issubclass(w.category, pd.errors.PerformanceWarning)
        for w in captured
    ), "PerformanceWarning was raised during transform"
```

feature_engine/imputation/missing_indicator.py
```diff
@@ -161,8 +161,10 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
 
         X = self._transform(X)
 
-        indicator_names = [f"{feature}_na" for feature in self.variables_]
-        X[indicator_names] = X[self.variables_].isna().astype(int)
+        X_indicators = X[self.variables_].isna().astype(int)
```
solegalli: How about replacing the 2 lines with this?

X_indicators = (
    X[self.variables_]
    .isna()
    .astype("int8")
    .add_suffix("_na")
)

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.add_suffix.html

PR #891, opened 2026-02-07
[MNT] handle pandas future warnings

solegalli:


ankitlade12:
Hi @solegalli,

You are absolutely correct. Your understanding aligns with the direction Pandas is heading in terms of stricter categorical data handling.

### Why this is necessary:
1. **Stricter Type Enforcement**: Pandas is moving away from "implicit" behavior where it silently introduces `NaNs` for unseen categories. In future versions (anticipating Pandas 4.0), operations that result in unobserved categories are expected to be much stricter.
2. **Explicit is Better than Implicit**: By explicitly converting unseen values to `NaN` before the transformation, we avoid the `FutureWarning` that warns about "assignment to a categorical column involving new categories."
3. **Avoiding Reindexing Issues**: As noted in the recent changes to [match_columns.py](cci:7://file:///Users/ankithemantlade/Desktop/feature_engine/feature_engine/preprocessing/match_columns.py:0:0-0:0), using `reindex()` with a `fill_value` that isn't already a category can trigger warnings. Explicitly adding missing columns and then setting values is a more robust pattern for the future.

### Conclusion:
The current implementation in `match_columns` and `match_categories` correctly future-proofs the transformer. It ensures we don't rely on `pd.Categorical`'s default behavior of setting "unseen" values to `-1` (code for `NaN`) without explicit handling. 

I'm happy with this approach as it makes the transformers much more resilient to the upcoming Pandas changes.

PR #915, opened 2026-03-10
feat: Add GroupStandardScaler for scaling variables relative to a giv…

ankitlade12:
## Description

This PR introduces the `GroupStandardScaler` to the `feature_engine.scaling` module. 

Currently, native scalers like `StandardScaler` scale a numerical feature globally across an entire dataset. However, it is an extremely common pattern in data science to scale a feature *relative to its group* (e.g., standardizing `house_price` relative to its `neighborhood`, or scaling a student's `exam_score` relative to their `class_id`).

The `GroupStandardScaler` resolves this by taking both `variables` and `reference` variables (the grouping keys). During `fit`, it learns the mean and standard deviation for each numerical variable per group. During `transform`, it scales the variables using their respective group parameters. It gracefully handles unseen groups during transform by falling back to the global mean and standard deviation.

## Changes:
- Added `GroupStandardScaler` class in `feature_engine/scaling/group_standard.py`.
- Exported `GroupStandardScaler` in `feature_engine/scaling/__init__.py`.
- Included rigorous tests for single-reference scaling, missing values handling, unseen groups fallback, and parameter validation.
- Added full API documentation in `docs/api_doc/scaling/GroupStandardScaler.rst`.
- Added User Guide explanations and examples in `docs/user_guide/scaling/GroupStandardScaler.rst`.

## Examples:
```python
import pandas as pd
from feature_engine.scaling import GroupStandardScaler

df = pd.DataFrame({
    "House_Price": [100000, 150000, 120000, 500000, 550000, 480000],
    "Neighborhood": ["A", "A", "A", "B", "B", "B"]
})

scaler = GroupStandardScaler(
    variables=["House_Price"],
    reference=["Neighborhood"]
)

scaler.fit(df)
df_scaled = scaler.transform(df)
```

## Checklist:
- [x] I have read the contribution guidelines.
- [x] I have tested my code locally.
- [x] I have added documentation for my new feature.
- [x] I have added unit tests for my changes.

### DIFF
--- a/docs/api_doc/scaling/GroupStandardScaler.rst
+++ b/docs/api_doc/scaling/GroupStandardScaler.rst
@@ -1,0 +1,5 @@
+GroupStandardScaler
+===================
+
+.. autoclass:: feature_engine.scaling.GroupStandardScaler
+    :members:
--- a/docs/api_doc/scaling/index.rst
+++ b/docs/api_doc/scaling/index.rst
@@ -10,3 +10,4 @@ given columns
    :maxdepth: 1
 
    MeanNormalizationScaler
+   GroupStandardScaler
--- a/docs/user_guide/scaling/GroupStandardScaler.rst
+++ b/docs/user_guide/scaling/GroupStandardScaler.rst
@@ -1,0 +1,79 @@
+.. _group_standard_scaler:
+
+.. currentmodule:: feature_engine.scaling
+
+GroupStandardScaler
+===================
+
+:class:`GroupStandardScaler()` scales numerical variables relative to a group. It standardises variables by removing the mean and scaling to unit variance per group. This means that for each group within the reference column, the scaler learns the mean and standard deviation of each variable to be scaled. During transformation, it applies the standard z-score formula.
+
+The :class:`GroupStandardScaler()` requires numerical variables to be scaled, and at least one reference variable which acts as the grouping key.
+
+Python example
+--------------
+
+We'll show how to use :class:`GroupStandardScaler()` through a toy dataset. Let's create a toy dataset:
+
+.. code:: python
+
+    import pandas as pd
+    from feature_engine.scaling import GroupStandardScaler
+
+    df = pd.DataFrame({
+        "House_Price": [100000, 150000, 120000, 500000, 550000, 480000],
+        "Neighborhood": ["A", "A", "A", "B", "B", "B"]
+    })
+
+    print(df)
+
+The dataset looks like this:
+
+.. code:: python
+
+       House_Price Neighborhood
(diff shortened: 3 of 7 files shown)


solegalli:
Hi @ankitlade12 

Thanks a lot for this contribution. Would you by any chance have a reference where this type of transformation is applied?

I am not sure it belongs to the scaling module. Scaling is usually done to change the scale without really affecting the overall shape of the variable distribution, but this transformation will indeed change it's shape, so it does not belong in the scaling module.

It's also not a variable stabilizing transformation, so not suitable for the transformation module. I don't really know where it should be placed.

if you could send a few references or share more about how/when/how frequently this is used, maybe we can take it from there?

ankitlade12:
Hi @solegalli, thanks for the thoughtful feedback!

**References / use cases:**

This technique is commonly known as **within-group standardization** (or group-wise z-scoring) and appears frequently across several domains:

- **Econometrics & panel data**: Within-group centering/scaling is a standard preprocessing step in fixed-effects and multilevel/hierarchical models to separate within-group variation from between-group variation. See Gelman & Hill, *Data Analysis Using Regression and Multilevel Models* (Ch. 12–13) for discussion of centering predictors within groups.
- **Education & psychometrics**: Standardizing student scores relative to their school, cohort, or test form is routine practice to make scores comparable across groups (e.g., equating exam difficulty across sessions).
- **Sports analytics**: Player performance metrics are regularly standardized relative to position or league to enable fair cross-group comparisons.
- **Healthcare / clinical trials**: Lab values are often standardized relative to site or demographic group to account for systematic between-group differences before modeling.
- **General ML pipelines**: Any time you have grouped/hierarchical data and want to remove between-group scale differences before feeding features into a model, this is the standard approach. It's the preprocessing counterpart to scikit-learn's `GroupKFold`.

**On module placement:**

You raise a valid point — this transformation does change the marginal distribution shape by removing between-group variation, which makes it different from a global scaler like `StandardScaler`.

That said, the core mechanic is still mean-centering and dividing by standard deviation — it's standardization conditioned on a grouping variable. I'd argue it's closest in spirit to scaling, but I'm open to alternatives. A few options:

1. **Keep it in `scaling`** — the operation *is* standardization, just group-conditional. Users looking for "scaling by group" would naturally look here.
2. **A new submodule** like `feature_engine.group_transforms` — if you anticipate other group-conditional operations (group-wise min-max, group-wise robust scaling, etc.), this could be a clean home.
3. **Place it in `transformation`** — though I agree it's not a stabilizing transformation, so this feels like a weaker fit.

I'm happy to move the class to whichever module you think is best. What's your preference?

PR #919, opened 2026-03-27
[DOC] update categorical encoders user guide

solegalli:


feature_engine/discretisation/geometric_width.py
```diff
@@ -42,29 +42,32 @@
...
-    of intervals.
+    where Max and Min are the variable's maximum and minimum value, and n is the
+    number of intervals.
 
-    The sizes of the intervals themselves are calculated with a geometric progression:
+    The upper boundary of the i-th interval is then calculated as:
 
         .. math::
-            a_{subindex} = a_i cw
+            b_{bin_index} = Min + cw^{bin_index}
```
solegalli, line 70: Hi @glevv 

I am updating the docs for various classes, and in doing so, I noticed that what we had as explanation for how the geometric discretiser works did not match the code output that was produced by the class. So evidently, the documentation is wrong, in the sense, that it does not explain what the class does.

Claude said that I need to update the formula like this to match the code base. Would you agree?

docs/user_guide/discretisation/GeometricWidthDiscretiser.rst
```diff
@@ -6,90 +6,108 @@ GeometricWidthDiscretiser
...
 intervals.
 
-The sizes of the intervals themselves are calculated with a geometric progression:
+The upper boundary of the i-th interval is then calculated as:
+
+.. math::
 
-    .. math::
-        a_{i+1} = a_i cw
+        b_i = Min + cw^i
```
solegalli, line 25: @glevv 

Could you also check whether these changes make sense?

feature_engine/discretisation/geometric_width.py
glevv: He wrote the formula straight out of the code, it is fine. The original formula tried to convey the general idea of geometric progression, the new is closer to the specific implementation

docs/user_guide/discretisation/GeometricWidthDiscretiser.rst
solegalli: awesome, thank you!

docs/user_guide/encoding/StringSimilarityEncoder.rst
```diff
@@ -114,12 +133,12 @@ string similarity to the seen categories.
 No text preprocessing is applied by :class:`StringSimilarityEncoder()`. Be mindful of preparing
 string categorical variables if needed.
 
-:class:`StringSimilarityEncoder()` works with categorical variables by default. And it has the option to
+:class:`StringSimilarityEncoder()` works with categorical variables by default. Tt has the option to
```
glevv: `Tt`
Is this an error?

feature_engine/discretisation/geometric_width.py
glevv: They are both fine IMO, but the Claude version is closer to the code

docs/user_guide/encoding/StringSimilarityEncoder.rst
solegalli: Yes, I'll fix it, thank you!

docs/user_guide/discretisation/GeometricWidthDiscretiser.rst
glevv: It is correct

PR #928, opened 2026-05-31
feat: add drop parameter to OneHotEncoder (#913)

### LINKED ISSUE #913: Add `drop` parameter to OneHotEncoder to control which dummy category is dropped
In the current implementation, when `drop_last=True`, `OneHotEncoder` always drops the last category (alphabetically). This makes it impossible for users to control which category is used as the reference group. In many modeling scenarios (for example logistic regression or other linear models), the choice of the reference category matters and users may want to drop a different category.

Add a `drop` parameter that allows users to control which dummy category is dropped.

```python
drop: str = "last"  # options: "last", "first", "most_frequent"
```

* `"last"` (default): preserves current behaviour — drops the last category alphabetically.
* `"first"`: drops the first category alphabetically.
* `"most_frequent"`: drops the most frequent category found during `fit()`, which can be a more statistically meaningful reference group.

If `drop="most_frequent"` and multiple categories have the same highest frequency, the transformer should raise a `UserWarning` and fall back to dropping the first category found.

The existing `drop_last` parameter should remain for backward compatibility, but a deprecation warning should be raised if it is used together with the new `drop` parameter.

**Describe alternatives you've considered**
Users can currently control the reference category only by manually reordering or preprocessing the categorical values before applying the encoder. However, this is inconvenient and error-prone, especially in larger pipelines.

**Additional context**
Adding this parameter would make `OneHotEncoder` more flexible and align it better with common machine learning workflows, particularly when building statistical or linear models where the reference category is important.

BALOGUN-DAVID:
Closes #913 

### Description
This PR adds a `drop` parameter to the `OneHotEncoder` to allow users to control which dummy category is dropped when performing k-1 encoding. Previously, only the last category in the unique insertion order could be dropped via `drop_last=True`.

**New `drop` strategies:**
- `'last'`: Drops the last category in alphabetical order.
- `'first'`: Drops the first category in alphabetical order.
- `'most_frequent'`: Drops the most frequent category found during `fit()`. In the event of a tie, a `UserWarning` is raised and the first category alphabetically among the ties is dropped.

**Backward Compatibility:**
The `drop_last` parameter is fully preserved for backward compatibility. If both `drop_last` and `drop` are specified, `drop` takes precedence and a `FutureWarning` is emitted.

### Checklist
- [x] Added `drop` parameter and implementation logic
- [x] Added `FutureWarning` when both `drop` and `drop_last` are set
- [x] Updated docstrings for the new parameter
- [x] Added comprehensive tests for all `drop` strategies and tie-breaking
- [x] All tests and estimator checks pass

### DIFF
--- a/feature_engine/encoding/one_hot.py
+++ b/feature_engine/encoding/one_hot.py
@@ -1,6 +1,7 @@
 # Authors: Soledad Galli <solegalli@protonmail.com>
 # License: BSD 3 clause
 
+import warnings
 from typing import List, Optional, Union
 
 import numpy as np
@@ -94,6 +95,18 @@ class OneHotEncoder(CategoricalMethodsMixin, CategoricalInitMixin):
         to `True`, will ensure that for every binary variable in the dataset, only 1
         dummy is created.
 
+    drop: str, default=None
+        Controls which category to drop when creating k-1 dummy variables. Only used
+        if `top_categories` is None. If `drop` is not None and `drop_last` is also
+        True, a `FutureWarning` is raised and `drop` takes precedence.
+
+        - ``None``: No category is dropped (k dummies). Equivalent to ``drop_last=False``.
+        - ``'last'``: Drops the last category in alphabetical order.
+        - ``'first'``: Drops the first category in alphabetical order.
+        - ``'most_frequent'``: Drops the most frequent category found during ``fit()``.
+          If there is a tie, a ``UserWarning`` is raised and the first category
+          alphabetically among the tied categories is dropped.
+
     {variables}
 
     {ignore_format}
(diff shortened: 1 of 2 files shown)


BALOGUN-DAVID:
Hi @solegalli, it's me again! 👋

I've pushed the fixes and the PR is now ready for your review. 

This adds the `drop` parameter to `OneHotEncoder` to allow dropping the `'first'`, `'last'`, or `'most_frequent'` category, while preserving full backward compatibility for `drop_last` (as requested in #913). All tests and estimator checks are passing locally. 

Let me know if you need any adjustments or if this is good to merge!

BALOGUN-DAVID:
@solegalli  can you have a look at this, so it can be merge?

PR #937, opened 2026-07-18
add return_empty attr to base_numerical.py

PastorJordi:
* `fit()` from `BaseNumericalTransformer` now uses internal attribute `return_empty` as argument when an explicit list of variables is not defined (i.e `None`)
(This breaks around 150 tests, which are expected to be fixed during the sprint) 
* added a test

### DIFF
--- a/tests/test_base_transformers/test_base_numerical_transformer.py
+++ b/tests/test_base_transformers/test_base_numerical_transformer.py
@@ -17,7 +17,6 @@ class MockClass(BaseNumericalTransformer):
 
 
 def test_empty_find_numerical_variables(df_vartypes):
-    # without return empty_flag, dfs without numerical variables should raise error 
     transformer = MockClass()
     with pytest.raises(TypeError):
         transformer.fit(df_vartypes.drop(columns=["Age", "Marks"]))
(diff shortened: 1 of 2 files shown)


tests/test_base_transformers/test_base_numerical_transformer.py
```diff
@@ -9,11 +9,25 @@
...
         self.variables = None
+        self.return_empty = False
 
     def transform(self, X):
         return self._check_transform_input_and_state(X)
 
 
+
+def test_empty_find_numerical_variables(df_vartypes):
+    # without return empty_flag, dfs without numerical variables should raise error
```
solegalli:
```suggestion
```

tests/test_base_transformers/test_base_numerical_transformer.py
```diff
@@ -9,11 +9,24 @@
...
     def transform(self, X):
         return self._check_transform_input_and_state(X)
 
 
+
+def test_empty_find_numerical_variables(df_vartypes):
+    transformer = MockClass()
+    with pytest.raises(TypeError):
+        transformer.fit(df_vartypes.drop(columns=["Age", "Marks"]))
+    # Else, no problem
```
solegalli:
```suggestion
```

Issue #420, opened 2022-04-17
expand non_fitted_error_checks for methods get_feature_names_out and inverse transform


solegalli:
Methods transform, inverse_transform, get_feature_names_out and other specific methods (see DropMissingData) should run only after the transformer was fit.

At the moment, we test that this is the case only for the transform method. We should expand this to test all methods.

solegalli:
This was taken care of in #519

Morgan-Sell:
@solegalli,

Is this as simple as adding `check_is_fitted(self)` to the abovementioned methods?

solegalli:
Nothing is as simple as it looks @Morgan-Sell lol

We basically need to ensure that we have what you say in all methods across all transformers, and then make a test, and ensure we test all transformers.

I am thinking about refactoring the way we test. It seems to be readable only to me, lol. I tried to copy what sklearn is doing, but I am not pleased with the outcome. So, this is not a simple one unfort.

Morgan-Sell:
@solegalli, you had me at "refactoring the way we test." Although, I don't know what's wrong with our current approach. How are you thinking we design our tests?

solegalli:
I am entertaining the idea of having individual tests in check_estimators, but instead of then summoning all tests within a master test, which we then summon from the test_estimators.py for all transformers, we should just run them from each transformer test script individually. 

Hard to explain without going over the scripts. But I'd like it to be clear for each user which tests are being run for each transformer, and I would like to avoid adding tags as we do now, as much as possible

TemidayoA:
Opened #953 covering the inverse_transform half of this — the shared check now calls inverse_transform on an unfitted clone (accepting NotImplementedError for the transformers that deliberately don't support inversion). All transformers get covered automatically through check_feature_engine_estimator. Happy to adjust if you'd prefer this shaped differently given the test-refactor plans discussed above.

khan-sudais:
Update: after inspecting the current main-branch history, I found that PR #969 already implemented this request and its merge commit explicitly closes #420. I won’t duplicate the work. Thanks!

PR #953, opened 2026-07-19
Extend non-fitted error check to cover inverse_transform

TemidayoA:
## What this does

Part of #420 (the `inverse_transform` half — `get_feature_names_out` checks were added in #519).

The shared estimator check `check_raises_non_fitted_error` only verified that `transform()` raises `NotFittedError` before `fit()`. Every transformer that implements `inverse_transform` already guards it (via `_check_transform_input_and_state` / `check_is_fitted`), but no test covered that behaviour, so a regression would go unnoticed.

This extends the shared check to also call `inverse_transform` on an unfitted clone. Transformers that deliberately don't support inversion (`OneHotEncoder`, `StringSimilarityEncoder`) raise `NotImplementedError`, which the check accepts.

Because the check runs through `check_feature_engine_estimator`, all transformers across all modules are covered automatically — no per-transformer test changes needed.

## Verification

- Full test suite: 2001 passed
- Sanity check that the extended check bites: a dummy transformer with an unguarded `inverse_transform` fails it; real transformers pass
- `flake8` clean on the changed file

### DIFF
--- a/tests/estimator_checks/non_fitted_error_checks.py
+++ b/tests/estimator_checks/non_fitted_error_checks.py
@@ -9,13 +9,22 @@ from tests.estimator_checks.dataframe_for_checks import test_df
 
 def check_raises_non_fitted_error(estimator):
     """
-    Check if transformer raises error when transform() method is called before
-    calling fit() method.
+    Check if transformer raises error when transform() or inverse_transform()
+    methods are called before calling fit() method.
 
     The functionality is provided by sklearn's `check_is_fitted` function.
+
+    Transformers that do not support inverse_transform raise NotImplementedError
+    instead, regardless of whether the transformer was fitted.
     """
     X, y = test_df()
     transformer = clone(estimator)
     # Test when fit is not called prior to transform.
     with pytest.raises(NotFittedError):
         transformer.transform(X)
+
+    # Test when fit is not called prior to inverse_transform.
+    transformer = clone(estimator)
+    if hasattr(transformer, "inverse_transform"):
+        with pytest.raises((NotFittedError, NotImplementedError)):
+            transformer.inverse_transform(X)
--- a/tests/estimator_checks/non_fitted_error_checks.py
+++ b/tests/estimator_checks/non_fitted_error_checks.py
@@ -12,19 +12,58 @@ def check_raises_non_fitted_error(estimator):
     Check if transformer raises error when transform() or inverse_transform()
     methods are called before calling fit() method.
 
-    The functionality is provided by sklearn's `check_is_fitted` function.
+    For transform(), and for inverse_transform() when the transformer implements
+    it, the expected error is NotFittedError, provided by sklearn's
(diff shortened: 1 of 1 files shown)


TemidayoA:
Note on the CI failures on the first run, for whoever reviews:

- `test_feature_engine_py39`: all 51 failures are `urllib.error.HTTPError: HTTP Error 403: Forbidden` in the `test_sklearn_wrapper` dataset-download tests — a network flake on the runner, unrelated to this PR's diff (the non-fitted checks themselves passed; 1950 tests passed in the same job). Pushed an empty commit to retrigger.
- `test_type`: also failing on `main`'s latest pipeline (build 28828), and this PR only touches `tests/`, which mypy doesn't check — pre-existing.
- `upload_codecov`: downstream of the failed job.

tests/estimator_checks/non_fitted_error_checks.py
```diff
@@ -9,13 +9,22 @@
...
     X, y = test_df()
     transformer = clone(estimator)
     # Test when fit is not called prior to transform.
     with pytest.raises(NotFittedError):
         transformer.transform(X)
+
+    # Test when fit is not called prior to inverse_transform.
+    transformer = clone(estimator)
+    if hasattr(transformer, "inverse_transform"):
+        with pytest.raises((NotFittedError, NotImplementedError)):
```
solegalli: Thanks for the PR @TemidayoA 

Failing on both NotFittedError and NotImplementedError is a bit dirty, because we won't know if it fails for one or the other, so it won't detect changes that we do looking forward. If there are transformers that have the attribute inverse_transform and it's not implemented, we need to fail specifically for those with the NotImplementedError and for the rest with the NotFittedError

TemidayoA:
Thanks @solegalli, that's a fair point — the tuple would have hidden exactly the regressions this check exists to catch. Fixed in the latest commit.

The check now determines, per transformer, which error it *should* raise and asserts only that one:

- `inverse_transform` is treated as **not implemented** when a fitted clone raises `NotImplementedError` (OneHotEncoder, RareLabelEncoder, StringSimilarityEncoder) → asserts `NotImplementedError`.
- every other transformer → asserts `NotFittedError`.

I detect implementation status on a *fitted* clone rather than a hardcoded list, so it stays correct for the conditional cases too (e.g. `SklearnTransformerWrapper`, which checks `fit` first and so raises `NotFittedError` before fit regardless of the wrapped transformer, and `MeanEncoder` whose default config is implemented).

I confirmed it's now strict: a transformer whose `inverse_transform` is implemented but drops its fit-guard (raising `NotImplementedError` before fit) now fails the check, where the tuple passed it. Full suite: 2001 passed, flake8 clean.

The classification fits on `test_df(categorical=True, datetime=True)` so every transformer that reaches this check can be fitted — the same input the neighbouring `get_feature_names_out` check already uses.

solegalli:
Thank you @TemidayoA for the quick update!

PR #964, opened 2026-07-21
[MNT] add standalone performance benchmark suite with pre-2.0 baseline

solegalli:
Measures fit() and transform() wall-clock time for 40 representative transformers across all modules, on synthetic data of configurable size. Supports saving a baseline and comparing later runs against it, failing when a benchmark regresses beyond a threshold. Not collected by pytest and not run in CI; see benchmarks/README.md.

Baseline recorded before the 2.0 refactoring (polars support, numpy compute) as the reference to verify that the refactoring speeds up the transformers rather than slowing them down.


solegalli:
This should have been closed in #984

PR #965, opened 2026-07-21
[MNT] narwhals migration

solegalli:


ojassharma7:
Hi @solegalli — I'd like to help with the narwhals migration.

If it is still free, I can take **`feature_engine/scaling`** first (small surface: mainly `MeanNormalisationScaler`) as a single-module PR, following the dataframe_checks pattern from #966.

Please let me know if that module is already spoken for — happy to pick another (e.g. a simpler preprocessing piece) instead.

solegalli:
That is actually a good one to start with. The tests should pass with pandas. I am not sure they will pass with polars because we need to change the functions that select variables, on which I am working on right now and will soon make a PR.

ojassharma7:
Started on **scaling** as discussed — opened a PR against this branch: will link here once created (see latest open PR from @ojassharma7 titled migrate scaling module to narwhals).

Pandas tests for the module pass locally. As you said, polars may still need your variable-selection updates.

ojassharma7:
Scaling PR: https://github.com/feature-engine/feature_engine/pull/979

feature_engine/dataframe_checks.py
```diff
@@ -2,324 +2,241 @@
...
-    with Scikit-learn transformers like the `SimpleImputer`, which return by default
-    Numpy arrays.
+    Checks that X is a dataframe from any library supported by narwhals (for example
+    pandas, polars, modin, cuDF, or PyArrow).
 
     Parameters
     ----------
-    X : pandas Dataframe or numpy array.
-        The input to check and copy or transform.
+    X : dataframe (pandas, polars, or any other library supported by narwhals).
```
FBruzzesi: I would add a mention to "eager". We support a bunch more dataframe libraries that are lazy only

feature_engine/dataframe_checks.py
```diff
@@ -2,324 +2,241 @@
...
-import pandas as pd
-from scipy.sparse import issparse
+from narwhals.typing import IntoDataFrame, IntoSeries
 from sklearn.utils.validation import _check_y, check_consistent_length, column_or_1d
 
-from feature_engine.variable_handling._variable_type_checks import is_object
 
-
-def check_X(X: Union[np.generic, np.ndarray, pd.DataFrame]) -> pd.DataFrame:
+def check_X(X: IntoDataFrame):
```
FBruzzesi: As input/output is the same underlying dataframe type, you could do:

```suggestion
def check_X(X: IntoDataFrameT) -> IntoDataFrameT:
```

(importable from the typing module as `IntoDataFrame`, `IntoSeries`)

feature_engine/dataframe_checks.py
```diff
@@ -2,324 +2,241 @@
...
             "requires y to be passed, but the target y is None",
             "Expected array-like (array or non-string sequence), got None",
             "y should be a 1d array",
         )
 
-    elif isinstance(y, pd.Series):
-        if y.isnull().any():
+    if nwd.is_into_series(y):
+        nw_y = nw.from_native(y, series_only=True)
+        if nw_y.is_null().any():
```
FBruzzesi: Notice that null and nan's are not the same necessarily the same (see [Null/NaN handling
](https://narwhals-dev.github.io/narwhals/concepts/null_handling/)). We also support [`Series.is_nan`](https://narwhals-dev.github.io/narwhals/api-reference/series/#narwhals.series.Series.is_nan) but only for numeric types:


```suggestion
        if nw_y.is_null().any() or (nw_y.dtype.is_numeric() and nw_y.is_nan().any():
```

feature_engine/dataframe_checks.py
```diff
@@ -2,324 +2,241 @@
...
-        if not is_object(y) and not np.isfinite(y).all():
-            raise ValueError("y contains infinity values.")
-        if y_numeric and is_object(y):
-            y = y.astype("float64")
-        y = y.copy()
-
-    elif isinstance(y, pd.DataFrame):
-        if y.isnull().any().any():
+        if nw_y.dtype.is_numeric():
+            if not np.isfinite(nw_y.to_numpy()).all():
```
FBruzzesi, line 99:
```suggestion
            if not nw_y.is_finite().all():
```

(see [`Series.is_finite()`](https://narwhals-dev.github.io/narwhals/api-reference/series/#narwhals.series.Series.is_finite))

feature_engine/dataframe_checks.py
```diff
@@ -2,324 +2,241 @@
...
+                raise ValueError("y contains infinity values.")
+        elif y_numeric:
+            nw_y = nw_y.cast(nw.Float64())
+        return nw_y.to_native()
+
+    if nwd.is_into_dataframe(y):
+        nw_y = nw.from_native(y, eager_only=True)
+        if nw_y.select(nw.all().is_null().any()).to_numpy().any():
```
FBruzzesi: You can avoid casting to numpy:


```suggestion
        if nw_y.select(nw.any_horizontal(nw.all().is_null().any())).item():
```

feature_engine/dataframe_checks.py
```diff
@@ -2,324 +2,241 @@
...
+        "Some of the variables in the dataset contain NaN. Check and "
+        "remove those before using this transformer."
+    )
+    error_msg_ignore = (
+        "Some of the variables in the dataset contain NaN. Check and "
+        "remove those before using this transformer or set the parameter "
+        "`missing_values='ignore'` when initialising this transformer."
+    )
+    nw_X = nw.from_native(X, eager_only=True)
+    if nw_X.select(nw.col(variables).is_null().any()).to_numpy().any():
```
FBruzzesi: Similar as above, you can use any_horizontal

FBruzzesi (commented):
Hi @solegalli - Following up on my [discord comment](https://discord.com/channels/1235257048170762310/1237430779584446546/1541054832155299890). I added a few comments, mostly focusing on the changes in `feature_engine/dataframe_checks.py` - I hope you find them helpful

I noticed that a lot of tests were refactored as well: if you want to test the same behavior for many dataframes, I would reference what I did for fairlearn (see their [conftest](https://github.com/fairlearn/fairlearn/blob/main/test/conftest.py) file), namely create fixture dataframe constructor for all the dataframe types you want to test. Ideally I would like to move that into narwhals as well (see https://github.com/narwhals-dev/narwhals/pull/3552), but that's still work-in-progress and under discussion 🙏🏼

feature_engine/dataframe_checks.py
solegalli: Hi @FBruzzesi , thanks for the suggestion. It seems that using numpy is faster than using narwhals both for pandas and polars (mostly so for pandas). Is this a known issue?

FBruzzesi: * For the polars case we run its native functionality [`polars.Series.is_finite`](https://docs.pola.rs/api/python/stable/reference/series/api/polars.Series.is_finite.html#polars.Series.is_finite). I am surprised that's faster than numpy, at least at scale
* For pandas-like, we do `(s > float("-inf")) & (s < float("inf"))`. IIRC that's to avoid using numpy with non-numpy backed series (e.g. pyarrow backed series, cudf series that live in the GPU, etc). If the delta is large at scale, we can take a look for a refactor with performance in mind.

For context: in general we tend to use the native dataframe libraries API/functionalities. pandas is a special kid as we need to do quite some gymnastic for null vs nan's, its datatype system, its multiple backends, etc..

So please keep reporting these kind of performance issues - we aim to keep overhead at the minimum

solegalli: Thanks for replying so quickly. These are the values I've got (on pandas and polars, 200k rows × 20 cols):

Check	                        pandas	                                polars
null check (multi-col)	narwhals-native 1.2x slower	narwhals-native 4x slower
inf check (multi-col)	narwhals-native 2.4x slower	narwhals-native 1.3x slower
is_finite (single series)	narwhals-native 10x slower	~same

is_finite is the same for polars, the inf and null checks make it a bit slower respect to numpy.

FBruzzesi: For pandas I just opened a PR to use numpy/cupy/pyarrow.compute native functionalities directly: see https://github.com/narwhals-dev/narwhals/pull/3874

For polars, I cannot tell why numpy is faster than their native implementation - If interested, you can double check with them either in discord or in their repo

solegalli: thank you!

Issue #958, opened 2026-07-20
unify class names to british spelling


solegalli:
At the moment we have some classes with the british spelling ie discretisers and some with the american spelling ie winsorizer and mean normalizer.

I would like to unify all to british spelling, but to keep backward compatibility we could have the classes with both spellings, we just show the british one in the docs, but keep the other one functionning as well, so users can use any, at least for the few next releases.

British spelling
- [x] Winsorizer should be with s winsoriser
- [x] MeanNormalizationScaler should be with s in normalisation

favourone:
Hi @solegalli , I'd like to work on this issue.

My plan is to add the British-spelled public names `Winsoriser` and
`MeanNormalisationScaler`, keep the existing American-spelled names
backward compatible, update the exports and documentation, and add tests
confirming that both spellings behave identically.

Would you prefer both changes in one PR or separate PRs?

solegalli:
Hi @favourone 

Thanks for the support! it would be great if you could make 1 PR per class :)

I look forward to your contributions !

favourone:
> 你好[@favourone](https://github.com/favourone)
> 
> 感谢支持！如果每个班级都能提交一个 PR 就太好了 :)
> 
> 期待您的贡献！

Thanks! I'll start with `Winsoriser` and submit it as the first PR. Then I'll work on `MeanNormalisationScaler` in a separate branch and PR.

solegalli:
resolved in #967 and #968

PR #968, opened 2026-07-22
[MNT] rename MeanNormalisationScaler public name to british spelling

favourone:
## Summary

- Add `MeanNormalisationScaler` as the preferred public class name.
- Retain `MeanNormalizationScaler` as a backward-compatible alias.
- Export both names from `feature_engine.scaling`.
- Update current API documentation, user guide, and README.
- Preserve existing module and documentation paths for compatibility.
- Add an explicit identity test for the compatibility alias.

## Tests

- `python -m pytest tests/test_scaling/test_mean_normalization.py -q` — 9 passed
- `python -m pytest tests/test_scaling -q` — 9 passed
- `flake8` — passed
- `black --check` — passed
- `git diff --check` — passed

Part of #958.


favourone:
The two required CircleCI jobs failed because the California housing dataset download returned HTTP 403 Forbidden. The other 11 checks passed. I do not have permission to rerun the workflow. Could a maintainer please rerun it?

docs/user_guide/scaling/MeanNormalizationScaler.rst
```diff
@@ -28,23 +28,25 @@ Mean normalisation is given by the following formula:
...
-:class:`MeanNormalizationScaler()` scales variables using mean normalisation.
+:class:`MeanNormalisationScaler()` scales variables using mean normalisation.
 
 .. note::
 
-    :class:`MeanNormalizationScaler()` only works with non-constant numerical variables.
+    :class:`MeanNormalisationScaler()` only works with non-constant numerical variables.
     If the variable is constant, the scaler will raise an error.
 
+    ``MeanNormalizationScaler`` remains available as a backward-compatible alias.
```
solegalli: Could we remove this line from the user docs?

docs/index.rst
```diff
@@ -328,7 +328,7 @@ procedures to a subset of the variables only, check out the :doc:`api_doc/wrappe
 
 Feature-engine extends scikit-learn's scaling functionality with the following transformer:
 
-- :doc:`api_doc/scaling/MeanNormalizationScaler`: scale variables using mean normalisation
+- :doc:`MeanNormalisationScaler <api_doc/scaling/MeanNormalizationScaler>`: scale variables using mean normalisation
```
solegalli:
```suggestion
- :doc:`<api_doc/scaling/MeanNormalizationScaler>`: scale variables using mean normalisation
```

feature_engine/scaling/mean_normalization.py
```diff
@@ -51,6 +51,8 @@ class MeanNormalizationScaler(BaseNumericalTransformer):
 
     More details in the :ref:`User Guide <mean_normalisation_scaler>`.
 
+    The ``MeanNormalizationScaler`` spelling remains available as a
```
solegalli: I'd remove this from the documentation. If people have legacy code, they'll realise there's been a change with a warning that should raise the class with the american spelling

feature_engine/scaling/mean_normalization.py
```diff
@@ -189,3 +191,7 @@ def inverse_transform(self, X: pd.DataFrame) -> pd.DataFrame:
         X[self.variables_] = X[self.variables_] * self.range_ + self.mean_
 
         return X
+
+
+# Backward-compatible alias for the original American spelling.
+MeanNormalizationScaler = MeanNormalisationScaler
```
solegalli: Is there a way to trigger a warning when the user calls the class with z instead of s? I though that if instead of equal we used inheritance, then we could add the warning in the init or something like that. Could you research this possibility? Something like MeanNormalizationScaler was deprecated in favour of MeanNormalisationScaler in version 2.0.0 and will be removed in version 2.1.0. To silence the warning, use MeanNormalisationScaler instead.

tests/test_scaling/test_mean_normalization.py
```diff
@@ -4,10 +4,15 @@
 import pytest
 from sklearn.exceptions import NotFittedError
 
-from feature_engine.scaling import MeanNormalizationScaler
+from feature_engine.scaling import MeanNormalisationScaler, MeanNormalizationScaler
 from tests.estimator_checks.fit_functionality_checks import check_return_empty
 
 
+def test_mean_normalization_scaler_is_backward_compatible_alias():
```
solegalli: is this the best way to assess that both classes have the same behaviour? Somehow I thought that we could pass the different classes to all tests using parametrize. We should also test that using the old class raises the future warning.

solegalli (commented):
Hey @favourone 

Great job working on this fix. Well done. 

Could you explore the possibility of adding a future warning to the old version of the class, so that when a user (or an LLM) use the old class in their projects, they see the warning and revert to the new way of calling it? I added more details in the comments.

I think we should raise this warning, instead of pointing to the backward compatible class in the user guide or docstrings, which we would remove anyways in future releases.

solegalli:
> The two required CircleCI jobs failed because the California housing dataset download returned HTTP 403 Forbidden. The other 11 checks passed. I do not have permission to rerun the workflow. Could a maintainer please rerun it?

I am aware of this issue. I need to fix it :)

solegalli (commented):
Hi @favourone 

This is good to merge. great work. 

The only thing left is to rename the files in the user guide and api folder to **MeanNormalisationScaler.rst‎** and then update the toctrees that reference them. it sounds minor but this updates the url once we build the documentation. 

Thanks a lot!

feature_engine/scaling/mean_normalization.py
```diff
@@ -189,3 +189,20 @@ def inverse_transform(self, X: pd.DataFrame) -> pd.DataFrame:
         X[self.variables_] = X[self.variables_] * self.range_ + self.mean_
 
         return X
+
+
```
solegalli, line 193:
```suggestion
# TODO: remove in version 2.1.0
```

PR #969, opened 2026-07-23
expand non_fitted_error_checks for methods get_feature_names_out and inverse transform

### LINKED ISSUE #420: expand non_fitted_error_checks for methods get_feature_names_out and inverse transform

ojassharma7:
Fixes #420.

`check_raises_non_fitted_error` only verified that `transform()` and `inverse_transform()` raise `NotFittedError` when called before `fit()`. This extends it to the other public methods that also require a fitted transformer:

- `get_feature_names_out()`
- `transform_x_y()` and `return_na_data()` — guarded with `hasattr`, so they are only exercised on the transformers that implement them (e.g. `DropMissingData`), which the issue pointed to specifically.

The docstring is updated to list the methods now covered. All tests pass locally.

### DIFF
--- a/tests/estimator_checks/non_fitted_error_checks.py
+++ b/tests/estimator_checks/non_fitted_error_checks.py
@@ -9,18 +9,23 @@ from tests.estimator_checks.dataframe_for_checks import test_df
 
 def check_raises_non_fitted_error(estimator):
     """
-    Check if transformer raises error when transform() or inverse_transform()
-    methods are called before calling fit() method.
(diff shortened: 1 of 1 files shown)


solegalli:
Looks great! Thank you @ojassharma7 

We look forward to more of your contributions :)

PR #970, opened 2026-07-24
[MNT] Deprecate LogCpTransformer in favour of LogTransformer with C parameter

### LINKED ISSUE #957: [MNT] join LogTransformers into 1 class
At the moment we have 2 transformers that apply the logarithm transformation, 1 that does just that, and the second one that can add a constant. I think it makes sense to merge them into 1 class, the original one, which would now support adding a constant. While we are at it, I would like to explore options to have the class not fail when the variable has 0 and the user has not declared the addition of a constant. Not sure what would be possible here.

adityaanikam:
Closes #957

## Summary

Merges `LogCpTransformer` into `LogTransformer` as approved in the issue
discussion. `LogTransformer` gains a `C` parameter (`"auto"` / int / float /
dict) mirroring `LogCpTransformer`'s exact semantics, defaulting to `C=0` so
existing `LogTransformer()` usage is unchanged. `LogCpTransformer` becomes a
thin subclass defaulting to `C="auto"`, noted as being consolidated into
`LogTransformer` in favor of `LogTransformer(C=...)`.

## Backward compatibility

- `LogTransformer()`'s behavior and error message/timing are unchanged:
  fit-time validation still fires exactly when `C` resolves to `0` (the new
  default), with the original message.
- `LogCpTransformer`'s numerical behavior and `C_` computation are unchanged.

## Things worth flagging

1. `LogCpTransformer`'s `base=` error message lost its dynamic
   `Got {base} instead.` suffix — it now uses `LogTransformer`'s original
   plain message, since that's the class whose wording was promised
   unchanged. Updated the corresponding test to match.
2. `LogCpTransformer` gets its own `_more_tags()`/`__sklearn_tags__()`
   overrides resetting to the un-xfailed tags, so it doesn't silently
   inherit `LogTransformer`'s zero-value xfail exemptions that it never
   needed (its `"auto"` default always finds a valid shift).
3. There's no existing convention in this codebase for deprecating a class
   (searched — no precedent), so I kept the notice to a plain docstring
   `.. note::` rather than inventing a formal `.. deprecated::` version
   directive. I also didn't touch `docs/user_guide/transformation/
   LogCpTransformer.rst`, which still describes it as fully independent —
   happy to update that in this PR or a follow-up, whichever you'd prefer.

The 0-handling-without-a-declared-constant question from the issue is left
for a follow-up, per the discussion.

## Testing

`pytest tests/test_transformation/` — 117 passed, including a new regression
test proving the `C=0` default preserves the original fail-fast contract.

### DIFF
--- a/docs/user_guide/transformation/LogTransformer.rst
+++ b/docs/user_guide/transformation/LogTransformer.rst
@@ -28,9 +28,10 @@ LogTransformer
 
 .. note::
 
-    Note that the logarithm can only be applied to positive values. Thus, if the variable contains 0 or negative values, this transformer will return an error.
+    By default (``C=0``), the logarithm can only be applied to positive values. If the variable contains 0 or negative values, the transformer will return an error, unless you use the ``C`` parameter described below.
 
-To transform non-positive variables you can add a constant to shift the data points towards positive values. You can do this by using :class:`LogCpTransformer()`.
+To transform non-positive variables, pass a value to the ``C`` parameter to shift
+the data points towards positive values, as shown in the section below.
 
 .. attention::
 
@@ -159,6 +160,64 @@ In the following plots we see histograms showing the variables in their original
 .. tip::
 
     Following the transformations with scatter plots and residual analysis of the regression models helps understand if the transformations are useful in our regression analysis.
+
+
+Transforming non-positive variables with C
+------------------------------------------
+
+:class:`LogTransformer()` can also transform variables that contain zero or
+negative values, by adding a constant ``C`` before applying the logarithm,
+i.e. ``log(x + C)``.
+
+Let's load the diabetes dataset that comes with scikit-learn, which contains
+variables with negative values:
+
+.. code:: python
+
+    from sklearn.model_selection import train_test_split
+    from sklearn.datasets import load_diabetes
+    from feature_engine.transformation import LogTransformer
+
+    # Load dataset
+    X, y = load_diabetes(return_X_y=True, as_frame=True)
+
+    # Separate into train and test sets
+    X_train, X_test, y_train, y_test = train_test_split(
+        X, y, test_size=0.3, random_state=0)
+
+Let's set up :class:`LogTransformer()` with ``C="auto"``, so it automatically
+finds the constant needed to shift each variable's distribution to positive
+values before applying the logarithm:
+
+.. code:: python
+
+    logt = LogTransformer(variables=["bmi", "s3"], C="auto")
+    logt.fit(X_train)
+
+We can inspect the constant values that will be added to each variable in the
+``C_`` attribute:
+
+.. code:: python
+
+    logt.C_
+
+Since these variables are not strictly positive, the transformer found the
+minimum value needed to make their values positive:
+
+.. code:: python
+
+    {'bmi': 1.0848862355291056, 's3': 1.102307050517416}
+
+We can now transform the data:
+
+.. code:: python
+
+    train_t = logt.transform(X_train)
+    test_t = logt.transform(X_test)
+
+You can also pass a fixed constant to all variables (``C=5``), or a dictionary
+mapping each variable to its own constant (``C={"bmi": 2, "s3": 3}``), the same
+way you would with the deprecated :class:`LogCpTransformer()`.
 
 
 Additional resources
--- a/docs/user_guide/transformation/LogTransformer.rst
+++ b/docs/user_guide/transformation/LogTransformer.rst
@@ -28,10 +28,7 @@ LogTransformer
 
 .. note::
 
-    By default (``C=0``), the logarithm can only be applied to positive values. If the variable contains 0 or negative values, the transformer will return an error, unless you use the ``C`` parameter described below.
-
-To transform non-positive variables, pass a value to the ``C`` parameter to shift
-the data points towards positive values, as shown in the section below.
+    The logarithm can only be applied to positive values; if the variable contains 0 or negative values, the transformer will raise an error. To avoid this, add a constant to the variables by using the `C` parameter (more details below).
 
 .. attention::
 
@@ -163,7 +160,7 @@ In the following plots we see histograms showing the variables in their original
 
 
 Transforming non-positive variables with C
-------------------------------------------
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 :class:`LogTransformer()` can also transform variables that contain zero or
 negative values, by adding a constant ``C`` before applying the logarithm,
@@ -174,8 +171,8 @@ variables with negative values:
 
 .. code:: python
 
-    from sklearn.model_selection import train_test_split
     from sklearn.datasets import load_diabetes
+    from sklearn.model_selection import train_test_split
     from feature_engine.transformation import LogTransformer
 
     # Load dataset
--- a/feature_engine/transformation/log.py
+++ b/feature_engine/transformation/log.py
@@ -143,7 +143,8 @@ class LogTransformer(BaseNumericalTransformer, FitFromDictMixin):
 
         if not isinstance(C, (int, float, dict)) and C != "auto":
             raise ValueError(
-                f"C can take only 'auto', integers or floats. Got {C} instead."
+                "C can take only 'auto', integers, floats or dictionaries. "
+                f"Got {C} instead."
             )
 
         _check_return_empty_is_bool(return_empty)
--- a/feature_engine/transformation/log.py
+++ b/feature_engine/transformation/log.py
@@ -42,13 +42,18 @@ from feature_engine.tags import _return_tags
     fit_transform=_fit_transform_docstring,
     inverse_transform=_inverse_transform_docstring,
 )
-class LogTransformer(BaseNumericalTransformer):
+class LogTransformer(BaseNumericalTransformer, FitFromDictMixin):
     """
     The LogTransformer() applies the natural logarithm or the base 10 logarithm to
-    numerical variables. The natural logarithm is the logarithm in base e.
+    numerical variables, optionally after adding a constant C, i.e., log(x + C).
 
-    The LogTransformer() only works with positive values. If the variable
-    contains a zero or a negative value the transformer will return an error.
+    By default, C=0, so LogTransformer() only works with positive values and
+    behaves exactly as it always has: if a variable contains a zero or a negative
+    value, the transformer raises an error.
+
+    To transform variables that contain zero or negative values, pass a non-zero
+    C: either an explicit constant, "auto" to let the transformer determine a
+    shift per variable, or a dictionary mapping each variable to its own constant.
 
     A list of variables can be passed as an argument. Alternatively, the transformer
     will automatically select and transform all variables of type numeric.
@@ -65,9 +70,25 @@ class LogTransformer(BaseNumericalTransformer):
         Indicates if the natural or base 10 logarithm should be applied. Can take
         values 'e' or '10'.
 
+    C: "auto", int, float or dict, default=0
+        The constant C to add to the variable before the logarithm, i.e., log(x + C).
+
+        - If 0 (the default), no constant is added and the variable must be
+          strictly positive, matching the transformer's original behavior.
+        - If int or float, then log(x + C).
+        - If "auto", then C = abs(min(x)) + 1.
+        - If dict, dictionary mapping the constant C to apply to each variable.
+
+        Note, when C is a dictionary, the parameter `variables` is ignored.
+
     Attributes
     ----------
     {variables_}
+
+    C_:
+        The constant C added to each variable. Equal to C, unless C = "auto", in
+        which case it is a dictionary with C = abs(min(variable)) + 1. For strictly
+        positive variables, C = 0.
 
     {feature_names_in_}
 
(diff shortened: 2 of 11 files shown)


feature_engine/transformation/log.py
```diff
@@ -65,10 +70,26 @@ class LogTransformer(BaseNumericalTransformer):
         Indicates if the natural or base 10 logarithm should be applied. Can take
         values 'e' or '10'.
 
+    C: "auto", int, float or dict, default=0
+        The constant C to add to the variable before the logarithm, i.e., log(x + C).
+
+        - If 0 (the default), no constant is added and the variable must be
+          strictly positive, matching the transformer's original behavior.
```
solegalli:
```suggestion
          strictly positive.
```

feature_engine/transformation/log.py
```diff
@@ -42,13 +42,18 @@
...
+class LogTransformer(BaseNumericalTransformer, FitFromDictMixin):
     """
     The LogTransformer() applies the natural logarithm or the base 10 logarithm to
-    numerical variables. The natural logarithm is the logarithm in base e.
+    numerical variables, optionally after adding a constant C, i.e., log(x + C).
 
-    The LogTransformer() only works with positive values. If the variable
-    contains a zero or a negative value the transformer will return an error.
+    By default, C=0, so LogTransformer() only works with positive values and
+    behaves exactly as it always has: if a variable contains a zero or a negative
```
solegalli:
```suggestion
    If a variable contains a zero or a negative
```

feature_engine/transformation/log.py
```diff
@@ -42,13 +42,18 @@
...
-class LogTransformer(BaseNumericalTransformer):
+class LogTransformer(BaseNumericalTransformer, FitFromDictMixin):
     """
     The LogTransformer() applies the natural logarithm or the base 10 logarithm to
-    numerical variables. The natural logarithm is the logarithm in base e.
+    numerical variables, optionally after adding a constant C, i.e., log(x + C).
 
-    The LogTransformer() only works with positive values. If the variable
-    contains a zero or a negative value the transformer will return an error.
+    By default, C=0, so LogTransformer() only works with positive values and
```
solegalli:
```suggestion
    By default, C=0, so LogTransformer() only works with positive values.
```

feature_engine/transformation/log.py
```diff
@@ -42,13 +42,18 @@
...
+    behaves exactly as it always has: if a variable contains a zero or a negative
+    value, the transformer raises an error.
```
solegalli: Could we add a note saying that C will be changed from 0 to "auto" in version 2.1.0? I think this is the easiest way to avoid the error on negative values that we were discussing earlier.

feature_engine/transformation/log.py
```diff
@@ -65,10 +70,26 @@ class LogTransformer(BaseNumericalTransformer):
...
+        - If dict, dictionary mapping the constant C to apply to each variable.
+
+        Note, when C is a dictionary, the parameter `variables` is ignored.
+
     Attributes
     ----------
     {variables_}
 
+    C_:
+        The constant C added to each variable. Equal to C, unless C = "auto", in
```
solegalli:
```suggestion
        The constant C added to each variable. Equal to `C`, unless `C = "auto"`, in
```

feature_engine/transformation/log.py
```diff
@@ -109,24 +130,33 @@ def __init__(
...
         base: str = "e",
+        C: Union[int, float, str, Dict[Union[str, int], Union[float, int]]] = 0,
     ) -> None:
 
         if base not in ["e", "10"]:
             raise ValueError("base can take only '10' or 'e' as values")
 
+        if not isinstance(C, (int, float, dict)) and C != "auto":
+            raise ValueError(
+                f"C can take only 'auto', integers or floats. Got {C} instead."
```
solegalli: it can also take dictionaries.

feature_engine/transformation/log.py
```diff
@@ -109,24 +130,33 @@ def __init__(
...
-        This transformer does not learn parameters.
+        Learn the constant C to add to the variable before the logarithm
+        transformation, if C="auto". Otherwise, this transformer does not learn
+        parameters.
 
-        Selects the numerical variables and determines whether the logarithm
-        can be applied on the selected variables, i.e., it checks that the variables
-        are positive.
+        Selects the numerical variables and, when C=0 (the default), determines
+        whether the logarithm can be applied on the selected variables, i.e., it
```
solegalli: I think we should remove lines 157-159.

feature_engine/transformation/log.py
```diff
@@ -231,80 +286,17 @@ def __sklearn_tags__(self):
...
-    Attributes
-    ----------
-    {variables_}
+    variable to transform and C is a positive constant.
 
-    C_:
-        The constant C to add to each variable. If C = "auto" a dictionary with
-        C = abs(min(variable)) + 1. For strictly positive variables, C = 0.
+    .. note::
+        `LogCpTransformer` is being consolidated into `LogTransformer`. New
```
solegalli: Could we add the version changes and future deprecations in this note? Something like:

`LogCpTransformer` is consolidated into `LogTransformer` in version 2.0.0 and will be deprecated in version 2.1.0. New code....

feature_engine/transformation/log.py
```diff
@@ -231,80 +286,17 @@ def __sklearn_tags__(self):
...
-    variables_=_variables_attribute_docstring,
-    feature_names_in_=_feature_names_in_docstring,
-    n_features_in_=_n_features_in_docstring,
-    fit_transform=_fit_transform_docstring,
-    inverse_transform=_inverse_transform_docstring,
-)
-class LogCpTransformer(BaseNumericalTransformer, FitFromDictMixin):
+class LogCpTransformer(LogTransformer):
     """
     LogCpTransformer() applies the transformation log(x + C), where x is the
```
solegalli, line 294:
```suggestion
    #TODO: remove in version 2.1.0
    LogCpTransformer() applies the transformation log(x + C), where x is the
```

feature_engine/transformation/log.py
```diff
@@ -335,124 +327,16 @@ def __init__(
...
-        # check input dataframe and if class was fitted
-        X = self._check_transform_input_and_state(X)
-
-        # check variable is positive after adding c
-        error_msg = (
-            "Some variables contain zero or negative values after adding"
-            + " constant C, can't apply log."
+        super().__init__(
+            variables=variables, return_empty=return_empty, base=base, C=C
         )
```
solegalli, line 343: Could we add a future warning in the init saying that LogCP... was deprecated in favour of LogTransformer in version 2.0.0 and will be removed in version 2.1.0. Use LogTransformer() instead? So every time someone uses this class, they are alerted that it will disappear in the next version?

tests/test_transformation/test_log_transformer.py
```diff
@@ -137,3 +137,19 @@ def test_inverse_e_plus_user_passes_var_list(df_vartypes):
     assert transformer.n_features_in_ == 5
     # test transform output
     pd.testing.assert_frame_equal(X, df_vartypes)
+
+
```
solegalli, line 144: We should bring the tests for C, from test_logcp to this file as well, so we know it works as intended.

solegalli (commented):
Hi @adityaanikam 

This is looking really good! Thank you so much for the first draft.

I left a few comments, mostly to add future warnings about the deprecation, and to add the tests for C in the logtransform tests.

The last bit that needs updating to have this one ready is the user guide. Inside the docs folder, useguide/transformers, we need to add the demo of the use of C to Logtransfor.rst. You could probably re-adapt the examples that we have for logcp.

Thanks a lot!

adityaanikam:
Hi @solegalli, just wanted to flag something separate from the code review. CI has failed in the same way on three different runs of this branch now. Most recently, `test_feature_engine_py313` and `test_feature_engine_py314` failed on `7eae67fd`, each with the same result:

```
51 failed, 1969 passed
```

All 51 failures seem to come from `fetch_california_housing()` failing with:

```
sklearn.datasets._base._fetch_remote → urlretrieve → urllib.error.HTTPError: HTTP Error 403: Forbidden
```

That accounts for 48 failures in `test_wrappers/test_sklearn_wrapper.py`, 2 in `test_discretisation`, and 1 in `test_selection/test_mrmr.py`. They’re all tests that fetch the California housing dataset, and none of the failures are in `test_transformation`.

The Python version that fails also seems to vary between runs  py310 on one, py314 on another, and py313 + py314 most recently  so it looks more like a flaky download/CI issue than something caused by this change.

I had a look at the dataset URL as well. It redirects to a presigned S3 URL with `X-Amz-Expires=10`, so the signature only has a 10 second window. I wonder if a runner occasionally takes too long to follow the redirect and ends up hitting an expired URL. The download works consistently for me locally.

Since `fetch_california_housing()` caches the dataset under `~/scikit_learn_data` after the first successful download, caching that directory in CircleCI might avoid this altogether.

Happy to look into that if it would be useful, but wanted to flag it first in case you’re already aware of it or have a preferred way of handling it.

docs/user_guide/transformation/LogTransformer.rst
```diff
@@ -28,9 +28,10 @@ LogTransformer
 
 .. note::
 
-    Note that the logarithm can only be applied to positive values. Thus, if the variable contains 0 or negative values, this transformer will return an error.
+    By default (``C=0``), the logarithm can only be applied to positive values. If the variable contains 0 or negative values, the transformer will return an error, unless you use the ``C`` parameter described below.
```
solegalli:
```suggestion
    The logarithm can only be applied to positive values; if the variable contains 0 or negative values, the transformer will raise an error. To avoid this, add a constant to the variables by using the `C` parameter (more details below).
```

docs/user_guide/transformation/LogTransformer.rst
```diff
@@ -28,9 +28,10 @@ LogTransformer
...
 
-To transform non-positive variables you can add a constant to shift the data points towards positive values. You can do this by using :class:`LogCpTransformer()`.
+To transform non-positive variables, pass a value to the ``C`` parameter to shift
```
solegalli:
```suggestion
```

docs/user_guide/transformation/LogTransformer.rst
```diff
@@ -28,9 +28,10 @@ LogTransformer
...
+the data points towards positive values, as shown in the section below.
```
solegalli:
```suggestion
```

solegalli: We can remove these sentences, since it's included in the note above.

docs/user_guide/transformation/LogTransformer.rst
```diff
@@ -161,6 +162,64 @@ In the following plots we see histograms showing the variables in their original
     Following the transformations with scatter plots and residual analysis of the regression models helps understand if the transformations are useful in our regression analysis.
 
 
+Transforming non-positive variables with C
+------------------------------------------
```
solegalli: We need to replace the lines under the hading -- with sublevel heading so ˜˜˜, so that this heading appears under the heading Python implementaiton

docs/user_guide/transformation/LogTransformer.rst
```diff
@@ -161,6 +162,64 @@ In the following plots we see histograms showing the variables in their original
...
+negative values, by adding a constant ``C`` before applying the logarithm,
+i.e. ``log(x + C)``.
+
+Let's load the diabetes dataset that comes with scikit-learn, which contains
+variables with negative values:
+
+.. code:: python
+
+    from sklearn.model_selection import train_test_split
+    from sklearn.datasets import load_diabetes
```
solegalli, line 174: this is a nitpick, but since we are editing still, it's nicer to have imports in alphabetical order, so datasets, before model_selection

tests/test_transformation/test_logcp_transformer.py
```diff
@@ -28,7 +28,7 @@ def test_c_parameter(c):
 
 @pytest.mark.parametrize("c", ["string", [1, 2]])
 def test_c_raises_error(c):
```
solegalli, line 30: we need to add a test for the future deprecation warning to this file

solegalli (commented):
Hi @adityaanikam 

This is almost ready, thank you so much! 

A few more things are still needed:

We need to remove LogCpTransformer from the following files:

[README.md:122](vscode-webview://15avd2gfac1odomq3hh0nltja0niulr44uvvvs8n4aarnjf1putt/README.md#L122) — plain bullet list entry
[docs/index.rst:240](vscode-webview://15avd2gfac1odomq3hh0nltja0niulr44uvvvs8n4aarnjf1putt/docs/index.rst#L240) — API summary table
[docs/user_guide/transformation/index.rst:31,52](vscode-webview://15avd2gfac1odomq3hh0nltja0niulr44uvvvs8n4aarnjf1putt/docs/user_guide/transformation/index.rst#L31) — comparison table + toctree
[docs/user_guide/transformation/PowerTransformer.rst:77](vscode-webview://15avd2gfac1odomq3hh0nltja0niulr44uvvvs8n4aarnjf1putt/docs/user_guide/transformation/PowerTransformer.rst#L77) — "see also" list
[docs/api_doc/transformation/index.rst:13](vscode-webview://15avd2gfac1odomq3hh0nltja0niulr44uvvvs8n4aarnjf1putt/docs/api_doc/transformation/index.rst#L13) — API toctree

While you are at it, please add that the LogTransformer now adds, optionally a constant if relevant. For example in the table (not sure it fits, I don't remember what the table says at the moment).

We also need a test for LogCP to corroborate that the future warning is being triggered.

Finally, we need a note in this file docs/user_guide/transformation/LogCpTransformer.rst mentioning that it was deprecated in version 2.0.0 and will be removed in version ...

After that, we are ready to merge!

Could you take a look at this and other comments that I left in the files themselves?

adityaanikam:
Hi @solegalli, pushed everything from this round removed <code>LogCpTransformer</code> from the 5 listing files, updated the comparison table, applied the note rewrite/heading/import fixes on <code>LogTransformer.rst</code>, added the deprecation warning note to <code>LogCpTransformer.rst</code>, and added the <code>FutureWarning</code> test to <code>test_logcp_transformer.py</code>.</p>
<p>One CI issue along the way, for the record: removing <code>LogCpTransformer</code> from the toctrees while keeping its <code>.rst</code> pages (per your comment) left those two pages unreachable from any toctree, which <code>sphinx-build -W</code> treats as a hard error. Fixed by adding <code>:orphan:</code> to both <code>docs/api_doc/transformation/LogCpTransformer.rst</code> and <code>docs/user_guide/transformation/LogCpTransformer.rst</code> verified with a full clean local build (213/213 pages, no warnings) before pushing.</p>
<p>Two small things from your earlier round I want to flag explicitly rather than let slide through silently, since you said this was close to merge ready:</p>
<ol>
<li>The <code>LogCpTransformer</code> note currently says "deprecated in 2.0.0, will be removed in 2.1.0" rather than your literal suggestion ("consolidated in 2.0.0 and will be deprecated in 2.1.0"). I went with this wording because it's the one consistent with the actual <code>FutureWarning</code> message, which fires now (in 2.0.0) rather than waiting until 2.1.0  so the class is already deprecated, not "about to be." Happy to change it if you'd rather phrase it differently.</li>
<li>The <code># TODO: remove in version 2.1.0</code> sits as a plain comment above the <code>class</code> statement rather than inside the docstring where your suggestion diff placed it. I kept it outside because a docstring is the API documentation Sphinx renders — inside it, that line would show up as user-facing text on the published docs page, which I don't think was the intent. Let me know if you did want it in the docstring for some reason I'm missing.</li>

solegalli:
Hi @adityaanikam 

Thank you very much for the changes. You were right to mention the thing about the orphan pages. We shouldn't remove the rst files just yet, because we need to inform users that it still exists but not for longer. So I reverted those changes and will merge once the tests pass. 

Thanks a lot for your contribution and you are welcome to contribute again, whenever you feel like it.

PR #972, opened 2026-07-26
Rename imputer classes and add deprecation aliases

### LINKED ISSUE #971: rename imputer classes
- [ ] MeanMedianImputer into MeanImputer
- [ ] AddMissingIndicator into MissingIndicator
- [ ] NumericalAbitraryImputer into ArbitraryImputer

We use 1 cycle of deprecation warnings, so old classes should still remain with a future warning that is raised at init saying that this class was deprecated in version 2.0.0 in favour of... and will be removed in version 2.1.0. User should use [new name of class] instead.

Add #TODO remove in version 2.1.0 on top of deprecated class.

tests should pass for both classes in version 2,0,0

aniq63:
## What this PR does

This PR renames the imputer classes requested  while maintaining backward compatibility through a deprecation cycle.

Fixes #971 

### Changes

* Renamed `MeanMedianImputer` to `MeanImputer`
* Renamed `AddMissingIndicator` to `MissingIndicator`
* Renamed `ArbitraryNumberImputer` to `ArbitraryImputer`
* Added deprecated aliases with `FutureWarning`
* Updated package exports
* Updated tests to cover both the new and deprecated class names, including deprecation warnings

### Testing

* All updated tests pass successfully.


solegalli:
Hi @aniq63 


Thanks a lot for your contribution. 

We need to split this PR into 3, one per class. 

We also need to update the documentation relevant for each transformer, which you'll find inside the docs folder. Basically, we need to replace all calls to the old transformer names by the new transformer names. 

We also need to rename files from if necessary, for example if a file was called meanmedianimputer.py or .rst, we need to rename that to the new name of the transformer.

There is also this PR that adds the same changes #972, so maybe you could start with MissingIndicator, and we take it from there?

Note that the classes are tested in additional files and need to be updated there too: ([tests/check_estimators_with_parametrize_tests.py](vscode-webview://15avd2gfac1odomq3hh0nltja0niulr44uvvvs8n4aarnjf1putt/tests/check_estimators_with_parametrize_tests.py), [tests/test_imputation/test_check_estimator_imputers.py](vscode-webview://15avd2gfac1odomq3hh0nltja0niulr44uvvvs8n4aarnjf1putt/tests/test_imputation/test_check_estimator_imputers.py))

aniq63:
Closing this PR as requested since the changes have been split into separate PRs. Thank you for the feedback!

The new Seprate PR are following:

Rename MeanMedianImputer to MeanImputer #975 

Rename AddMissingIndicator to MissingIndicator with deprecation support- #976
#976

Rename ArbitraryNumberImputer to ArbitraryImputer with deprecation support- #977
#977

PR #981, opened 2026-07-30
Rename SklearnTransformerWrapper to SklearnWrapper

solegalli:
## Summary

- Renames `SklearnTransformerWrapper` to `SklearnWrapper` for a shorter name.
- The old class name is kept as a deprecated alias (`FutureWarning` on instantiation), to be removed in 2.1.0, following the same pattern used in the recent `MeanImputer`, `MissingIndicator`, `Winsoriser`, `MeanNormalisationScaler`, `LogTransformer` and `ArbitraryImputer` renames.
- The `feature_engine.wrappers` package/module names are unchanged.
- Updates API docs, user guide, README, docstrings, and tests to reference the new name.

## Test plan

- [x] `pytest tests/` (2073 passed)
- [x] `flake8 feature_engine tests`
- [x] `mypy feature_engine/wrappers`
- [x] `sphinx-build -W -b html docs <out>` (clean build, 0 warnings)

### DIFF
--- a/feature_engine/wrappers/wrappers.py
+++ b/feature_engine/wrappers/wrappers.py
@@ -237,7 +237,7 @@ class SklearnWrapper(TransformerMixin, BaseEstimator):
 
         if transformer.__class__.__name__ == "OneHotEncoder":
             msg = (
-                "The SklearnWrapper can only wrap the OneHotEncoder if the "
+                "SklearnWrapper can only wrap OneHotEncoder if the "
                 "sparse is set to False."
             )
             if getattr(transformer, "sparse", False) or getattr(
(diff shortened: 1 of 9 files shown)


feature_engine/wrappers/wrappers.py
```diff
@@ -236,7 +237,7 @@ def __init__(
 
         if transformer.__class__.__name__ == "OneHotEncoder":
             msg = (
-                "The SklearnTransformerWrapper can only wrap the OneHotEncoder if the "
+                "The SklearnWrapper can only wrap the OneHotEncoder if the "
```
solegalli:
```suggestion
                "SklearnWrapper can only wrap OneHotEncoder if the "
```

PR #986, opened 2026-08-17
Fix incorrect precision example in GeometricWidthDiscretiser docs

maupatel:
The `GeometricWidthDiscretiser` guidance on the `precision` parameter gives an inconsistent worked example.

`precision` is the number of decimal places (forwarded to `pd.cut(..., precision=...)`), so `precision = N` corresponds to a smallest step of `10**-N`. The tip establishes this with its first example (`precision = 3` -> `0.001`), but then states `precision = 7` -> `0.0001`, which is `10**-4`, not `10**-7`. The correct value is `0.0000001`.

Fixed in both places the sentence appears (kept in sync):
- `feature_engine/discretisation/geometric_width.py` (docstring)
- `docs/user_guide/discretisation/GeometricWidthDiscretiser.rst`

Documentation-only; no code or behavior changes.


solegalli:
Good catch! Thank you!

PR #987, opened 2026-08-19
Speed up MathFeatures reductions with NumPy

### LINKED ISSUE #576: MathFeatures seems much slower than pandas.sum()
Need to check what is going on and fix

khan-sudais:
## Summary

- use vectorized NumPy reducers for common `MathFeatures` row-wise aggregations
- preserve pandas semantics for string and NumPy callable reductions, including `ddof` and missing values
- retain the existing pandas `agg` fallback for custom functions, uncommon reducers, and nullable extension dtypes
- document and test the optimized behavior while keeping the benchmark comparable

## Why

`MathFeatures` currently routes common reductions through row-wise `DataFrame.agg`, which adds substantial per-row Python overhead. On a local 100,000-row x 5-feature benchmark using `sum`, `mean`, `std`, `min`, and `max`, the legacy logic took 30.49 seconds while the optimized path took 0.0159 seconds (about 1,920x faster).

## Compatibility

- matches pandas results for missing and all-missing rows
- string `std` and `var` retain pandas' `ddof=1` behavior
- NumPy callable semantics remain version-aware
- custom callables and nullable dtypes continue through the legacy pandas fallback

## Validation

- 36 `MathFeatures` tests passed on pandas 2.3.3 and pandas 3.0.5
- 184 creation tests passed on both pandas versions
- 2,062 tests passed in the initial full suite; the 51 dataset-cache failures were rerun with a writable cache as part of 163 passing affected-file tests
- `flake8` passed
- `mypy` passed for the changed module
- changed module reached 99% statement coverage with no missed statements
- Sphinx documentation build passed with warnings treated as errors

Closes #576

### DIFF
--- a/benchmarks/test_creation.py
+++ b/benchmarks/test_creation.py
@@ -18,8 +18,8 @@ NUM_VARS = numerical_vars()
     "func", [["sum", "mean"], ["sum", "mean", "std", "min", "max"]]
 )
 def test_math_features_transform(benchmark, df_tiny, func):
-    # MathFeatures aggregates row-wise, which is orders of magnitude slower per
-    # row than the vectorised transformers, hence the smallest dataframe.
+    # Keep the original dataset size so results remain comparable with the
+    # pre-NumPy baseline recorded for issue #576.
     creator = MathFeatures(variables=NUM_VARS, func=func)
     creator.fit(df_tiny)
     benchmark(creator.transform, df_tiny)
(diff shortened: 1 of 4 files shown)


solegalli:
@codspeedbot explain why this is faster

solegalli:
Great work @Sudaiskhan1133 thank you!

PR #988, opened 2026-08-22
BUG - Compute lexical_diversity as unique words over total words

VenishPaneliya:
`lexical_diversity` is documented as:

> **lexical_diversity**: Ratio of unique words to total words

but the implementation divides the other way round:

```python
"lexical_diversity": lambda x: x.str.strip().str.split().str.len()
/ x.str.lower().str.split().apply(set).str.len(),
```

so it returns the reciprocal of the documented value.

Two things make this visible:

- A type-token ratio cannot exceed 1, but the current feature returns values greater than 1 for any text that repeats a word.
- It grows with repetition instead of with variety, which is the opposite of what the user guide states: *"in more complex texts, both the number of unique words and the lexical diversity are greater."*

```python
import pandas as pd
from feature_engine.text import TextFeatures

X = pd.DataFrame({"review": [
    "the cat sat on the mat",       # 6 words, 5 unique
    "good good good good",          # 4 words, 1 unique
    "all words here are distinct",  # 5 words, 5 unique
]})
print(TextFeatures(variables=["review"],
                   features=["lexical_diversity"]).fit_transform(X))
```

On `main`:

```
review_lexical_diversity
                     1.2
                     4.0
                     1.0
```

With this PR:

```
review_lexical_diversity
                0.833333
                0.250000
                1.000000
```

`"good good good good"` is four tokens of a single word — the least diverse row in the frame — and it currently gets the highest score, 4.0, instead of 1/4.

This swaps the operands so the numerator is the unique word count, reusing exactly the expression the neighbouring `unique_word_count` feature already uses. The example output in `docs/user_guide/text/TextFeatures.rst` is updated to match (row 1 goes from an impossible `1.25` to `0.8`).

Added a regression test. It fails on `main` and passes here; the rest of `tests/test_text/` is unaffected (38 passed before, 39 after), since no existing test covered a text with a repeated word.

### DIFF
--- a/docs/user_guide/text/TextFeatures.rst
+++ b/docs/user_guide/text/TextFeatures.rst
@@ -246,7 +246,7 @@ The output dataframe contains all 20 text features extracted from the `review` c
 
        review_unique_word_count  review_lexical_diversity
     0                         7                       1.0
-    1                         4                      1.25
+    1                         4                       0.8
     2                         9                       1.0
     3                         4                       1.0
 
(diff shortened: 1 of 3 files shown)


solegalli:
Thank you! Good catch!

PR #990, opened 2026-08-24
refactor creation base for narwhals compatibility

solegalli:


feature_engine/_base_transformers/mixins.py
```diff
@@ -15,40 +17,54 @@
...
             as many rows as those left in X_new.
         """
         X, y = check_X_y(X, y)
-        X = self.transform(X)
-        y = y.loc[X.index]
+
+        if nwd.is_pandas_dataframe(X) is True:
+            X = self.transform(X)
+            y = y.loc[X.index]
+        else:
```
solegalli, line 49: Hi @FBruzzesi !

I just wanted to know if this is the best way to drop rows based on index in dfs that don't have index like polars? Basically, feature engine will drop some rows on the training set if they have nans for example, or outliers. Then, it automatically adjusts the target variable which is in a different series (or df) by removing the same rows. In pandas is straightforward. But my understanding is that polars does not have an index. So this is a workaround. 

What are your thoughts?

FBruzzesi: Yes I think using `__getitem__` is the best you can do. I would double check if you can use a series directly without converting to a list. I would also consider having a fast path for transformers that guarantee to not shuffle the row order/change the input length

solegalli: Thank you!

feature_engine/_base_transformers/mixins.py
```diff
@@ -15,40 +17,54 @@
...
+            row_index_col = "__feature_engine_row_index__"
+            nw_X = nw.from_native(X, eager_only=True).with_row_index(row_index_col)
+            X = self.transform(nw_X.to_native())
+            nw_X = nw.from_native(X, eager_only=True)
+            row_positions = nw_X.get_column(row_index_col).to_list()
```
solegalli:
```suggestion
            row_positions = nw_X.get_column(row_index_col)
```

PR #996, opened 2026-08-24
Migrate DecisionTreeFeatures to narwhals, add polars support

solegalli:
Follows the same pandas-native / narwhals-generic split established for GeoDistanceFeatures (this transformer also reimplements fit()/transform() directly, not via BaseCreation): .columns extraction, column reorder, prediction-column assignment, and drop_original all split by backend, consistent with every other operation in this module that's been benchmarked as a real (not minimal) loss when routed through narwhals on pandas.

Confirmed empirically before designing: sklearn's DecisionTreeRegressor/ Classifier and GridSearchCV accept a polars DataFrame directly for both fit() and predict()/predict_proba(), so the actual tree training/inference calls are unchanged - only the surrounding column selection, extraction, and reassembly needed migrating.

Fixed a pre-existing bug found while rewriting the exact code path it lived in: single-feature combos with an integer column name (e.g. DecisionTreeFeatures(features_to_combine=1) on a dataframe with columns 0, 1, ...) crashed, since the original `isinstance(features, str)` check missed the int case and fell through to plain X[features] indexing, which returns a 1D Series rather than the 2D input sklearn requires. Widened to isinstance(features, (str, int)); verified the same single-feature narwhals path (get_column().to_frame()) already handles both cleanly.

Regression, binary classification, and multiclass classification paths all verified to produce identical predictions between pandas and polars input. return_empty=True + polars remains untestable here too (same nw.col([]) bug in dataframe_checks.py found during CyclicalFeatures, still tabled) - this is the second transformer it blocks.

docs/user_guide/creation/DecisionTreeFeatures.rst is large (511 lines) and built around actual cross-validated tree fitting on the real California housing dataset across many sections - re-verified the cheap, deterministic parts (the raw data table) but did not re-run every tree-fitting example given the cost of repeated grid-search CV fits; unlike the other three creation-module docs this pass touched, the rest of this file's numbers are unverified. Added a self-contained "With polars" section using simple synthetic data instead, fully verified.

### DIFF
--- a/feature_engine/creation/decision_tree_features.py
+++ b/feature_engine/creation/decision_tree_features.py
@@ -141,10 +141,7 @@ class DecisionTreeFeatures(TransformerMixin, BaseEstimator, GetFeatureNamesOutMi
         of its computation, which avoids the overhead of copying the entire
         dataframe to separate worker processes. `None` means 1, i.e. sequential
         training (this transformer's original behaviour); `-1` means using all
-        available processors. There is a real speedup only when there are many
-        feature combinations and/or a large `param_grid` to search — with just
-        a handful of combinations, thread-dispatch overhead outweighs the gain,
-        which is why the default stays sequential.
+        available processors.
 
     {missing_values}
 
--- a/feature_engine/creation/decision_tree_features.py
+++ b/feature_engine/creation/decision_tree_features.py
@@ -4,6 +4,7 @@ from typing import Any, Dict, Iterable, List, Optional, Union
 import narwhals as nw
 import narwhals.dependencies as nwd
 import numpy as np
+from joblib import Parallel, delayed
 from narwhals.typing import IntoDataFrame, IntoSeries
 from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.model_selection import GridSearchCV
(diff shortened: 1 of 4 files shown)


feature_engine/creation/decision_tree_features.py
```diff
@@ -302,39 +334,48 @@ def fit(self, X: pd.DataFrame, y: pd.Series):
             how_to_combine=self.features_to_combine, variables=variables_
         )
 
+        is_pandas = nwd.is_pandas_dataframe(X) is True
```
solegalli:
```suggestion
        is_pandas = nwd.is_pandas_dataframe(X)
```

feature_engine/creation/decision_tree_features.py
```diff
@@ -351,50 +392,59 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
         _check_contains_na(X, self.variables_)
         _check_contains_inf(X, self.variables_)
 
+        is_pandas = nwd.is_pandas_dataframe(X) is True
```
solegalli:
```suggestion
        is_pandas = nwd.is_pandas_dataframe(X)
```

PR #1000, opened 2026-08-25
Migrate ArcSinhTransformer to narwhals, add polars support

solegalli:
Same elementwise-math shape as ArcsinTransformer: extract the transform columns to one numpy array via narwhals' to_numpy(), apply np.arcsinh((x - loc) / scale) once, reassign via nw.new_series + with_columns. Benchmarked against the old pandas-native .loc assignment across 10k-100k rows and 1-10 columns: narwhals-on-pandas was consistently faster than the old code (0.48x-0.83x of old runtime), so merged into one narwhals-generic path with no backend branch.

Found a pre-existing stale docstring while verifying output against the old code: the class docstring's example table (arcsinh of np.random.randn(100) * 1000 with seed 42) printed values that don't match what either the old or new code actually produces (e.g. 7.516076 vs the real 6.901163 for the first row) - confirmed by running the old (pre-migration) code directly, so this predates the migration. Fixed the docstring numbers to the verified real output. The docs/user_guide/transformation/ArcSinhTransformer.rst walkthrough's printed tables were re-run and already matched exactly, so those were left as-is; added a verified "With polars" section to both the docstring and the user guide.

Rewrote test_arcsinh.py to parametrize every behavior over pandas and polars input (previously pandas-only).

### DIFF
--- a/docs/user_guide/transformation/ArcSinhTransformer.rst
+++ b/docs/user_guide/transformation/ArcSinhTransformer.rst
@@ -553,6 +553,44 @@ The recovered data:
     493  -3.258723  9405.785347
     122  30.047946  1448.874284
 
+With polars
+-----------
+
+:class:`ArcSinhTransformer()` works in the same way with a polars dataframe:
+
+.. code:: python
+
+    import polars as pl
+    from feature_engine.transformation import ArcSinhTransformer
+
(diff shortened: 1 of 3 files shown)


solegalli:
Why were the test dfs changed to floats?

PR #979, opened 2026-07-29
[MNT] migrate scaling module to narwhals

ojassharma7:
## Summary
Migrates `feature_engine/scaling` (`MeanNormalisationScaler`) so `fit` / `transform` / `inverse_transform` use **narwhals** for the mean / range / scaling math, returning a native dataframe of the same type as the input.

## Context
Part of #965. @solegalli suggested `scaling` as a good first module. Variable selection (`find_numerical_variables` / friends) is still pandas-backed on this branch — as noted, pandas tests should pass; full polars support waits on the variable-handling work Sole is doing.

## Test plan
- [x] `pytest tests/test_scaling/` (pandas) — 8 passed locally
- [ ] CI green against `narwhals-migration`

PR #1004, opened 2026-08-25
[MNT] migrate scaling module to narwhals

ojassharma7:
Redo of #979 against the current `narwhals-migration` (that branch had gotten far ahead — dataframe_checks was refactored again in #989, variable_handling rewritten in #978, creation base rewritten in #990 — so a fresh branch was safer than rebasing).

`fit()`/`transform()`/`inverse_transform()` now go through a single narwhals+numpy path (no pandas/polars branch): wrap via `nw.from_native`, do the reduction/elementwise math in numpy, write back via `nw.new_series` + `with_columns` — same-named series replace the column in place.

Benchmarked narwhals-expressions vs. narwhals+numpy before choosing (100/10k/200k rows, 3/20 variables, both backends): numpy wins 2x-73x at small/medium scale, and even where narwhals-expr pulls ahead (200k rows, polars) it's only ~2x — well under the ~1.7x bar this migration's been using to justify a backend split, so this stays unified like `RelativeFeatures`.

`return_empty=True` guarded explicitly (same fix `CyclicalFeatures` needed — narwhals' `select([])` collapses row count too).

Found and fixed while verifying, not assumed: the docstring's example values were literally the raw pre-normalization random draws, never the actual `transform()` output, and the user guide's inverse_transform table showed `Age` as a bare int when both the old and new code have always produced `float64` there. Added a "With polars" section per AGENTS.md.

Tests rewritten to the single-parametrized-over-both-backends convention.

**Verification:** full suite run twice — this branch and unmodified `narwhals-migration` HEAD — identical 68 pre-existing failures in both (diffed directly, not just counted), 2273→2287 passed. flake8 + mypy clean.


solegalli:
Hi @ojassharma7, thanks for this migration. As maintainers we pushed an update to this branch so it lines up with the rest of the narwhals-migration PRs:

- Rebased onto #1045, which adds the shared test fixtures and helpers now used by all migrated transformers. This branch was based on a commit from before #1019, so it is now stacked on #1045 and its diff includes that PR's commits until it merges.
- `tests/test_scaling/test_mean_normalization.py` rewritten to the shared structure: the `make_df` fixture instead of repeated `parametrize("make_df", ...)`, `isinstance(X, make_df)` plus `to_dict()` checks (`pytest.approx` for floats) instead of the local `assert_df_equal` / `_none_to_nan` helpers, and `pytest.raises(match=...)` for the NA errors.
- Your source changes are untouched: `MeanNormalisationScaler` goes through `BaseNumericalTransformer`, which already handles the narwhals-returning `check_X`.

Locally `tests/test_scaling` passes (33 tests). If you have local commits on this branch, please reset to the updated remote before continuing.

PR #1017, opened 2026-08-28
datetime: propagate NaN for week and days_in_month with missing_values=ignore

AlejandroCoronadoN:
When `missing_values="ignore"`, `DatetimeFeatures.transform()` raises for the `week` and `days_in_month` features if the input has a missing date. Both cast to `np.int64`, which fails on the NaT-derived NA:

```
IntCastingNaNError: Cannot convert non-finite values (NA or inf) to integer
```

The other datetime features (month, year, day_of_month, ...) do not force an int cast and correctly return NaN for the missing row, so only these two were affected.

Repro:

```python
import pandas as pd
from feature_engine.datetime import DatetimeFeatures

X = pd.DataFrame({"date": pd.to_datetime(["2022-01-01", None, "2022-03-01"])})
DatetimeFeatures(features_to_extract=["days_in_month"],
                 missing_values="ignore").fit(X).transform(X)
```

This casts both to `float64` instead, so the missing row becomes NaN like the other features. Added a test covering `week` and `days_in_month` with `missing_values="ignore"`.


feature_engine/datetime/_datetime_constants.py
```diff
@@ -73,7 +73,7 @@
     "year_start": lambda x: x.dt.is_year_start.astype(np.int64),
     "year_end": lambda x: x.dt.is_year_end.astype(np.int64),
     "leap_year": lambda x: x.dt.is_leap_year.astype(np.int64),
-    "days_in_month": lambda x: x.dt.days_in_month.astype(np.int64),
+    "days_in_month": lambda x: x.dt.days_in_month.astype("float64"),
```
solegalli:
```suggestion
    "days_in_month": lambda x: x.dt.days_in_month,
```

solegalli:
Thank you @AlejandroCoronadoN !!

PR #1026, opened 2026-08-30
Migrate DecisionTreeEncoder to narwhals, add polars support

solegalli:
Depends on #1058 (integer column names in the discretisers). Until #1058 is merged, this PR also shows its files.

Migrates `DecisionTreeEncoder` to narwhals, with polars support.

- `fit` uses the same pipeline as before the migration: `OrdinalEncoder` (`encoding_method`, `ignore_format`) followed by `DecisionTreeDiscretiser` (`cv`, `scoring`, `param_grid`, `regression`, `random_state`, `n_jobs`), both already migrated. The category-to-prediction mapping is read from the unique category and prediction pairs of the pipeline output, rounded with `precision`.
- New `n_jobs` parameter, passed to `DecisionTreeDiscretiser` to fit the trees in parallel.
- Tests follow the shared test structure, including a list/array target test and a pandas-only integer column name test.

`encoder_dict_` is identical to the previous hand-written implementation in 37 cases (pandas and polars, both encoding methods, regression and classification, `precision` None/0/2, list target, numeric variables, `n_jobs=2`, integer column names). The pipeline is about 10% slower on pandas and 2% slower on polars than that implementation, because it transforms and predicts every row; most of the time is the grid search in both.


solegalli:
Updated this branch:

- No source changes.

Locally: `test_decision_tree_encoder.py` 60 passed; `tests/test_encoding` has no failures beyond those already on `narwhals-migration`.

solegalli:
waiting for #1042

PR #1027, opened 2026-08-30
Migrate MeanEncoder.fit() to narwhals, add polars support

solegalli:
Migrates `MeanEncoder.fit()` to narwhals with polars support. `transform()` / `inverse_transform()` already come dataframe-agnostic from `CategoricalMethodsMixin`.

`fit()` computes, per variable, the mean of `y` per category (and, with `smoothing="auto"`, the target variance per category), blended with the overall target mean.

**Merge vs split:** benchmarked a pure-narwhals `fit()` (`group_by`/`agg`) vs pandas-native (10k–100k rows × 1–10 cols × 5–50 categories). narwhals-on-pandas ran ~1.5x–2.9x slower, worst at the most common shape (1–2 cols, 50k–100k rows), crossing the ~1.7x real-loss threshold; narwhals-on-polars competitive to faster. So this splits on `is_pandas = nwd.is_pandas_dataframe(X)`: pandas keeps a close variant of its `value_counts`/`groupby` code, polars/other backends go through `group_by()`/`agg()`. No cross-backend tie-break fix needed here — `encoder_dict_` is a category→mean dict, not a rank-ordered list (verified identical dicts pandas vs polars across `smoothing=0.0/100/"auto"` and all three `unseen` settings).

**Bug fixed** (pre-existing, confirmed against the unmodified file): the old `fit()` always called `y.groupby(X[var])`, raising `AttributeError` when `y` is a numpy array (list/array-like `y`, as sklearn's `check_X_y` produces) — same bug the `OrdinalEncoder` sibling found. Fixed on the pandas branch by pairing `X[var]` with `y` via `.assign()` when `y` isn't a Series (positional alignment, matching how a Series `y` aligned by index), and on the narwhals branch via `nw.new_series`. Fixes `test_inverse_transform_when_*` (3) and `test_encoders_when_x_pandas_y_numpy[encoder1]`.

Tests: every test in `test_mean_encoder.py` rewritten as one `make_df in [pd.DataFrame, pl.DataFrame]` case per behaviour (41, up from 20), `y` passed as a plain list in most so the numpy-`y` fix is exercised throughout. `test_variables_cast_as_category` stays pandas-only.

Verified: `test_mean_encoder.py` 41 passed; full `tests/test_encoding` 345 passed / 13 failed — same IDs as the unmodified base minus the 4 fixed here. flake8 / mypy clean, sphinx -W clean. Docstring + `MeanEncoder.rst` non-Titanic examples verified against real output, "With polars" section added (Titanic examples untouched — no network in sandbox). Downstream `_prediction/base_predictor.py` and `target_mean_selection.py` test suites: 66 passed.


solegalli:
Updated this branch:

- No source changes.

Locally: `test_mean_encoder.py` 45 passed; `tests/test_encoding` has no failures beyond those already on `narwhals-migration`.

PR #1028, opened 2026-08-30
Migrate OneHotEncoder to narwhals, add polars support

solegalli:

Migrates `OneHotEncoder` to narwhals with polars support.

Uses narwhals' `to_dummies()` for the expansion rather than a manual numpy/dict loop (it's a real vectorized one-hot op on both backends). Two edge cases `to_dummies()` doesn't cover directly:

- A fixed-length prefix placeholder (`"__ohe_tmp__"`) swapped back out by slicing, not by the real column name: `to_dummies()` only prefixes with the Series name when it's truthy, so a falsy real name (e.g. an int column literally named `0`) would silently drop the prefix.
- Learned categories absent from a given transform batch get an explicit all-0 column, so unseen categories encode as 0 across the board — matching pre-narwhals behaviour exactly.

`fit()`'s `value_counts()`/`unique()` and `transform()`'s reassembly are a single unified narwhals path — no pandas/polars split, verified identical dummy columns/values on both backends for identical input.

Tests rewritten to the cross-backend-parametrized convention (local dict fixtures over `make_df in [pd.DataFrame, pl.DataFrame]`, narwhals-based column/sum assertions). `test_variables_cast_as_category` stays pandas-only.

Verified: 43/43 own tests; full encoding suite 340 passed / 17 pre-existing failures (matches the `narwhals-encoding-base` baseline exactly); flake8 / mypy clean; sphinx -W clean; no pandas import in this file.


solegalli:
Updated this branch:

- Column-order assertions are kept, since output column order is part of this encoder's contract.
- No source changes.

Locally: `test_onehot_encoder.py` 43 passed; `tests/test_encoding` has no failures beyond those already on `narwhals-migration`.

PR #1029, opened 2026-08-30
Migrate OrdinalEncoder.fit() to narwhals, add polars support

solegalli:

Migrates `OrdinalEncoder.fit()` to narwhals with polars support. `transform()` / `inverse_transform()` already come dataframe-agnostic from `CategoricalMethodsMixin`.

`fit()` has two paths: `"arbitrary"` (`X[var].unique()`) and `"ordered"` (target mean per category via `y.groupby(X[var])`).

**Merge vs split:** benchmarked a pure-narwhals `fit()` at 10k–100k rows × 1–10 cols × 5–50 categories — 5x–18x slower than pandas-native at every size (a large, consistent loss, unlike the ~1.1x for the transform hot path). So `fit()` splits on `is_pandas = nwd.is_pandas_dataframe(X)`: pandas keeps a close variant of its `groupby`/`unique` code (like-for-like benchmark within noise of the old code), polars/other backends go through `group_by()`/`agg()`/`sort()`/`unique()`. `"arbitrary"` is untouched.

Two issues, both confirmed pre-existing against the unmodified file:

1. **Bug (fixed):** `"ordered"` always called `y.groupby(X[var])`, raising `AttributeError` on a numpy `y` (list/array-like target, as sklearn's `check_X_y` produces). This is exactly what `test_encoders_when_x_pandas_y_numpy` exercises for `OrdinalEncoder` (encoder2, added 2022 for #376) — failed on the unmodified file, now passes. Fixed on both branches (pandas: `.assign()` to pair `X[var]` with `y`; narwhals: `nw.new_series`).
2. **Cross-backend ordering hazard (avoided):** grouping by category then sorting by target mean doesn't guarantee the same tie-break across backends (polars reversed two tied categories vs pandas). Old pandas code effectively tie-broke on the category itself; reproduced explicitly with a compound `.sort([target_name, var])`. Verified pandas and polars now produce the same dict for a tied-mean fixture, matching the old order.

Tests: every test rewritten as one `make_df in [pd.DataFrame, pl.DataFrame]` case per behaviour (43, up from 26). `test_variables_cast_as_category` stays pandas-only.

Verified: `test_ordinal_encoder.py` 43 passed; full `tests/test_encoding` 344 passed / 16 failed — identical IDs to the unmodified base (17; one being the bug fixed here). flake8 / mypy clean, sphinx -W clean. `OrdinalEncoder.rst` examples verified against real output (California Housing), "With polars" section added; Titanic examples untouched (no network in sandbox).


solegalli:
Updated this branch:

- No source changes.

Locally: `test_ordinal_encoder.py` 47 passed; `tests/test_encoding` has no failures beyond those already on `narwhals-migration`.

PR #1030, opened 2026-08-30
Migrate RareLabelEncoder to narwhals, add polars support

solegalli:
Migrates `RareLabelEncoder` to narwhals with polars support.

`fit()`: pandas `.unique()` / `.value_counts(normalize=True)` replaced with narwhals `Series.n_unique()` (cardinality check — matches pandas' plain `unique()` length, which counts null as its own category) and `drop_nulls().value_counts(sort=True, normalize=True)` (same `drop_nulls()`/`sort=True` reasoning as `CountEncoder.fit()`).

`transform()` doesn't reuse `CategoricalMethodsMixin._encode()` (that's a dict-based numeric remap; this encoder keeps frequent categories as-is and only replaces the rest). Rewritten from pandas' `.loc[~isin(...), feature] = replace_with` onto `nw.when(<Series>).then(<Series>).otherwise(nw.lit(replace_with))`. Series from `get_column()` (not `nw.col()`) keeps this working for pandas integer column names. A pandas Categorical column still needs `add_categories(replace_with)` first — kept as a small `is_pandas`-gated block (structural, like the base's existing reorder branches); polars has no matching restriction.

**Merge vs split:** first pass (`zip_with` + full replacement Series) averaged 2.41x pandas-native at 50k–100k rows, mostly the cost of building a full-length Series every `transform()`. Switched to `nw.when(keep).then(col).otherwise(nw.lit(replace_with))` so the backend broadcasts the scalar — dropped to 1.60x average, 1.12x–1.54x at 100k/10 cols. narwhals-on-polars faster than pandas-native throughout. Single narwhals path, no performance split — remaining overhead is fixed per-call cost, a few ms at most.

Tests rewritten to one parametrized test per behaviour over `make_df in [pd.DataFrame, pl.DataFrame]` (39, up from 29). Integer-column-name and category-dtype tests kept pandas-only. `test_max_n_categories_with_numeric_var` split into a pandas-only version and a new polars-only version documenting the real expected difference (polars can't hold mixed int/str in one column, so a numeric var + string `replace_with` casts the whole column to string).

Verified: `test_rare_label_encoder.py` 39 passed; full `tests/test_encoding` 336 passed / 17 pre-existing failures with identical IDs to the baseline. flake8 / mypy clean, sphinx -W clean. `RareLabelEncoder.rst` examples verified against actual output (also fixed an unrelated pre-existing `value_counts()` Series-name drift while there); "With polars" section added. Titanic section left untouched (SSL blocks urllib in sandbox; workaround showed matching `encoder_dict_`/transform output).


solegalli:
Updated this branch:

- Removes the `_to_pandas` helper: polars' `to_pandas()` needs pyarrow, which isn't a dependency, so the polars cases were failing.
- No source changes.

Locally: `test_rare_label_encoder.py` 39 passed; `tests/test_encoding` has no failures beyond those already on `narwhals-migration`.

PR #1031, opened 2026-08-30
Migrate StringSimilarityEncoder to narwhals, add polars support

solegalli:

Migrates `StringSimilarityEncoder` to narwhals with polars support.

`fit()` rebuilds `encoder_dict_` with narwhals `cast(nw.String)` / `value_counts`, matching the `CountEncoder` / `RareLabelEncoder` convention. `cast()` preserves nulls as null on both backends (verified), unlike pandas' `astype(str)` which stringifies NaN to `"nan"` — so `"impute"` mode can `fill_null("")` directly and `"ignore"` mode `drop_nulls()` before casting. This replaces the old `"nan"`/`"<NA>"` text-sentinel workaround with a real `col.is_null()` check that can't collide with a genuine category named `"nan"` or `"<NA>"` (edge cases still covered by `test_string_dtype_with_literal_nan_strings`).

`transform()`'s per-row `difflib.SequenceMatcher` similarity has no vectorized narwhals equivalent, so it's computed once per unique value via numpy broadcasting (`np.unique`'s inverse index fans the small per-unique matrix back to all rows) and reassembled with `nw.new_series()` / `with_columns()` — same pattern `DecisionTreeFeatures` uses for externally-computed columns.

**Merge vs split:** benchmarked a pandas-specific fast path (`X.join(dict-of-columns)`) vs the unified `with_columns()` across 10k–100k rows × 1–10 cols × 5–50 categories. Assembly overhead ranges 0.9x–6.25x by shape, but the `difflib` computation dominates wall time by 1–3 orders of magnitude in every realistic case (e.g. 30ms difflib vs <1ms assembly at 100k rows/20 categories). Unified implementation, no `is_pandas` split.

Tests rewritten as single parametrized cases over `make_df in [pd.DataFrame, pl.DataFrame]`; pandas-NA-sentinel tests (`np.nan`/`pd.NA`/`None`, `StringDtype`) kept pandas-only. All doc examples including the Titanic worked example re-verified against actual output; "With polars" section added.


solegalli:
Updated this branch:

- Tests of pandas-specific NA sentinels and the nullable string dtype stay pandas-only.
- No source changes.

Locally: `test_similarity_encoder.py` 57 passed; `tests/test_encoding` has no failures beyond those already on `narwhals-migration`.

PR #1032, opened 2026-08-30
Migrate WoEEncoder to narwhals, add polars support

solegalli:

Migrates `WoEEncoder` to narwhals with polars support.

`fit()` splits by backend: pandas keeps `_calculate_woe()`'s existing two-`groupby` implementation unchanged (it's directly unit-tested for that exact pandas-Series-with-category-index contract); polars/other narwhals backends use one `group_by()`, deriving the negative-class count as the complement of the positive-class count per category — benchmarked competitive with, and often faster than, pandas-native at 50k–100k rows. Zero-count-per-class `fill_value` handling preserved exactly.

**Bug fix:** `_check_fit_input()` assumed `y` was always a pandas Series (`y.nunique()`/`y.min()`/`y.max()`), breaking on a numpy `y` (list/array-like target, which sklearn's `check_X_y` converts via `column_or_1d`). Wrapped numpy `y` into a narwhals Series aligned to `X`'s backend; for pandas, also aligned it to `X`'s actual index, since `_calculate_woe()`'s `y.groupby(X[var])` aligns by index and a mismatched default `RangeIndex` silently drops every row instead of raising, leaving `encoder_dict_` empty. Fixes `test_encoders_when_x_pandas_y_numpy`'s `WoEEncoder` case (was failing on the unmigrated file — confirmed pre-existing).

Verified: 44/44 own tests; full encoding suite 342 passed / 16 failed (was 17 pre-existing on the `narwhals-encoding-base` baseline — one fewer here since this branch's own numpy-`y` bug is now fixed, rest confirmed unrelated). flake8 / mypy clean, sphinx -W clean.


solegalli:
Updated this branch:

- No source changes.

Locally: `test_woe_encoder.py` 48 passed; `tests/test_encoding` has no failures beyond those already on `narwhals-migration`.

PR #1033, opened 2026-08-30
Migrate BaseOutlier and WinsorizerBase to narwhals, add polars support

solegalli:
Shared base for all outlier transformers (`ArbitraryOutlierCapper` extends `BaseOutlier` directly; `Winsoriser` / `OutlierTrimmer` extend `WinsorizerBase`). Migrates the column reorder + NA/Inf checks in `_check_transform_input_and_state()`, the fold-limit estimation in `WinsorizerBase.fit()` (gaussian / iqr / mad / quantiles), and the capping step in `BaseOutlier._transform()` to be dataframe-agnostic.

**Capping (merge vs split):** benchmarked pandas-native `.clip()` loop vs a single narwhals `with_columns(...clip...)` vs grouping columns by which bound(s) apply and running up to 3 vectorized numpy calls (`np.clip`/`minimum`/`maximum`) — 10k/50k/100k rows × 1/2/10 cols. narwhals-generic alone was already near parity (0.95–1.49x), but the numpy-grouped version was faster: 0.16–0.82x on the homogeneous case (single tail, all columns share a bound — the common Winsoriser/OutlierTrimmer case) and 0.42–1.52x on mixed-coverage dicts (ArbitraryOutlierCapper). Adopted the numpy-grouped version as the single merged path for both backends.

- A first numpy attempt used a blanket `-inf`/`inf` sentinel for the missing side per column — that's a correctness bug: mixing an int64 array with a float `inf` bound upcasts the whole column to float64 even when the real present bound is int (e.g. `max_capping_dict=dict(x1=8)` expects int64 out). Grouping columns into "both bounds" / "right only" / "left only" buckets and passing only the bounds that exist avoids ever introducing an inf — dtype promotion now matches pandas `.clip()` exactly (verified byte-for-byte across all 4 methods × 3 tails, plus int-dtype and mixed-dict cases).
- Bug fixed (introduced while migrating `fit()`): plain `np.mean`/`std`/`quantile`/`median` propagate NaN, unlike pandas' NaN-skipping defaults. With `missing_values="ignore"` and NaN present this silently produced NaN caps. Fixed with the nan-aware variants. Caught by `test_winsorizer.py::test_transformer_ignores_na_in_df`.

`variables` / feature names can be int or str; `.select(list_of_ints)` raises `InvalidIntoExprError`, so every `.select()` call uses `nw.col(*variables)`.

Verified: `tests/test_outliers` — 83 passed, 3 pre-existing `check_estimator` failures (raw numpy-array input, rejected by `check_X()` under the narwhals dataframe-only contract; identical before and after). flake8 / mypy clean, sphinx -W clean. Runs `fit`/`_transform` end-to-end on polars with pandas import blocked.

Not migrated here (belongs to the follow-on transformer PRs): `ArbitraryOutlierCapper.fit()`/`transform()`, Winsoriser's `add_indicators` branch, `OutlierTrimmer.transform()`.


solegalli:
Updated this branch:

- Rebased onto #1045 (shared backend test fixtures and helpers). This branch was based on a commit from before #999/#1019, so it is now stacked on #1045 and its diff includes that PR's commits until it merges.
- Adapted `BaseOutlier` and `WinsorizerBase` to `check_X` returning a narwhals frame (#1019), mirroring the imputation and encoding modules: `fit()` binds `nw_X = check_X(X)` and passes the native `X` to the variable and NA/inf checks; `_check_transform_input_and_state()` returns the narwhals frame reordered to the train set; `_transform()` returns the native frame.
- Added `tests/test_outliers/conftest.py` (`data_normal_dist`, `data_na`), used by #1034, #1035 and #1036.

Locally: `tests/test_outliers` fixes 38 of the 63 failures on `narwhals-migration`, with none new; the stacked PRs fix the rest. flake8 and mypy clean.

PR #1034, opened 2026-08-30
Migrate ArbitraryOutlierCapper to narwhals, add polars support

solegalli:
Migrates `ArbitraryOutlierCapper` to narwhals with polars support.

`fit()` only builds dicts from user input and validates via `check_numerical_variables` (already narwhals-generic) — no numeric computation. The only pandas-specific line was `feature_names_in_ = X.columns.to_list()`, replaced with the same `is_pandas`-guarded pattern `WinsorizerBase.fit()` already uses. `transform()` was already dataframe-agnostic via `BaseOutlier._transform()`; only type hints changed (`pd.DataFrame` → `IntoDataFrame`).

**Merge vs split:** benchmarked fit+transform end-to-end at 10k/50k/100k rows × 1/2/10 cols. pandas-native (pre-migration) vs migrated-on-pandas were within noise (~0.9–1.1x); polars ran 2–4x faster on both. No split — single merged path. Confirmed the module needs zero pandas (reloaded with `sys.modules["pandas"] = None`, ran fit/transform on polars; int64 stays int64 for a same-dtype capping dict).

**Known pre-existing bug (flagged, not fixed here):** in the already-merged `BaseOutlier._transform()`, when a capping dict spans columns of different dtypes that land in the same bound-group (e.g. `max_capping_dict={"age": 50, "fare": 200}`, age int64 / fare float64, both "right only"), the group's columns are stacked into one 2D array via `to_numpy()` before `np.clip`, forcing a common dtype and upcasting age to float64. The pre-narwhals code clipped each column independently. Reproduces identically on both backends; lives in `base_outlier.py`, shared with Winsoriser/OutlierTrimmer, out of this file's scope.

Tests rewritten to one parametrized test per behaviour over `pd.DataFrame`/`pl.DataFrame`, using `nw.from_native(...).to_dict(as_series=False)` for backend-agnostic assertions. Docs "With polars" section added (float dtypes throughout, to sidestep the dtype-upcast issue above); pandas Titanic walkthrough untouched (no network in sandbox).

Verified: `tests/test_outliers` — 88 passed (up from 83), same 3 pre-existing `check_estimator` failures. flake8 / mypy clean, sphinx -W clean.

---


solegalli:
Updated this branch:

- `fit()` adapted to `check_X` returning a narwhals frame: `feature_names_in_` / `n_features_in_` now come from the narwhals frame, replacing the pandas/polars branching.
- Tests rewritten to the shared structure (`make_df`, `isinstance(X, make_df)` plus `to_dict()`, missing values as `None`, `pytest.raises(match=...)`).

Locally: `test_arbitrary_capper.py` 18 passed; no new failures in `tests/test_outliers`. flake8 and mypy clean.

PR #1035, opened 2026-08-30
Migrate OutlierTrimmer to narwhals, add polars support

solegalli:
Migrates `OutlierTrimmer` to narwhals with polars support.

`transform()` now filters rows via a single narwhals `.filter()` call built from a combined boolean expression (AND of each variable's right/left cap conditions), instead of a pandas `.loc` masking loop.

**Merge vs split:** benchmarked against pandas-native and a numpy boolean-mask extraction at 10k/50k/100k rows × 1/2/10 cols. The narwhals filter is within 1.4–1.75x of pandas-native at 10k rows (sub-ms absolute) and faster from 50k rows up (0.58x–0.97x). Single merged path — row filtering is exactly what narwhals `.filter()` pushes down to the native backend efficiently (unlike `BaseOutlier`'s elementwise capping, which benefits from numpy grouping).

**Bug fixed** in `TransformXyMixin.transform_x_y()` (`_base_transformers/mixins.py`): the non-pandas branch added a row-index marker column via `with_row_index()` and passed it to `self.transform()`, but never widened `feature_names_in_` / `n_features_in_`, so a column-count-checking `transform()` raised `ValueError` on the extra column. Latent because `OutlierTrimmer` is the first narwhals-migrated class to combine `TransformXyMixin` with a column-count-checking `transform()` on a non-pandas backend. Fix (guarded widen/restore around the `transform()` call) carried over verbatim from the same fix on `narwhals-drop-missing-data` (commit fd99caf), not yet merged into this branch.

Tests rewritten to one parametrized test per behaviour over `make_df in [pd.DataFrame, pl.DataFrame]`, plus a new test asserting caps on two different variables combine with AND. Docs: user-guide Titanic numbers had drifted from the current openml dataset (predates this migration) — corrected here, "With polars" section added.

---


solegalli:
Updated this branch:

- Resolved a conflict in `_base_transformers/mixins.py` with #1024: kept its `n_features_in_` widening in `transform_x_y()` and added this branch's `feature_names_in_` widening, which is still needed because `BaseOutlier` reorders X to `feature_names_in_` and would otherwise drop the row-index column on non-pandas backends.
- `transform()` now filters the narwhals frame returned by `_check_transform_input_and_state()` directly.
- Tests rewritten to the shared structure (`data_normal_dist` / `data_na` fixtures, `y` via `make_series`, `isinstance(X, make_df)` plus `to_dict()`).

Locally: `test_outlier_trimmer.py` 24 passed; full `tests/` suite has no failures beyond `narwhals-migration` (46 of them fixed on this stack). flake8 and mypy clean.

PR #967, opened 2026-07-22
[MNT] rename Winsoriser public name to british spelling

favourone:
## Summary

- Add `Winsoriser` as the preferred public class name.
- Retain `Winsorizer` as a backward-compatible alias.
- Export both names from `feature_engine.outliers`.
- Update current API documentation, user guide, and README.
- Preserve existing documentation paths and anchors for compatibility.
- Add an explicit identity test for the compatibility alias.

## Tests

- `python -m pytest tests/test_outliers/test_winsorizer.py -q` — 27 passed
- `python -m pytest tests/test_outliers -q` — 60 passed
- `flake8` — passed
- `git diff --check` — passed

Part of #958.

PR #1036, opened 2026-08-30
Migrate Winsoriser/Winsorizer to narwhals, add polars support

solegalli:
Migrates `Winsoriser` / `Winsorizer` to narwhals with polars support.

`WinsorizerBase.fit`/`transform` (shared base) are migrated on `narwhals-outliers-base`. This covers the `Winsoriser`-specific piece: `transform()`'s `add_indicators` path, which compares the capped output against the original input to build per-tail boolean flag columns and previously only worked on pandas. Module-level `import pandas`/`numpy` removed; X type hints use `IntoDataFrame`.

**Merge vs split:** benchmarked the `add_indicators` comparison+concat step at 10k/50k/100k rows × 1/2/10 cols. pandas-native (boolean comparison + `pd.concat`) is up to ~3x faster than the narwhals `with_columns` equivalent on pandas input, and the loss grows with column count (10 cols: ~2–3x slower). That crosses the "keep pandas fast path" threshold, so `transform()` splits on `nwd.is_pandas_dataframe`, matching `MissingIndicator`'s precedent: pandas keeps its comparison+concat logic (obtaining `pd` via `nw.from_native(...).__native_namespace__()` instead of importing it); a new narwhals `with_columns` path (per-column Series comparison, cast to Float64) covers polars and other backends.

Deprecation preserved exactly: `Winsoriser` is the current public name (British spelling, #967); `Winsorizer` is a deprecated subclass raising `FutureWarning`, removal in 2.1.0 — the reverse of what the class names suggest.

Tests: `test_winsorizer.py` converted from pandas-only fixtures to local dicts parametrized over `make_df in [pd.DataFrame, pl.DataFrame]`, asserting identical capping values, indicator columns and `get_feature_names_out()` on both backends. Missing-value dicts use `None` not `np.nan` in string columns (polars rejects float NaN in a string column).

Verified: `test_winsorizer.py` 93 passed; full `tests/test_outliers` 123 passed / 3 pre-existing `check_estimator` failures (identical against a `narwhals-outliers-base` baseline run). flake8 / mypy clean, sphinx -W clean. `Winsoriser.rst` examples verified against real output (house_prices dataset available), "With polars" sections added. Full polars `fit_transform` incl. `add_indicators` runs with pandas import blocked.

---


solegalli:
Updated this branch:

- Fixed the `add_indicators=True` path: it compared the capped output against `check_X(X)`, which is now a narwhals frame, so the pandas branch mixed pandas and narwhals objects (broadcast errors, wrong indicators). It now compares against the user's native `X`.
- Tests rewritten to the shared structure (`data_normal_dist` / `data_na` fixtures, `isinstance(X, make_df)` plus `to_dict()`, `pytest.raises(match=...)`).

Locally: `test_winsorizer.py` 93 passed; no new failures in `tests/test_outliers`. flake8 and mypy clean.

PR #1037, opened 2026-08-30
Migrate BaseDiscretiser to narwhals, add polars support

solegalli:
Shared base for `ArbitraryDiscretiser`, `EqualFrequencyDiscretiser`, `EqualWidthDiscretiser` and `GeometricWidthDiscretiser` (not `DecisionTreeDiscretiser`, which extends a different base). Only `transform()` needed migrating — `_fit_setup()`, `_get_feature_names_in()` and `_check_transform_input_and_state()` are inherited unchanged from the already-migrated `BaseNumericalTransformer`.

`transform()`'s only pandas dependency was `pd.cut` per column. Replaced with a plain numpy implementation: `pandas.cut` is itself built on `bins.searchsorted()` internally (verified against pandas 3.0's `_bins_to_cuts` source), so `np.searchsorted` + the same `include_lowest` index-1 special case reproduces its bin-index logic exactly, no per-backend branch — values come from `nw_X.get_column(feature).to_numpy()` regardless of backend, results re-attached via `nw.new_series()`/`with_columns()`.

**Merge vs split:** benchmarked old `pd.cut` vs the new numpy+narwhals path at 10k/50k/100k rows × 1/2/10 cols:
- `return_boundaries=False` (bin codes): narwhals-on-pandas ~1.0–1.2x at realistic sizes (the ~1.9x at 10k/1-col is sub-ms fixed overhead) — minimal loss, single path. narwhals-on-polars ~1.0–1.3x *faster*.
- `return_boundaries=True` (interval-label strings): the numpy path is 12–20x faster than `pd.cut` on pandas itself (100k×10: 647ms → 40ms) — `pd.cut`'s Categorical/IntervalIndex machinery has heavy per-call overhead. polars ~1.2x faster still.

`return_boundaries=True`'s `"(lower, upper]"` label formatting replicates `pd.cut`'s `_round_frac`/`_infer_precision`/lowest-edge-adjustment in pure numpy (verified against real `pd.cut(...).astype(str)` across positive/negative/duplicate-inducing/inf-edge bins). `return_object=True` builds a `nw.Object` column instead of a pandas-only `astype("O")`.

Verified: `tests/test_discretisation` unchanged (109 passed, 5 pre-existing `check_estimator` failures — raw numpy-array input, rejected by `check_X()` since the narwhals dataframe-only contract; reproduced on the unmodified file). Manually diffed `transform()` output against real `pd.cut` across ~10 edge cases plus the three sibling discretisers' doctest examples — all numerically identical. flake8 / mypy clean, sphinx -W clean.

The other four discretisers' own test files stay pandas-only for now — their `fit()` methods still call `pd.cut`/`pd.qcut` directly and aren't migrated by this branch.


solegalli:
Updated this branch:

- Rebased onto #1045 (shared backend test fixtures and helpers). This branch was based on a commit from before #1019, so it is now stacked on #1045 and its diff includes that PR's commits until it merges.
- No source changes needed: the discretisers go through `BaseNumericalTransformer`, which already handles the narwhals-returning `check_X` since #1022.
- Added `tests/test_discretisation/conftest.py` (`data_california`, `data_normal_dist`, `data_vartypes`, `data_na`), used by #1038-#1042, and rewrote the base discretiser tests to the shared structure.

Locally: `test_base_discretizer.py` 20 passed; no new failures in `tests/test_discretisation`.

PR #1038, opened 2026-08-30
Migrate ArbitraryDiscretiser to narwhals, add polars support

solegalli:
Migrates `ArbitraryDiscretiser` to narwhals with polars support.

`fit()` needed no changes — it delegates entirely to the already-migrated `FitFromDictMixin._fit_from_dict()`. The pandas dependency was in `transform()`'s post-hoc NaN-introduced check, which used `X[...].isnull().sum().sum()` / `.columns` / `.any()` / `.tolist()` — pandas-only calls that broke on polars input coming back from the now-migrated `BaseDiscretiser.transform()`.

Replaced with a narwhals-based per-column check, branched on `return_boundaries` rather than dtype:
- labels (`return_boundaries=True`) use `None` for missing values, which narwhals' `is_null()` detects on both backends.
- codes (`return_boundaries=False`) are numeric, so a numpy float cast + `np.isnan` is used. This isn't just style: narwhals' `is_null()` does **not** see a boxed `np.nan` inside a polars Object-dtype column (`return_object=True`'s output) — verified `is_null().any()` returns `False` there, silently swallowing the warning this method exists to raise. `is_nan()` isn't usable either (narwhals raises "is_nan only supported for numeric dtype, not Object"). The numpy-float-cast approach sidesteps both, confirmed to raise/warn correctly across all pandas/polars × `return_object` × `return_boundaries` combinations.

**Merge vs split:** benchmarked old (pandas-only) vs new (narwhals) `transform()` at 10k/50k/100k rows × 1/2/10 cols on pandas input. `return_object=False` at parity (0.9–1.05x); `return_object=True` 1.15–1.3x slower (100k×10: 36.2ms → 44.9ms) — within the "minimal loss" band, single narwhals path, no `is_pandas` branch. polars faster than pandas at every size.

Tests: `test_arbitrary_discretiser.py` rewritten to one parametrized test per behaviour over `pd.DataFrame`/`pl.DataFrame`; `pytest.raises`/`pytest.warns` switched to `match=`. Verified: `tests/test_discretisation` — 114 passed, same 5 pre-existing `check_estimator` failures. flake8 / mypy clean, sphinx -W clean. "With polars" example added to docstring and `ArbitraryDiscretiser.rst`, output verified.

---


solegalli:
Updated this branch:

- The California housing input comes from the `data_california` fixture instead of a converted pandas frame.
- No source changes.

Locally: `test_arbitrary_discretiser.py` 12 passed; no new failures in `tests/test_discretisation`.

PR #1039, opened 2026-08-30
Migrate EqualFrequencyDiscretiser.fit() to narwhals, add polars support

solegalli:
Migrates `EqualFrequencyDiscretiser.fit()` to narwhals with polars support.

`fit()`'s only pandas dependency was `pd.qcut(duplicates="drop")` for quantile-based bin edges per variable. Replaced with `np.quantile()` on each column's narwhals-extracted numpy array + `np.unique()` to sort and drop duplicate edges — reproducing `qcut`'s `duplicates="drop"` without any per-backend branch.

Getting a **bit-exact** match (not just close) took two fixes verified against pandas 3.0's `qcut` source:
- pandas masks out NaN before calling `np.quantile(values, qs, method="linear")` itself rather than using `np.nanquantile` (not always bit-identical). Moot here — `_fit_setup()` already rejects NaN in `variables_`.
- `qcut` nudges each quantile not exactly representable in base 2 up via `np.nextafter` (rounding up, not to nearest). Skipping this shifted edges by ~1e-13 and broke an existing exact-equality test.

With both applied, verified `np.array_equal` against real `pd.qcut(retbins=True)` across large random floats, many-duplicate data, all-identical data, negative floats, and n<q data.

**Merge vs split:** benchmarked old `pd.qcut` vs the new numpy+narwhals path at 10k/50k/100k rows × 1/2/10 cols — the new path is consistently faster than the old pandas-native code on **both** backends (narwhals-on-pandas 0.19x–0.47x of old `pd.qcut`, narwhals-on-polars 0.12x–0.46x). A narwhals-native quantile-expression alternative was 2–3x *slower* than old `pd.qcut` on pandas. No case for a split.

Verified: `tests/test_discretisation` unchanged (114 passed, 5 pre-existing `check_estimator` failures, reproduced on the unmodified branch tip). flake8 / mypy clean, sphinx -W clean. `test_equal_frequency_discretiser.py` rewritten to one parametrized test per behaviour over `[pd.DataFrame, pl.DataFrame]`. `EqualFrequencyDiscretiser.rst` examples verified against real output (two stale float digits in the `binner_dict_` printout reproduce with the old `pd.qcut` fit — doc staleness, not a regression); "uses pandas.qcut() under the hood" line corrected, "With polars" section added.

---


solegalli:
Updated this branch:

- The polars cases were failing: converting the pandas fixtures to polars needs pyarrow. The vacuous "every bin code is present" check now compares the exact set of codes.
- No source changes.

Locally: `test_equal_frequency_discretiser.py` 12 passed; no new failures in `tests/test_discretisation`.

PR #1040, opened 2026-08-30
Migrate EqualWidthDiscretiser to narwhals, add polars support

solegalli:
Migrates `EqualWidthDiscretiser` to narwhals with polars support.

`fit()`'s only pandas dependency was `pd.cut(bins=int, retbins=True, duplicates="drop")`, used purely to compute equal-width bin edges from each variable's min/max (the discretised codes come from `transform()`, migrated on `narwhals-discretisation-base`). Replaced with `_equal_width_edges()`: a plain `np.linspace(min, max, bins+1)`, reproducing `pandas.cut`'s own edge computation exactly — verified against pandas 3.0's `_nbins_to_bins`/`_bins_to_cuts` source, including the `mn == mx` 0.1%-range widening for constant columns and the `duplicates="drop"` collapse for degenerate float edges. `fit()` now pulls all variables in one `.select(variables_).to_numpy()` call (min/max per column via `axis=0`), following `CyclicalFeatures.fit()`.

**Merge vs split:** benchmarked old pandas-native (`pd.cut` per column) vs the new narwhals+numpy `fit()` at 10k/50k/100k rows × 1/2/10 cols. narwhals-on-pandas is *faster* everywhere except the smallest 10k/1-col case (2.58x slower, sub-ms); at realistic sizes 2–6x faster (100k×10: 19.3ms → 3.0ms). polars faster still. The new path is a speedup on pandas — no `is_pandas` branch, single numpy-driven path for every backend.

Verified `binner_dict_` numerically identical to the old `pd.cut`-based `fit()` across 53 diff cases (random/int/negative values, constant columns, tiny near-duplicate float ranges, two-point and single-value arrays, bins=1) — zero mismatches.

Tests: `test_equal_width_discretiser.py` converted to one parametrized test per behaviour over `[pd.DataFrame, pl.DataFrame]`. Also fixed two vacuous assertions in the original numeric-output test (generator expressions checking truthiness of an always-empty sequence) with real comparisons against `pd.cut` ground truth, and added a constant-column case exercising the `mn == mx` widening branch. Fixed a pre-existing copy-paste bug in `EqualWidthDiscretiser.rst` (a "Return bin boundaries" example set up an `EqualFrequencyDiscretiser`).

Verified: `tests/test_discretisation` — 116 passed, same 5 pre-existing `check_estimator` failures. flake8 / mypy clean, sphinx -W clean. Runs `fit_transform()` on polars with pandas blocked at `builtins.__import__`.

---


solegalli:
Updated this branch:

- The polars cases were failing: converting pandas frames to polars needs pyarrow.
- No source changes.

Locally: `test_equal_width_discretiser.py` 14 passed; no new failures in `tests/test_discretisation`.

PR #1041, opened 2026-08-30
Migrate GeometricWidthDiscretiser.fit() to narwhals, add polars support

solegalli:
Migrates `GeometricWidthDiscretiser.fit()` to narwhals with polars support.

`fit()`'s only pandas dependency was `X[var].min()`/`.max()` to compute the geometric progression's anchors — everything downstream (`np.power`/`np.r_`/`np.sort` bin-edge math) was already plain numpy. Replaced the pandas indexing with `nw.from_native(X, eager_only=True).get_column(var).min()`/`.max()`, which returns a numpy/python float scalar on both backends and feeds `np.power` identically.

**Merge vs split:** benchmarked old pandas-native `fit()` vs narwhals-on-pandas / narwhals-on-polars at 10k/50k/100k rows × 1/2/10 cols (min/max dominate cost; bin-edge math is O(bins) not O(n)):
- narwhals-on-pandas: 1.0–1.3x at realistic sizes (the 1.8x at 10k/1-col is sub-ms fixed overhead) — minimal loss, single narwhals path, no `is_pandas` split.
- narwhals-on-polars: ~0.35–0.7x (1.4–2.8x *faster*).

Verified new `fit()` bin edges against the old pandas implementation across edge cases (skewed/normal/mixed-sign distributions, two-point range, `min == max` degenerate case) on both backends — exact equality. Cross-checked full `fit_transform()` (both `return_object` and `return_boundaries`) between pandas and polars — identical.

Tests: dataframe-touching tests in `test_geometric_width_discretiser.py` parametrized over `pd.DataFrame`/`pl.DataFrame`, replacing pandas-only fixtures with local dicts. Init-only param-validation tests unchanged.

Verified: `tests/test_discretisation` — 114 passed, same 5 pre-existing `check_estimator` failures. flake8 / mypy clean, sphinx -W clean. User-guide worked example re-verified against real output; "With polars" section added.

---


solegalli:
Updated this branch:

- The local `_normal_dist_data` / `_get_column_values` / `_get_column_dtype` helpers are gone.
- No source changes.

Locally: `test_geometric_width_discretiser.py` 32 passed; no new failures in `tests/test_discretisation`.

PR #1042, opened 2026-08-30
Migrate DecisionTreeDiscretiser to narwhals, add polars support

solegalli:
Migrates `DecisionTreeDiscretiser` to narwhals with polars support.

`DecisionTreeDiscretiser` now accepts pandas or polars input via narwhals, extending `BaseNumericalTransformer` directly — **independent of the `BaseDiscretiser` migration**, so this branch is a single commit off `narwhals-migration` with no base-branch stack. Never imports pandas; confirmed the module loads with pandas import blocked.

**Merge vs split:** single narwhals codepath (one `is_pandas` branch only at final column reassembly). Benchmarked at 10k/50k/100k rows × 1/2/10 cols — full fit+transform time is dominated by `GridSearchCV` tree training (10–1500ms) vs plumbing (~0.05–1.4ms per call, <1% of total even where a hand-branched pandas path was ~2x faster on the isolated microbenchmark). Merge also avoids the one-column-at-a-time write pattern that caused pandas fragmentation warnings in the `DecisionTreeFeatures` migration.

Added optional `n_jobs` (default `None` = sequential, unchanged behaviour), parallelising per-variable tree fits with joblib threads, mirroring `DecisionTreeFeatures`. Net loss on small workloads (0.6–0.8x), real win with enough work (1.4–2.3x). `n_jobs=2` produces identical trees and predictions to `n_jobs=None`.

**Bug fixed** (introduced by the base-transformer narwhals migration, not present pre-migration): `check_X` used to always copy its pandas input; the narwhals-based `check_X` no longer does, so the old `transform()`'s in-place `X[feature] = ...` assignments would have mutated the caller's original dataframe. Rewrote `transform()` to batch every replacement column and apply them in one non-mutating `.assign()` (pandas) / `.with_columns()` (polars) call.

Reimplemented `pandas.cut`'s binning (`bin_number`/`boundaries` outputs) without importing pandas: `np.digitize` for assignment, plus a from-scratch port of pandas' internal `_round_frac`/`_infer_precision` label-rounding so boundary labels are byte-for-byte identical to old `pd.cut` output (verified across 500 randomized trials, zero mismatches).

Tests rewritten to one parametrized test per behaviour over `[pd.DataFrame, pl.DataFrame]`. Fixed `test_non_fitted_error`, which instantiated `EqualWidthDiscretiser` instead of `DecisionTreeDiscretiser` (pre-existing copy-paste bug, confirmed on main).

Verified: `tests/test_discretisation` — 123 passed (was 108, +15 from parametrization), same 5 pre-existing `check_estimator` failures. flake8 / mypy clean, sphinx -W clean. "With polars" and "Training trees in parallel" doc sections added, verified against real output.


solegalli:
Updated this branch:

- **Now stacked on #1037** (previously standalone), so the shared discretisation test data lives in a single PR; rebased onto it (and therefore #1045). Its diff includes those PRs' commits until they merge.
- Tests rewritten to the shared structure: `data_normal_dist` fixture, the target built with `make_series`, a new test passing the target as a list and as a numpy array, `isinstance(X, make_df)` plus `to_dict()` checks, and `pytest.raises(match=re.escape(msg))`.
- No source changes.

Locally: `test_decision_tree_discretiser.py` 45 passed; no new failures in `tests/test_discretisation`.

PR #1043, opened 2026-09-01
Pass min_periods to rolling() in WindowFeatures

VenishPaneliya:
### What this fixes

`WindowFeatures` documents `min_periods` as a pandas passthrough:

> `min_periods: int, default None.` Minimum number of observations in the window required to have a value; otherwise, the result is np.nan. See parameter `min_periods` in pandas `rolling()` documentation for more details.

It is accepted and stored as `self.min_periods`, but `transform()` never reads it. Both branches build the window without it:

```python
X[self.variables_].rolling(window=win)          # list-of-windows branch
X[self.variables_].rolling(window=self.window)  # single-window branch
```

So `rolling()` keeps its default — a full window is required — and the leading rows stay `NaN` no matter what the user passes.

`ExpandingWindowFeatures` already does forward it (`.expanding(min_periods=self.min_periods)`), so the two transformers disagreed on a parameter they document in the same words.

### Evidence

A 6-row frame, `window=3`, default `functions="mean"`, `periods=1`:

| | result |
|---|---|
| `min_periods=None` | `[nan, nan, nan, 2.0, 3.0, 4.0]` |
| `min_periods=1` (before) | `[nan, nan, nan, 2.0, 3.0, 4.0]` — identical, flag inert |
| `min_periods=1` (after) | `[nan, 1.0, 1.5, 2.0, 3.0, 4.0]` |
| `X["x"].rolling(3, min_periods=1).mean().shift(1)` | `[nan, 1.0, 1.5, 2.0, 3.0, 4.0]` |

After the change the output matches pandas exactly, for a single window and for a list of windows.

### Compatibility

The default is unchanged. `min_periods=None` is precisely what `rolling()` was already assuming, so anyone who never set the parameter sees identical output. Only users who explicitly passed it see a change — and it is the documented behaviour they asked for.

### Testing

Added `test_min_periods_is_used` and `test_min_periods_is_used_with_multiple_windows`, comparing against pandas `rolling(...)` directly and asserting the default still requires a full window.

- without the fix: both fail
- with the fix: both pass
- `tests/test_time_series`: 121 passed
- flake8 and black clean on both files

### DIFF
--- a/feature_engine/timeseries/forecasting/window_features.py
+++ b/feature_engine/timeseries/forecasting/window_features.py
@@ -221,7 +221,7 @@ class WindowFeatures(BaseForecastTransformer):
             for win in self.window:
                 tmp = (
                     X[self.variables_]
-                    .rolling(window=win)
+                    .rolling(window=win, min_periods=self.min_periods)
                     .agg(self.functions)
                     .shift(periods=self.periods, freq=self.freq)
                 )
@@ -231,7 +231,7 @@ class WindowFeatures(BaseForecastTransformer):
         else:
             tmp = (
                 X[self.variables_]
-                .rolling(window=self.window)
+                .rolling(window=self.window, min_periods=self.min_periods)
                 .agg(self.functions)
                 .shift(periods=self.periods, freq=self.freq)
             )
--- a/tests/test_time_series/test_forecasting/test_window_features.py
+++ b/tests/test_time_series/test_forecasting/test_window_features.py
@@ -452,6 +452,43 @@ def test_multiple_windows(df_time):
     assert df_time_tr.equals(X_tr)
 
 
+def test_min_periods_is_used(df_time):
+    # min_periods was accepted and documented but never reached pandas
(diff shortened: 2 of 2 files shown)


VenishPaneliya:
Just to note for anyone triaging: the `py39` failure here looks unrelated to this change. All 52 failures are `test_check_estimator_from_sklearn` / `_SklearnCheckInputWrapper raised an exception`, across `test_transformation`, `test_wrappers` and the forecasting checks — estimators this PR doesn't touch. The change is two `min_periods=self.min_periods` arguments inside `WindowFeatures.transform`.

CI looks broadly red on the repo at the moment (#1042 is failing py311, py312, py313, py314 and codecov), so I assume this is the same environment/dependency issue rather than anything here.

Locally, on this branch: `tests/test_transformation/test_check_estimator_transformers.py`, `tests/test_wrappers/test_check_estimator_wrappers.py` and `tests/test_time_series` give 154 passed.

Happy to rebase once CI is green again if that helps.

solegalli:
Thank you @VenishPaneliya ! Well done on catching this bug :)

PR #1044, opened 2026-09-08
Use the numpydoc Returns section header so return values render

VenishPaneliya:
Six methods head their return section `Return` instead of `Returns`. numpydoc only recognises `Returns`, so the content is dropped — the documented return value doesn't render at all.

Feeding `TargetMeanRegressor.predict`'s current docstring to numpydoc:

```
UserWarning: potentially wrong underline length...
Return
-------
UserWarning: Unknown section Return

BEFORE (Return) : Returns entries = 0   <-- return value NOT rendered
AFTER  (Returns): Returns entries = 1  -> y_pred: ndarray of shape (n_samples,)
```

The rest of the package writes `Returns` in 97 places, so these six are outliers rather than a house style:

- `_prediction/base_predictor.py` — `_transform`, `_predict`
- `_prediction/target_mean_classifier.py` — `predict_proba`, `predict_log_proba`, `predict`
- `_prediction/target_mean_regressor.py` — `predict`

Three of those are public sklearn-API methods, so this is what users see on the API pages.

**`base_predictor._transform`** also had its whole docstring body indented one space deeper than the entry beneath it, which is what produced the second warning above ("potentially wrong underline length"). I normalised it to the 8-space indentation every other docstring in that file uses, which is why that file's diff is a little larger.

**`DropMissingData.return_na_data`** is a different case: it has no return section at all. Its return value was documented under `Parameters` as `X_na`, which left the actual parameter `X` undocumented:

```python
def return_na_data(self, X: pd.DataFrame) -> pd.DataFrame:
    """
    ...
    Parameters
    ----------
    X_na: pandas dataframe of shape = [n_samples_with_na, features]
        The subset of the dataframe with the rows with missing data.
    """
```

I moved `X_na` into a `Returns` section and documented `X` with the same wording `transform()` uses directly above it, so the two methods now read consistently.

Docs only — no code changed. `tests/test_imputation/test_drop_missing_data.py` passes (9 tests) and `flake8` is clean on all four files.

One note: `black --check` reports `drop_missing_data.py` as needing reformatting, but it does so on an unmodified checkout too, so that's pre-existing and not from this change.


solegalli:
Thanks for spotting this @VenishPaneliya, and for the numpydoc before/after check, it made the problem easy to confirm.

I've changed the target branch to narwhals-migration, which is where we're currently developing the next release. Because your branch was created from main, it also picked up two commits that aren't on narwhals-migration yet, so I rebased it onto narwhals-migration, keeping only your commit (you're still the author).

There was one conflict in drop_missing_data.py: on narwhals-migration that docstring already says "dataframe" instead of "pandas dataframe", so I kept that wording along with your new X parameter and Returns section. No action needed on your side. Thanks again!

